PR #33, opened 2016-12-31
add space after `\A, \E` for use by humans, translate accordingly in TLATeX

johnyf:
TLATeX translates:

``` tla
\A -> \A\,
\forall -> \forall\,
```

and `tlatex.sty` defines:

`\A` as `\forall`

However, a human using `tlatex.sty` to write TLA+ directly in LaTeX  (without calling TLATeX, for example loading `tlatex.sty` via `tla2texplus-book.sty`) gets quantifiers that are very close to the bound variables.
One could redefine `\A` in the preamble, but that won't work for using both TLATeX and human-written TLA+ math in the same LaTeX document.
(The patches to `tla2texplus-book.sty` that I considered won't work for both humans and TLATeX.)

Instead, I would like to suggest the TLATeX translation:

``` tla
\A -> \forall\,
\forall -> \forall\,
```

and the `tlatex.sty` definition:

`\A` as `\forall\,` (same as in `tla2.sty`)

This works for both humans (who can type the "high level" `\A`) and TLATeX (which generates the "low level" `\forall`). Moreover, anyone who wants greater control over spacing can still define custom operators using `\forall` (or redefine `\A`).

Similar changes have been made for `\E`.

### A slightly different approach

I considered the alternative:

```latex
\usepackage{amsopn}
\DeclareMathOperator{\A}{\forall}
\DeclareMathOperator{\E}{\exists}
```

This introduces the same amount of horizontal space after each operator as `\,`.
However, it introduces also some space before each operator, so the typeset document would change. Besides, it depends on the `amsopn` package.

### DIFF
--- a/tlatools/src/pcal/PcalBuiltInSymbols.java
+++ b/tlatools/src/pcal/PcalBuiltInSymbols.java
@@ -166,9 +166,9 @@ public final class PcalBuiltInSymbols
         add("]",   "]",           Symbol.RIGHT_PAREN, 0);
         add(">>",  "{\\rangle}",  Symbol.RIGHT_PAREN, 0);
 
-        add("\\A",         "\\A\\,",          Symbol.PREFIX, 0);
+        add("\\A",         "\\forall\\,",     Symbol.PREFIX, 0);
         add("\\forall",    "\\forall\\,",     Symbol.PREFIX, 0);
-        add("\\E",         "\\E\\,",          Symbol.PREFIX, 0);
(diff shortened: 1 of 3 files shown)


lemmy:
Hi Ioannis,

thanks for your efforts but I'm afraid we have to reject it. You propose to change the tlatex LaTeX package. This will change every document that uses that package. That is not Leslie's philosophy. He won't approve of no change to any LaTeX package he has written that is not a simple bug fix.

If the change is important to you, create new LaTeX package and use the support in tla2tex.TLA to read alternative styles.

M.

Issue #88, opened 2017-10-13
LevelNode and select subclasses potentially use SetOfLevelConstraints not as it was designed for


quaeler:
`SetOfLevelConstraints` is a Map that is "supposed to" use `SymbolNode` for its keys - as demonstrated by [its code's comments](https://github.com/tlaplus/tlaplus/blob/master/tlatools/src/tla2sany/semantic/SetOfLevelConstraints.java#L14), [its `toString()` implementation](https://github.com/tlaplus/tlaplus/blob/master/tlatools/src/tla2sany/semantic/SetOfLevelConstraints.java#L48) and its usage elsewhere in the codebase.

`LevelNode` is written so that it appears that it is using this Map with keys of LevelNode in its [`addTemporalLevelConstraintToConstants(HashSet, SetOfLevelConstraints)` method](https://github.com/tlaplus/tlaplus/blob/master/tlatools/src/tla2sany/semantic/LevelNode.java#L187) (even though [its own javadocs say otherwise](https://github.com/tlaplus/tlaplus/blob/master/tlatools/src/tla2sany/semantic/LevelNode.java#L235))

`LevelNode`'s subclasses `AssumeNode` and `AssumeProveNode` rely upon the above linked method for their implementations of `levelCheck(int)` so they too are involved in this.

* Should consumer code call `toString()` on an instance of `SetOfLevelConstraints` which `LevelNode` has been affecting, a class cast Exception would potentially be thrown
* Should `LevelNode.addTemporalLevelConstraintToConstants(...)` actually be receiving instances of `SymbolNode` in its HashSet and not `LevelNode`, then that method should be updated to reflect this.

quaeler:
Verifying all usages of the `addTemporalLevelConstraintToConstants(...)` method, there is a consistency in handing it a `HashSet<SymbolNode>` so it is the method mis-casting elements of the HashSet.

I will pick this up and fix it.

quaeler:
This has been addressed now - closing.

PR #91, opened 2017-10-15
88 level node add temporal level constraint to constants

quaeler:
Code addressing #88 as well as adding typing and annotations, fixing the odd misspelling and moving fully-qualified-class-in-method-signature to an import.


lemmy:
I pushed an amended the commit 6ab8080699355a3a353ed63e3254f04144b7a3d6 instead (does not define a serial version uid in SetOf*)

PR #96, opened 2017-10-21
Providing a manner in which to have TLC suspend execution waiting on a debugger

quaeler:
There is a way this could also be done by, for example, adding some to the launch configuration parameter with attribute name `distributedTLCVMArgs` and although something seems to be already populating that with a zero-length String (as opposed to it not existing in the configuration) in my runs, it's plausible that you'd reject this PR as needless in favor of that.

If not, this provides the functionality for the TLC to suspend waiting for a debugger on port 9321, should the Toolbox be started with the environment variable `LAUNCH_TLC_IN_DEBUG` defined.


quaeler:
No idea why the entire file is considered changed; i pasted in 4 lines of code and typed 3 more of comments at line 141.

In Stash i can say "ignore whitespace changes" in PR diffs, but can't seem to find such a thing in the GitHub view. :- /

lemmy:
You likely changed the line endings which is why Github shows the whole file changed.

Anyway, I'm not convinced we actually need the functionality. One can always add any JVM argument via the Model Editor's Advanced page under "TLC Options".

quaeler:
Was not aware of that on the Advanced page - totally suffices.

Issue #92, opened 2017-10-17
TLC reports zero coverage for conjunct of the next-state action


lemmy:
TLC reports 0 coverage for the indicated conjunct of the next-state action below:

```tla
  VARIABLE x
  Spec == /\ x = 0 
          /\ [][/\ x' = (x+1)%5
                /\ x' = (x+1)%5  \* <- 0 coverage
               ]_x
```

This is an oversimplification of a problem that occurred in a real spec.  A slightly more complicated example indicates how it can arise in practice.

```tla
   /\ A
   /\ \/ /\ x' # x
         /\ ...
      \/ /\ x' = x   \* <- 0 coverage
         /\ ... 
```
where

```tla
   A ==  \/ /\ ...
            /\ x' = ...
            /\ ...
         \/ /\ ...
            /\ x' = x
            /\ ...
```

Related, TLC reports zero coverage for the (automatically) generated `vars = <<x,y,z>>` definition of PlusCal unless `vars` appears as part of an UNCHANGED statement (see issue #152).

quaeler:
From experimentation, it also appears that:
```tla
Is92 == /\ state = "eins"  
        /\ y' \in 800..1000
        /\ y' \in 0..5000
        /\ state' = "deux"
        /\ state' = "deux"
```
has zero coverage on the second NSA for both `y` and `state`.

```tla
Is92 == /\ state = "eins"  
        /\ y' \in 800..1000
        /\ y' \in 0..5000
        /\ state' = "deux"
        /\ state' = "three"
```
has zero coverage for all actions.

```tla
Is92 == /\ state = "eins"  
        /\ y' \in 800..1000
        /\ y' \in 0..5000
        /\ y' = y + 5
        /\ state' = "deux"
        /\ state' = "deux"
```
has zero coverage for all actions.

quaeler:
This is happening because of the design of the code in `Tool.getNextStatesAppl(OpApplNode, ActionItemList, Context, TLCState, TLCState, StateVec)`

A second NSA for the same variable enters into this method carrying the `TLCState` (`TLCStatusMutSource`) which has already bound the variable to a value defined in the previous NSA.  The code handling as least the equals and in operators determines that the primed variable is already bound, and does not attempt to bind for the current pred.

It is the binding action, though, which marks an internal object array of `SemanticNode` in the `TLCState` which is later consulted for code coverage calculations.

I'll make a fix addressing this.

lemmy:
[...]

PR #98, opened 2017-10-22
Addressing the counting coverage of multiple same-variable NSAs at a single spec depth

quaeler:
This is to address #92 - this also fixes undocumented issues within the `TLCStateMutSource` code which could result in NPEs.


lemmy:
[A slightly modified version has been committed to master](fb99862c09cedf0bb02c59faeaddd6ca55c89bba).

Issue #101, opened 2017-10-24
Toolbox no longer opens progress dialog on TLC startup


lemmy:
The Toolbox used to open a modal progress dialog on TLC startup. The dialog was how the Toolbox communicated the TLC startup to the user. [Starting with Eclipse 4.7 clever people thought it a good idea to - after a decade - change the default to not open the dialog anymore](https://bugs.eclipse.org/bugs/show_bug.cgi?id=146205#c10).

The behavior can ([still](https://bugs.eclipse.org/bugs/show_bug.cgi?id=497429)) be manually changed back on the "Generals" preference page by unchecking the very descriptive "Always run in background" checkbox.

PR #107, opened 2017-11-02
Addresses toolbox no longer opening progress panel on TLC execution

quaeler:
This addresses issue #101.

There seems to be no discernible rhyme nor reason as to why one file gets all of its line endings changed while another one being worked on does not (unless the one which did not already had my flavor of them.)


lemmy:
Applied with commit 9fc9eabfe8319c4622c7008306310216fa4ede06

Issue #110, opened 2017-11-04
Preference storage is bifurcated between activators


quaeler:
Noticed while working on #104, we actually were setting the snapshotting preference default on preference initialization into the preferences hanging [off the `TLCUIActivator`](https://github.com/tlaplus/tlaplus/blob/master/org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/ui/preference/TLCPreferenceInitializer.java#L36).

On value change, it got set into the preferences hanging [off the `TLCActivator`](https://github.com/tlaplus/tlaplus/blob/master/org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/ui/preference/TLCPreferencePage.java#L37)

As part of the TLC execution the preference is checked in the preferences hanging [off the `TLCActivator`](https://github.com/tlaplus/tlaplus/blob/master/org.lamport.tla.toolbox.tool.tlc/src/org/lamport/tla/toolbox/tool/tlc/launch/TLCModelLaunchDelegate.java#L992).

Ideally, we would have preferences stored against one Activator; `TLCActivator` would not require plugin dependency modifications, but the preferences page is not configured to pull from those preferences. `TLCUIActivator` supports the preferences page, but would introduce dependency requirements to the plugin containing `TLCModelLaunchDelegate`.  ...

Regardless, i'm going to finish implementing #104 in a version of this bifurcation which fixes the initialization issue; then we can further address this later.

lemmy:
Is this still an issue?

quaeler:
It is.

lemmy:
Do we need to fix it?

quaeler:
It's an architectural detail but not necessary.

Issue #104, opened 2017-10-26
Implement snapshot history depth preference functionality


quaeler:
As first raised in #95, change the checkbox preferences widget present starting in 1.5.4 which turns snapshotting on and off to a spinner that defines how many snapshots to keep (with zero meaning perform no snapshotting.)

Implement the janitorial functionality as part of the snapshotting process.

lemmy:
Applied your changes in commit https://github.com/tlaplus/tlaplus/commit/3c087ca428554d796ad7404941fa7ae695b7a1dd. In the future don't bother to open pull requests. Just send me the commit id and I will cherry-pick from your fork.

PR #111, opened 2017-11-04
Address snapshot history size preferences and functionality

quaeler:
This addresses #104.

A note should be added to the 1.5.5 release notes (which i couldn't find in the repository or i would have included that in this PR) that strongly states that this new functionality has a default value of 10 and that should the user have more than the defined preference value of snapshots, the next time they run the model checker successfully, they will have the older ones deleted.

This also makes the initialization bug cited in #110 obsolete, but not the larger architectural issue discussed there.


quaeler:
`TLCModelLaunchDelegate`'s actual code change is at line 993-994 - the inability to ignore whitespace changes in GitHub diffing seems cuckoo-banana.

lemmy:
https://github.com/tlaplus/tlaplus/commit/3c087ca428554d796ad7404941fa7ae695b7a1dd

Issue #116, opened 2017-11-17
Make state graph more readable by humans


jazzdelightsme:
The new state graph feature is cool!

For my particular model, though, I can't really read it. My model has nine variables, the state of which is printed inside each state (49 distinct states in the error case that I am looking at; something more like 110 when fixed). It's very impressive-looking, but I can't use it or show it to anybody.

Some things that might improve it:
  * Inside each state, either show only the variable that changed, or perhaps highlight the one that changed (similar to the error trace view). That would allow me to visually walk through the graph, seeing "in this step, we increased the refcount; in this step, we acquired the lock; etc.".
  * It might also be helpful to label state transitions with step names. I don't know how well that will work visually, but it is how I would ideally be able to read the graph--as I trace my finger through the graph, I can say "the IncrementRefCount step was taken, then the AcquireLock step was taken, etc.". (see also #115)

lemmy:
@jazzdelightsme Do you have the bandwidth to work on this feature if I guide you?

jazzdelightsme:
> Do you have the bandwidth?

Unfortunately, not in the near future. :(

lemmy:
[http://www3.hhu.de/stups/prob/index.php/Main_Page](ProB) has more sophisticated [support to visualize state graphs](http://www3.hhu.de/stups/prob/index.php/State_space_visualization_examples#Alternating_Bit_Protocol) and - fortunately - can model check TLA+ specifications.
I suggest to try out ProB to identify those features that should be included in the TLA+ Toolbox.

will62794:
Hello,

This is a feature that I think would be nice to add to the TLA+ toolbox and I had some time to work on it. I played around with some simple specs and visualized them in the TLA+ toolbox and also in ProB, to see how they compare. I also started working on a POC for showing/highlighting the changed variables in state graph transitions. I used the following simple spec for testing state graph visualization:

```tla
VARIABLE varA, varB
Max == 3

Init == varA = 0 /\ varB = 0

IncA == (varA' = (varA+1) % Max)
IncB == (varB' = (varB+1) % Max)

Next == \/ IncA /\ UNCHANGED varB
        \/ IncB /\ UNCHANGED varA
        \/ IncA /\ IncB
```

Here is an example of how ProB visualizes the state space:

![image](https://user-images.githubusercontent.com/6131840/34903234-42724fec-f7fb-11e7-8d1d-bea444d07393.png)

And how the TLA+ toolbox would with the changes from my POC:

[...]

PR #138, opened 2018-02-03
Make state graph more readable with colorized edges and labels (Issue #116)

will62794:
Hi @lemmy, these are my changes to add the state graph edge colorization and labels, as we discussed on the comment thread for #116 . While the patch is mostly complete, I will note that I did not decide on a way to omit edge labels based on a "maximum state count threshold". I added a class parameter in `DotStateWriter` but not the logic for it. I am worried that to actually do this might introduce extra complexity, since it doesn't seem like we have any way up front to determine how many total states will be in the graph, until the state graph is finished being built. Wanted to see if you had any suggestions there.

Of course, just let me know if I should do anything differently, conform to certain conventions, etc. I was able to update and run the `DumpAsDotTest` unit test and it passes. I was able to run some other unit tests but when I tried to run them in bulk I ran into an issue where tests would fail because it seemed like they were trying to write to a `states` directory that already existed, or something like that. Perhaps I was just running the tests incorrectly.

For reference, this is one final sample with these latest changes applied:
[simple_spec_newest.pdf](https://github.com/tlaplus/tlaplus/files/1691379/simple_spec_newest.pdf)

Honestly, I am still a little bit wary of the edge labels, since they do add clutter and I think just colorization with a legend actually looks pretty good. Maybe we could consider leaving them out for now. See what you think. Thanks!


tlatools/src/tlc2/util/DotStateWriter.java
```diff
@@ -37,15 +47,60 @@
...
+	// Maximum number of unique colors. Determined by the Graphviz color scheme that is used.
+	static private final Integer colorGenMax = 12;
+	
+	// If the total number of states in the resulting graph exceeds this count, we
+	// omit action labels no matter what (even if the option was specified as
+	// 'true'). The edge labels add considerable visual clutter for graphs of
+	// non-trivial sizes, and so we aim to not make graphs unreadable. The choice of
+	// this threshold is somewhat arbitrary, but it serves the right purpose i.e.
+	// include edge labels on only the smallest of state graphs.
+	static private final Integer maxNumStatesForActionLabels = 35;
```
will62794: This is the parameter that I added but did not include logic for. Wanted to see if you had any suggestions on the simplest way to implement this.

lemmy:
You pull request with minor refactorings has been merged (https://github.com/tlaplus/tlaplus/commit/7e80f1d6301c42c5381ef63c2ed26166c35f12bc).

Do note though that DotStateWriter#state2html is broken as indicated on the TODO statement. I changed the code to use the old state2dot method instead. https://github.com/tlaplus/tlaplus/blob/7e80f1d6301c42c5381ef63c2ed26166c35f12bc/tlatools/src/tlc2/util/DotStateWriter.java#L258-L271

Issue #63, opened 2017-07-21
PlusCal call-goto enhancement


lemmy:
The user Chris Pacejo [provided a patch](https://groups.google.com/forum/#!topic/tlaplus/6M1oFOtN-5k) to enhance PlusCal go-to statements:

"Hi, I have found it desirable in PlusCal to be able to issue a call immediately followed by a goto without an intervening label (for example, in certain cases inside a with block).  I see no reason why this relaxation of the labeling rules should be invalid (indeed, it is not much different from the call-followed-by-return case).

Attached please find a patch which adds this feature to the PlusCal compiler.  Feedback would be appreciated as I'm sure I've missed something.  In particular I'm not sure which of the various .tla files in the source tree are kept up-to-date (at least one of them has "Old" in its name); I'd be glad to also modify the relevant modules if I can be pointed in the right direction.  I have tested the patch and it seems to work for my purposes."

This patch has been turned into the [feature branch mku-pluscal](https://github.com/tlaplus/tlaplus/tree/mku-pluscal) and awaits merging into master.

lemmy:
73a918cfdee627677d1c3c6916b3ef384c9f170a

PR #144, opened 2018-02-17
Permit PlusCal "call" statement to be followed directly by "goto" without intervening label

cpacejo:
As discussed here: https://groups.google.com/d/topic/tlaplus/6M1oFOtN-5k/discussion

If a call is followed directly by a goto, we can simply use the goto's target as the return label for the call, thus negating the requirement for the call to be followed by a label.  This helps eliminate dummy labels/states for certain cases.

The implementation is simply an adaptation of the existing machinery for a call followed directly by return.  Manual tests produce the desired output.

### DIFF
--- a/tlatools/src/pcal/AST.java
+++ b/tlatools/src/pcal/AST.java
@@ -17,7 +17,7 @@
 * However, there are the following classes that do not represent explicit  *
 * non-terminals of the +CAL grammar.                                       *
 *                                                                          *
-*     Uniprocess   MultiProcess   SingleAssign   CallReturn                *
+* Uniprocess   MultiProcess   SingleAssign   CallReturn   CallGoto         *
 *                                                                          *
 * Every AST has col and line fields that contain the position of the       *
 * first character of the corresponding portion of the algorithm text (as   *
@@ -100,6 +100,7 @@ public class AST
     public static AST.Call         CallObj         ;
     public static AST.Return       ReturnObj       ;
     public static AST.CallReturn   CallReturnObj   ;
+    public static AST.CallGoto     CallGotoObj     ;
     public static AST.Goto         GotoObj         ;
     public static AST.Macro        MacroObj        ;
     public static AST.MacroCall    MacroCallObj    ;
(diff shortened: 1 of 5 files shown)


lemmy:
@cpacejo Can you attach a spec that can be used as a test for this feature?

Also see issue #63

cpacejo:
Ah sorry didn't see that you had opened an issue!

Here are two TLA specs which make use of the feature, with accompanying translations.  (They are identical, except one is missing labels, so that it can be shown that the `-label` option does not add the now-unnecessary label before the `goto`.)

[CallGotoTest.txt](https://github.com/tlaplus/tlaplus/files/1735350/CallGotoTest.txt)
[CallGotoTestUnlabeled.txt](https://github.com/tlaplus/tlaplus/files/1735351/CallGotoTestUnlabeled.txt)

lemmy:
Superseeded by 73a918cfdee627677d1c3c6916b3ef384c9f170a

@cpacejo Can you please check why the unit test https://github.com/tlaplus/tlaplus/blob/73a918cfdee627677d1c3c6916b3ef384c9f170a/tlatools/test/pcal/CallGotoUnlabeledTest.java fails? Once the test passes, we can merge to master.

cpacejo:
Can you please give some instruction how to run the test?  It's not clear to me from the Ant build file how to run an individual test (the tests take a long time to run and don't appear to be incremental).

Nor am I sure whether the test would even be run by one of the test targets – does it need to be added to a manifest to ensure that it is?  If not, which target should I expect to run it?

lemmy:
The "test" target of the ant build runs all tests based on pattern matching. If you just want to run a CallgotoUnlabeledTest, change line 296 of customBuild.xml from "**/*Test*.java" to "**/CallgotoUnlabeledTest.java". However, it is easiest if you just run CallgotoUnlabeledTest from inside the Eclipse IDE.

If you need instructions on how to setup Eclipse for Toolbox development, please check out https://github.com/tlaplus/tlaplus/tree/master/general/ide. It only takes ~10 minutes a most of it is automated.

cpacejo:
Looks like the test is failing to expect the two label generation messages that CallGotoUnlabeledTest should generate.  Changing the assertions to:

                assertTrue(Arrays.toString(messages), messages.length == 6);

                assertEquals("The following labels were added:", messages[0]);
                assertEquals("  Lbl_1 at line 9, column 3", messages[1]);
                assertEquals("  Lbl_2 at line 17, column 3", messages[2]);
                assertEquals("Parsing completed.", messages[3]);
                assertEquals("Translation completed.", messages[4]);
                assertEquals("New file test-model/" + fileName + " written.", messages[5]);

seems to work.  (Although the last assertion fails on my setup because of some filename issue that isn't relevant.)

Alternatively just use CallGotoTest.tla for both tests, and expect that no labels should be generated.  Though I think that's a bit more brittle a test.

Issue #147, opened 2018-02-24
Multi-threaded TLC Simulation Mode


will62794:
Currently, TLC does not take advantage of multiple threads when running in simulation mode (`-simulate`). It would be desirable to have this. Presumably, each thread would start up a simulation run with a different seed, to increase total randomized coverage. 

As a workaround, it is currently possible to start up multiple TLC processes (separate JVM instances) with different simulation seeds to achieve essentially the same effect.

kapil-ms:
+1

will62794:
@lemmy I started to explore this a bit. It would likely consist of refactoring [Simulator.java](https://github.com/tlaplus/tlaplus/blob/bfdbe00e3d0638b46a29ca408c91c0e17d832661/tlatools/src/tlc2/tool/Simulator.java) to start up multiple worker threads that each generate a random trace with different seeds. Eventually, I wonder if the trace simulator could be ported over to be an  implementation of the `AbstractChecker.java` interface. For example, as a `RandomizedModelChecker.java`, but that's a more significant change. Perhaps it could be a desirable code cleanup down the road, though. Are you on board with the `Simulator.java` recfactor approach? If so, I'll see if I can work on a patch in the next few weeks/month. 

On a somewhat unrelated note, I have been trying to figure out how to run a specific JUnit test from the command line. I am able to compile and run all tests with a command like 
```
ant -f tlatools/customBuild.xml compile compile-test dist test-dist
````
but haven't figured out how to run just a single test.

will62794:
@lemmy Another somewhat related question: is it possible to run the unit tests in parallel? It seems like you can set a `threads` parameter for [this](https://github.com/tlaplus/tlaplus/blob/master/tlatools/customBuild.xml#L334) JUnit task, but I'm wondering if it's supported given the way tests are currently written.

lemmy:
There is no support for running tests in parallel right now. The test suite however has become sufficiently large for parallelization to indeed be desirable. A potential problem is a directory name collision if two or more tests try to create TLC's temporary storage directory concurrently. util.FileUtil.makeMetaDir(Date, String, String) has support to append milliseconds to the directory name to minimize the collision probability though.

Fundamentally though, the problem is that each test has to run in its own JVM instance. This is owed to the fact that TLC does not correctly re-initializes some internal data-structures which makes checking of two or more different specs/models in a single JVM instance impossible.

[...]

PR #193, opened 2018-09-14
Support multi-threaded simulation mode

will62794:
This is an initial patch for adding multi-threading support to TLC's random simulation mode (#147) . The bulk of the changes are in Simulator.java. This patch is fairly complete, but I would like to get your feedback to make sure you are on board with the changes, and in case there are some edge cases I missed or handled incorrectly. I would like to add more extensive test coverage for the Simulator before committing this change, but wanted to make sure the testing approach I took is reasonable. I tried to refactor things to make them cleaner where appropriate, but tried to keep any significant restructuring efforts to a minimum. Most of the diff is a result of moving the original trace generation logic into the SimulationWorker thread loop.

Additionally, it would be great add some benchmarks for simulation mode using multiple threads/cores. I did some cursory performance testing, but maybe you have thoughts on the best way to incorporate perf measurements in a maintainable/observable way.


tlatools/src/tlc2/tool/Simulator.java
```diff
@@ -103,41 +114,303 @@ public Simulator(String specFile, String configFile, String traceFile, boolean d
 	private final Action[] impliedActions; // the implied-actions to be checked
 	private final boolean checkDeadlock; // check deadlock?
 	private final boolean checkLiveness; // check liveness?
-	private long numOfGenStates;
+	
+	// The total number of states generated by all workers. Written to concurrently.
+	private AtomicLong numOfGenStates;
```
lemmy: [LongAdder](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/LongAdder.html) to reduce contention?

tlatools/src/tlc2/tool/Simulator.java
```diff
@@ -103,41 +114,303 @@ public Simulator(String specFile, String configFile, String traceFile, boolean d
...
 	private final long seed;
 	private long aril;
 	private final ObjLongTable astCounts;
 	private boolean isCancelled; // SZ Feb 24, 2009: cancellation added
 	private Value[] localValues = new Value[4];
 
+	// The set of all initial states for the given spec. This should be only be
+	// computed once and re-used whenever a new random trace is generated. This
+	// variable should not be written to concurrently, but is allowed to be read
+	// concurrently.
```
lemmy: It might makes sense to support TLC's continue option to keep state exploration going even if an error is found.

tlatools/src/tlc2/tool/Simulator.java
```diff
@@ -103,41 +114,303 @@ public Simulator(String specFile, String configFile, String traceFile, boolean d
...
+	private StateVec initStates = new StateVec(0);
+
+	// Appended to when a simulation worker completes. We only care about the first
```
lemmy: It might makes sense to support TLC's continue option to keep state exploration going even if an error is found.

tlatools/test/tlc2/tool/simulation/SimulatorMultiThreadTest.java
```diff
@@ -0,0 +1,39 @@
...
+ * This test intends to verify that, from a correctness perspective, running the
+ * Simulator with multiple threads is equivalent to running it with a single
+ * thread i.e. it should pass all of the same tests.
+ */
+public class SimulatorMultiThreadTest extends SimulatorTest {
+
+	public int numWorkers() {
+		// A somewhat arbitrary value, but 4 threads seems like a reasonable amount of
+		// parallelism.
+		return 4;
```
lemmy: A homage to https://www.engadget.com/2010/12/29/hackers-obtain-ps3-private-cryptography-key-due-to-epic-programm/

tlatools/test/tlc2/tool/simulation/SimulatorTest.java
```diff
@@ -0,0 +1,111 @@
...
+		runSimulatorTest("BasicMultiTrace", "MCUnderspecInit", false, 100, 100);
+		assertTrue(recorder.recorded(EC.TLC_STATE_NOT_COMPLETELY_SPECIFIED_INITIAL));
+	}
+	
+	@Test
+	public void testDeadlock() {
+		runSimulatorTest("BasicMultiTrace", "MC", true, 100, 100);
+		assertTrue(recorder.recorded(EC.TLC_DEADLOCK_REACHED));
+	}
+
```
lemmy: I noted this on issue #147, running multiple different specs/models in a single JVM might be problematic. I do not know if this is a problem here though.

lemmy:
The patch looks good to me (see my comments above). **Thank you very much for your efforts!**

WRT benchmarks, I have recently started using JMH. The approach seems fairly complete to me and - despite being targeting micro-benchmarks - should fit this use case. See:

- https://github.com/tlaplus/tlaplus/tree/master/tlatools/test-benchmark
- http://jmh.morethan.io/?sources=https://raw.githubusercontent.com/tlaplus/tlaplus/master/tlatools/test-benchmark/tlc2/tool/fp/LongArrayBenchmark-1535138934-548ce71f5.json
- https://github.com/tlaplus/tlaplus/blob/eb5d554d405d891fdc98cc448184df50ee7e2e24/tlatools/customBuild.xml#L486-L512

will62794:
Thanks for the review. I will work on addressing your comments, adding more test coverage, and look into adding a benchmark for multi-threaded simulation.

lemmy:
In case you want the benchmarks to run with different thread counts: http://hg.openjdk.java.net/code-tools/jmh/file/3769055ad883/jmh-core/src/main/java/org/openjdk/jmh/runner/options/OptionsBuilder.java#l296

lemmy:
You might want to consider using the high-level [CompletableFuture](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html) to handle concurrency.

tlatools/src/tlc2/tool/Simulator.java
will62794: Done.

tlatools/test/tlc2/tool/simulation/SimulatorMultiThreadTest.java
will62794: :)

tlatools/src/tlc2/tool/Simulator.java
```diff
@@ -103,41 +114,303 @@ public Simulator(String specFile, String configFile, String traceFile, boolean d
...
+		public void kill() {
+			this.running.getAndSet(false);
+		}
+
+		/**
+		 * The main worker loop. Continually generates random traces until there is an
+		 * error or until we hit the trace count limit.
+		 */
+		public final void run() {
+			this.running.set(true);
```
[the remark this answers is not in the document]
will62794: Yep, sure is. Altered the cancellation mechanism slightly.

tlatools/src/tlc2/tool/Simulator.java
```diff
@@ -103,41 +114,303 @@ public Simulator(String specFile, String configFile, String traceFile, boolean d
...
+		 *  b) Randomly pick an action to generate the successor states (more than 1) to the current initial state. 
+		 *  c) Check all of the generated successors for their validity. 
+		 *  d) Randomly pick a generated successor and make it the new current state.
+		 *
+		 * Returns true if the simulation worker should generate another random trace.
+		 * Returns false if it should stop generating traces and terminate.
+		 *
+		 * @param theInitStates
+		 */
+		private Boolean simulateRandomTrace() throws Exception {
```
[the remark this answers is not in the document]
will62794: Done.

tlatools/src/tlc2/tool/Simulator.java
will62794: Here are my thoughts on how the `-continue` flag should work in simulation mode. If `continue=false`, then on any error from a single worker, of any type, we terminate and kill all other workers. For `continue=true`, I think we should consider treating certain error varieties differently. I see two main error categories. 

The first error category indicates that there is likely something wrong with the way the specification is constructed. For example, a runtime evaluation error (e.g. a "type" error). I think that if one worker encounters this type of error, we should kill all workers and terminate, even if `continue=true`, since it's a reasonable indication that other workers might encounter similar issues with the spec. 

The second error category is normal invariant, temporal property violations. Those errors can obviously be "ignored" if `continue=true` i.e. we'll print some information about them but keep all workers running. 

Do you agree with this?

lemmy: Agreed.

By the way, for the first category lets use the term "error (call) stack" (see e.g. tlc2.tool.Tool.setCallStack()) and for the second one simply "error trace".

lemmy:
@will62794 Is this still work-in-progress? I saw your "Done" comments above but can't find the corresponding changes in the multithreaded_simulation branch.

will62794:
@lemmy Sorry for the confusion. I am in the habit or responding to comments as I address them locally. I am still working on a follow-up patch, though.

lemmy:
@will62794 Confusion is caused by my inexperience with the Github PR UI.

will62794:
@lemmy Do you know of a way to view code coverage reports from within Eclipse IDE? It may be helpful for me when writing unit tests.

will62794:
@lemmy I am wondering if the coverage mechanism used in the Simulator is thread-safe. Currently we appear to update the `astCounts` member [here](https://github.com/tlaplus/tlaplus/blob/f4763d800472fd3f5c4ce1913ee01cf46bf2a6d4/tlatools/src/tlc2/tool/Simulator.java#L223) during simulation. Unless `ObjLongTable` is safe to use concurrently (from looking at the class it doesn't appear to be), I'm unsure whether this approach will work for multi-threaded simulation. We could always just disable coverage metrics when running with more than 1 worker I suppose.

lemmy:
@will62794 I'm in the process of overhauling/extending the coverage mechanism which will automatically be thread-safe. I'd suggest to implement multi-threaded simulation feature with the current coverage and merge the feature with the new coverage later.

tlatools/test/tlc2/tool/simulation/SimulatorTest.java
lemmy: If you move each @Test into its own class, f9f35f35b74289d85b2f75b09f37703281f984de makes it possible to run tests in a single VM.

tlatools/src/tlc2/tool/Simulator.java
lemmy: An "error trace" shows the (prefix of a) behavior that violates an invariant or a liveness property. An "error call stack" reports an error in evaluating expressions in the spec or model. The first one means the state space - leading to the error state - has been fully evaluated while the latter means state space exploration stopped when the expression failed. Both conditions should be reported separately.

will62794:
I'll work with the benchmark a bit more to make sure I understand the results. One easy improvement I noticed would be to parse and process the spec upfront to get rid of this constant overhead for every invocation. Also, is there a way to completely disable message printing within TLC? It would be nice if I could do this for the benchmarks.

lemmy:
You will have to hack tlc2.output.MP to suppress messages. The reparsing sounds more important though.

will62794:
I did more work on the benchmark and have some new results. Sorry for the delay. First, I changed the spec I was using to give a different state space structure. I wanted a tree-like state space for which I could control the maximum depth and branching factor. This gives me an easier way to understand roughly how many states there are for a given model. The number of states in the lowest depth of the tree i.e. `BranchingFactor ^ Depth` is a rough approximation. Here is an example of the new BenchmarkSpec state space for a tiny model:

simtest2_branch2_depth4

I also decided to measure the total throughput of "trace generation" for different worker counts, since that seems like the core metric that should scale well. I did this by adding a global trace counter that all workers update, and periodically logging this value. To measure throughput, I ran a Simulator for `D` minutes with a fixed worker count on an extremely big model (parameters of `BranchingFactor=8, Depth=14`), so that the simulator would not terminate in any reasonable amount of time. After running for `D` minutes, I would record the average trace generation count per second at that point. I did this for various worker counts on a Linux workstation, and for `D` values of 5 minutes and 10 minutes. The chart below plots both the 5 and 10 minute runs, but not the average:

| nWorkers | Traces/sec. (300 secs.) | Traces/sec. (600 secs.) | Avg.   | Scale Factor |
|----------|-------------------------|-------------------------|--------|--------------|
| 1        | 81000                   | 80000                   | 80500  | 1.00         |
| 2        | 147000                  | 153000                  | 150000 | 1.86         |
| 3        | 217000                  | 212000                  | 214500 | 2.66         |
| 4        | 290000                  | 282000                  | 286000 | 3.55         |
| 5        | 335000                  | 343000                  | 339000 | 4.21         |
| 6        | 379000                  | 375000                  | 377000 | 4.68         |
| 7        | 410000                  | 419000                  | 414500 | 5.15         |
| 8        | 439000                  | 447000                  | 443000 | 5.50         |
| 9        | 446000                  | 452000                  | 449000 | 5.58         |
| 10       | 470000                  | 469000                  | 469500 | 5.83         |
| 11       | 476000                  | 489000                  | 482500 | 5.99         |
| 12       | 497000                  | 478000                  | 487500 | 6.06         |

tracegenerationthroughput

The scaling of trace generation throughput looks reasonable, albeit somewhat sub-linear. As you said, I'm skeptical of the performance gains that hyper-threading brings for such a CPU intensive workload. The Linux workstation I ran the benchmark on has 6 cores, and 2 (hyper) threads per core. I suspect this may explain the sub-linear scaling beyond 6 workers. Here is a summarized version of its `lscpu` output:

```
Architecture:          x86_64
CPU(s):                12
On-line CPU(s) list:   0-11
Thread(s) per core:    2
Core(s) per socket:    6
Socket(s):             1
Model:                 62
Model name:            Intel(R) Core(TM) i7-4930K CPU @ 3.40GHz
```

I also tested the updated JMH benchmark, which simply measures the average time for a simulator to produce an invariant violation. I tried to run these benchmarks with a large enough model so that the search time outweighed the upfront parsing and spec processing time, startup time, etc. Here are the numbers from that benchmark, which I ran with model parameters of `BranchingFactor=3, Depth=15`. I ran with these JMH benchmark arguments, and did two runs:

```
java -jar benchmarks.jar -r 0 -i 80 -f 1 -wi 2 SimulatorBenchmark
```

| nWorkers | Avg. Time 1 (secs) | Avg. Time 2 (secs) | Average | Scale Factor |
|----------|--------------------|--------------------|---------|--------------|
| 1        | 192                | 193                | 192.5   | 1.00         |
| 2        | 90                 | 92                 | 91      | 2.12         |
| 3        | 69                 | 69                 | 69      | 2.79         |
| 4        | 42                 | 42                 | 42      | 4.58         |
| 5        | 36                 | 38                 | 37      | 5.20         |
| 6        | 33                 | 34                 | 33.5    | 5.75         |
| 7        | 31                 | 32                 | 31.5    | 6.11         |
| 8        | 31                 | 32                 | 31.5    | 6.11         |

findinvariantbenchmark

I did play around with using Java Flight Recorder to profile the Simulator a bit, but I can't say I came away with definitive results on what may be an additional bottleneck or source of contention here. That may be due to my inexperience with it, though. I can continue to investigate that more if we want to. Would be interested to hear your thoughts on these scalability numbers.

tlatools/src/tlc2/tool/SimulationWorker.java
```diff
@@ -68,11 +68,12 @@
...
 	
-	// A counter that tracks the number of generated states. This counter may be
-	// shared among workers, so its count may be greater than the number of states
+	// A counter that tracks the number of generated states/traces. This counter may be
+	// shared among workers, so its count may be greater than the number of states/traces
 	// generated by this worker alone. It is the duty of this worker, though, to
-	// update it whenever it generates a new state.
+	// update it whenever it generates a new state or trace.
 	private final LongAdder numOfGenStates;
+	private final LongAdder numOfGenTraces;
```
will62794: Note: If we think it's useful enough, we could include this in the output of the Progress Reporter, in the form of average total traces generated per second.

tlatools/src/tlc2/tool/Simulator.java
```diff
@@ -252,9 +281,14 @@ public void simulate() throws Exception {
 		ArrayList<SimulationWorker> workers = new ArrayList<>();
 		Set<Integer> runningWorkers = new HashSet<>();
 		for (int i = 0; i < this.numWorkers; i++) {
-			SimulationWorker worker = new SimulationWorker(i, tool, initStates, this.workerResultQueue,
+			
+			// Give each worker its own Tool and liveness checker object.
+			Tool newTool = new Tool(specDir, specFile, configFile, resolver);
```
will62794: Note: I changed this to give every worker its own Tool, just to be safe in case Tool is not thread safe.

lemmy: Tool is already being used concurrently by exhaustive search. Thus, shouldn't be a problem to use a single instance.

tlatools/src/tlc2/tool/SimulationWorker.java
lemmy: Do you think traces/sec is useful for end-users?

lemmy:
@will62794 Thanks for the detailed scalability analysis. Your results look correct to me and I agree with your hypothesis about hyperthreading being the cause for the sublinear scalability. I have obtained similar results for other workloads. I do not think it is necessary, but you can obviously double check by running N parallel instances of TLC and comparing the numbers with the multithreaded one. I can also provide you with an AWS instance at no cost in case you want to analyze scalability on larger core counts. From my perspective though, once the code uses a single Tool instance, we should merge your changes to master.

By the way, are you aware of the [Grid5k spec](https://github.com/tlaplus/tlaplus/blob/master/general/performance/Grid5k/Grid5k.tla) we have been using for years run performance and scalability measurements? To analyze and visualize scalability results I have found [Guther's Universal Scalability Law](http://dx.doi.org/10.1007/978-3-540-31010-5) invaluable. Its advantage is that it quantifies both contention and coherence.There exists a similar named [R package](https://github.com/smoeding/usl) to generate beautiful graphs.

tlatools/src/tlc2/tool/SimulationWorker.java
will62794: I personally think that trace generation count is an interesting and useful metric for simulation mode, since generation of a trace is the relatively logical "unit of work" for each simulation worker. I'm fine with leaving it out of reporting for now, though. Are you ok if we leave the counter there if we want to add it to the metrics output later?

will62794:
@lemmy I changed the Simulator to once again share the Tool object among workers.

I would be happy to continue the scalability analysis further, but, like you said, the scalability demonstrated so far seems adequate. We may be able to squeeze more performance out, but that can be an ongoing/future effort. 

Thanks for the resources as well. I was not aware of the Grid5k spec, and I would be interested to test the multi-threaded simulator on it. In relation to Guther's law, I have previously come across [this summary](http://www.perfdynamics.com/Manifesto/USLscalability.html#tth_sEc1), which touches upon the same concepts you mention, but I haven't had time to study it thoroughly. It will certainly go on my reading list.

lemmy:
I will rebase will62794:multithreaded_simulation on top of today's tlaplus/master and commit/push to GitHub.

lemmy:
83559207381b08472f1d1dcf8bb07d08a7f9b089

will62794:
@lemmy I recently took some time to do some more scalability analysis of multi-threaded simulation mode on a `c5d.24xlarge` EC2 instance. Here is the full CPU info for the instance I used:
```
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              96
On-line CPU(s) list: 0-95
Thread(s) per core:  2
Core(s) per socket:  24
Socket(s):           2
NUMA node(s):        2
Vendor ID:           GenuineIntel
CPU family:          6
Model:               85
Model name:          Intel(R) Xeon(R) Platinum 8275CL CPU @ 3.00GHz
Stepping:            7
CPU MHz:             3281.157
BogoMIPS:            6000.00
Hypervisor vendor:   KVM
Virtualization type: full
L1d cache:           32K
L1i cache:           32K
L2 cache:            1024K
L3 cache:            36608K
NUMA node0 CPU(s):   0-23,48-71
NUMA node1 CPU(s):   24-47,72-95
Flags:               fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon rep_good nopl xtopology nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq monitor ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm abm 3dnowprefetch invpcid_single pti fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid mpx avx512f avx512dq rdseed adx smap clflushopt clwb avx512cd avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves ida arat pku ospke avx512_vnni
```

I wrote a simple benchmark that is stored [here](https://github.com/will62794/tlc-sim-benchmark). See [this script](https://github.com/will62794/tlc-sim-benchmark/blob/master/tlc-sim-test.sh) for the command line TLC invocation being used. I used the [Grid5k](https://github.com/tlaplus/tlaplus/blob/ea1d4cd252d1303f16b9fec07d3a8d1d36ddbeec/general/performance/Grid5k/Grid5k.tla) spec to test scalability of state generation speed for simulation mode. I ran TLC under simulation mode for a fixed time limit (60 seconds for this benchmark) and then measured the average states/min. before terminating the process. I did this for TLC worker counts from 1 up to 48, since there are 48 hardware cores on the EC2 instance. Here are the charted results:

Gnuplot

The blue line represents the results from the benchmark, and the red line indicates the ideal, linear scalability curve if we extrapolated the performance with 1 TLC worker. I expected simulation mode to scale much better than this since workers don't need to contend on any shared data structures. I am curious to investigate this more to understand why we are not achieving close to linear scalability. I wonder if you had any thoughts on certain data structures or classes that might be unintentionally causing contention between simulation workers. At some point I would like do more in-depth profiling to identify the bottlenecks here.

lemmy:
Thanks for looking into simulation's scalability.  The blue line does indeed indicate unintended contention.  In the past, [jfr](https://en.wikipedia.org/wiki/JDK_Flight_Recorder) has proven to be useful to diagnose contention (e.g. just two days ago it helped me to spot https://github.com/tlaplus/tlaplus/commit/d1c93e88709bdd3ed073839d7fa6184162b7e86c). The script ```general/performance/measure.sh``` shows how to set up jfr and a recent version of JMC (to analyze a recording) is provided by [Azul](https://jp.azul.com/products/zulu-mission-control/).

In addition to recording flight recordings to capture JVM (and [TLC](https://github.com/tlaplus/tlaplus/tree/master/tlatools/jfr)) internals, I found [Neil](https://en.wikipedia.org/wiki/Neil_J._Gunther)'s [Universal scalability law]('http://www.perfdynamics.com/Manifesto/USLscalability.html) invaluable to faithfully study and predict scalability: It distinguishes between contention and coherence.  Reporting worked best for me with the [usl package](https://cran.r-project.org/web/packages/usl/index.html) for R (see below).

Running TLC for only a minute might contribute to the poor line because it records a lot of warmup noise at the JVM level.  In future benchmarks, I would suggest to run TLC longer and experiment with different values for const ```L``` of Grid5k.tla.
 
Note that I made [progress](https://github.com/lemmy/tlaplus/commits/mku-pagequeue) on scaling BFS too.

```R
if(!require(usl)){
    ## Install usl (only needed once).
    install.packages("usl")
}
## Load usl library (needed upon each R invocation).
library(usl)

## CSV data
df.2016_10_29_x32 = read.table(sep=",", text="
Implementation,Processes,Insertions,Duration,Memory,LoadFactor,Locks,TblCnt,JunitTime,NUMABalancing,PageDefrag,Govenor,Test
MSB,001,17179869184,14909024,34359738368,0.5,256,2147483639,14915.836,1,[never],powersave,FingerPrintGenerator
MSB,001,17179869184,15187436,34359738368,0.5,256,2147483639,15194.144,1,[never],powersave,FingerPrintGenerator
MSB,001,17179869184,15464179,34359738368,0.5,256,2147483639,15470.225,1,[never],powersave,FingerPrintGenerator
Off,001,8589934592,5752717,34359738368,0.25,256,2147483639,5759.393,1,[never],powersave,FingerPrintGenerator
Off,002,17179869184,10160424,34359738368,0.5,512,2147483639,10166.944,1,
...
", header=TRUE)

## Calculate duration based on recorded $Insertions and $Duration.
## http://stackoverflow.com/questions/13013231/how-can-i-divide-one-column-of-a-data-frame-through-another
df.2016_10_29_x32 <- transform(df.2016_10_29_x32, Throughput = Insertions / Duration)

## Group by Implementation (poor-man's version)
off.2016_10_29_x32 <- subset(df.2016_10_29_x32, Implementation == "Off" & TblCnt == "68719476736" & LoadFactor=="0.5")
msb.2016_10_29_x32 <- subset(df.2016_10_29_x32, Implementation == "MSB" & TblCnt == "2147483639" & LoadFactor=="0.5")

## We don't average multiple runs. usl will do this for us.
#df.2016_10_29_x32 = aggregate(Throughput ~ Processes, data=df.2016_10_29_x32, FUN = function(x) c(M=mean(x)))

## Calculate speedup over serial run.
off.2016_10_29_x32$Speedup <- off.2016_10_29_x32$Throughput / aggregate(Throughput ~ Processes, data=subset(off.2016_10_29_x32, Processes == "1"), FUN = function(x) c(M=mean(x)))$Throughput
msb.2016_10_29_x32$Speedup <- msb.2016_10_29_x32$Throughput / aggregate(Throughput ~ Processes, data=subset(msb.2016_10_29_x32, Processes == "1"), FUN = function(x) c(M=mean(x)))$Throughput

## Have usl do its thing.
## https://cran.r-project.org/web/packages/usl/vignettes/usl.pdf
## https://cran.r-project.org/web/packages/usl/usl.pdf
## nlxb: Nash variant of Marquardt nonlinear least squares solution via qr linear solver.
## https://cran.r-project.org/web/packages/nlmrt/nlmrt.pdf
usl.off.2016_10_29_x32<- usl(Speedup ~ Processes, off.2016_10_29_x32, method = "nlxb")
usl.msb.2016_10_29_x32 <- usl(Speedup ~ Processes, msb.2016_10_29_x32, method = "nlxb")
```

lemmy:
What just occured to me: Is it possible that progress reporting blocks the worker threads?

lemmy:
[A 3 minute Java Flight Recording](http://tla.msr-inria.inria.fr/kuppe/simulation.jfr) on a 16 core Azure instance.  It doesn't point at contention or coherence (workers are CPU bound).

PR #214, opened 2018-12-08
Terminate TLC with a non-successful exit code when it fails to parse parameters.

edahlgren:
Note: The "flowpayau" identity is a Github.com bug and should (hopefully) be resolved shortly. The committer is "edahlgren".

--- From commit message:

Includes minimal, non-semantic changes to main:

* Adds comments to all major blocks.
* Organizes mail sending code logically together, so that it's clear
  that it's not a precondition on calling TLC.process() or setting
  up resolvers.

Categories: Bug, TLC

PR #215, opened 2018-12-08
Make it easier to run TLC standalone.

edahlgren:
Note: The "flowpayau" identity is a Github.com bug and should (hopefully) be resolved shortly. The committer is "edahlgren".

--- From commit message:

TLC terminates with a filesystem lookup error in SANY parsing code when
given a path to an arbitrary spec file on the user's local filesystem.

TLC is fully capable of being an easy to use, standalone tool. This change
fixes these filesystem lookup errors so that users of TLC can intuitively
use any spec file that exists on their filesystem.

---

Background:

* When a spec file is given to TLC explicitly (not part of a .jar file),
  a SimpleFilenameToStream resolver is used to lookup files on a filesystem
  (see tool/SpecObj.java).

* Code paths like tool/Simulator.java -> tool/Tool.java -> tool/Spec.java
  tool/SpecObj.java parse the spec directory from the spec file, but then
  subsequently ignore it (see specDir as dead code in tool/Spec.java).

* As a result, tool/SpecObj.java can only lookup specs from paths included
  in the filesystem resolver. It also means that in many cases the specDir
  is passed to Tool invocations for no reason.

Change details:

* This change parses the spec directory from the spec file and adds it as
  a library path in the resolver before TLC.process() is called. This means
  that SpecObj.java will look for the spec file in the spec directory first,
  before trying the "standard" module directories built into the resolver.
  This is what we want.

* The change leaves the dead code in tool/Spec.java to remain minimal. A
  follow-up change will remove the dead code and clean up initializers.

Testing:

I don't know how to create an automated integration test for this change. It
was minimally validated by running:

  $ cd tlatools
  $ ant -f customBuild.xml compile dist
  $ java -cp dist/tla2tools.jar tlc2.TLC -deadlock -simulate -config test-model/DieHard.cfg test-model/DieHard.tla

Categories: Bug, TLC, Standalone, Interface

### DIFF
--- a/tlatools/src/tlc2/TLC.java
+++ b/tlatools/src/tlc2/TLC.java
@@ -206,37 +206,53 @@ public class TLC
     {
         TLC tlc = new TLC();
 
-        // handle parameters
-        if (tlc.handleParameters(args))
-        {
-        	final MailSender ms = new MailSender();
-        	if (MODEL_PART_OF_JAR) {
-        		tlc.setResolver(new InJarFilenameToStream(ModelInJar.PATH));
-        	} else {
-        		tlc.setResolver(new SimpleFilenameToStream());
-        	}
-        	ms.setModelName(tlc.getModelName());
-        	ms.setSpecName(tlc.getSpecName());
-
-            // call the actual processing method
-            tlc.process();
-
-            // Send logged output by email
-            boolean success = ms.send(tlc.getModuleFiles());
-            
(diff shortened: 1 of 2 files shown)


edahlgren:
@lemmy Same as https://github.com/tlaplus/tlaplus/pull/214. PTAL, thanks.

lemmy:
b905a86e31714bf5002290141b51b41ffe280c8e

PR #238, opened 2019-02-22
Do not set queue size to distinct states

philipmw:
This is confusing and plain wrong.

The Toolbox already has support for rendering negative numbers as "--".

### DIFF
--- a/org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/output/data/StateSpaceInformationItem.java
+++ b/org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/output/data/StateSpaceInformationItem.java
@@ -56,7 +56,7 @@ public class StateSpaceInformationItem
 	}
 	
 	private StateSpaceInformationItem(Date time, long foundStates, long distinctStates) {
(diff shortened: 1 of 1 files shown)


org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/output/data/StateSpaceInformationItem.java
```diff
@@ -56,15 +56,15 @@ private StateSpaceInformationItem(Date time, long diameter,
 	}
 	
 	private StateSpaceInformationItem(Date time, long foundStates, long distinctStates) {
-		this(time, 0, foundStates, distinctStates, distinctStates, 0, 0);
+		this(time, 0, foundStates, distinctStates, -1, 0, 0);
```
lemmy: How does this manifest in the UI? Does it make sense to add a test case?

philipmw:
After thinking some more about this change, I now think that the original behavior is correct. If we found 10 initial states, those 10 states go into the queue. It now seems correct to say that initial states = queue size. So, I am closing this pull request. Sorry for crying wolf.

lemmy:
See 353583582ab0768aee49a1224c9da1bac5b54269

philipmw:
Perfect. Thanks.

PR #242, opened 2019-03-03
Fixing issues caught by ErrorProne.

tautomaton:
Using ErrorProne (https://errorprone.info) to catch issues when building tlatools.


lemmy:
Thanks!

PR #248, opened 2019-03-08
Javacc4 config

Henne:
These 28 patches update the input file and the in-use generated source files of the Configuration Parser from a JavaCC 0.7pre6 version to JavaCC 4.0, which is also used by the TLA parser.

The changes have been carefully split into smaller commits,
so that the changes may be easier reproduced by humans and the diff tool.

From now on the generated files of config.jj don't need any manual modification
and can simply be copied to the TLATools.


Henne:
Hm...

Mac failed with org.lamport.tla.toolbox.product.uitest and

Windows and Linux did not even try to build TLATools:

[INFO] ------------------------------------------------------------------------
[INFO] Skipping org.lamport.tlatools
[INFO] This project has been banned from the build due to previous failures.
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] Skipping org.lamport.tlatools.api
[INFO] This project has been banned from the build due to previous failures.
[INFO] ------------------------------------------------------------------------
[INFO]

lemmy:
@Henne Thanks for the patches.  How did you make sure that JavaCC generates an identical parser to the one we currently use?

By the way, do not worry about the Azure Pipelines builds. I will probably disable them because the maven build seems effectively incompatible with the Azure Pipeline free offering.

Henne:
I've looked at the source files of currently used parser and generated a new parser with JavaCC 4.0.
Then I compared the new one to the current one with git diff and made small commits for each set of changes which belonged together.
After each commit I recompiled and checked that DieHard.tla works and the results have not changed.

tlatools/src/tla2sany/configuration/ConfigurationTokenManager.java
```diff
@@ -1024,9 +1024,7 @@ static public void ReInit(SimpleCharStream stream)
 {
    jjmatchedPos = jjnewStateCnt = 0;
    curLexState = defaultLexState;
-   // Bug: The following line used to read "input_stream = stream; but this bug was corrected
-   // by DRJ.  The bug should also be corrected in the .jcc file, but isn't.
-   input_stream = null;
+   input_stream = stream;
```
lemmy: Please clarify, the bug has not been fixed in the .jcc file?  If not, why?

lemmy:
Can you please rebase your branch on top the current master? Commit e3807eb781cbfa5555152ea3415d6fa027a05656 will help us better check for regressions.

tlatools/src/tla2sany/configuration/ConfigurationTokenManager.java
Henne: It is good that you have asked for that.
I have thought about that and I'll try to explain it:

With a static parser generated with JavaCC you have a static object,
which is called Configuration here. If you use that static parser on multiple input streams,
then you have to call one of the ReInit() methods before each run,
which get the new input stream as an argument.
That means you need to know, at this point of time, the stream which should be parsed.

The current implementation woks like this:
The only object which uses the Configuration object is in tla2sany/drivers/SANY.java,
which calls frontEndMain() which calls frontEndInitialize() for each of maybe multiple TLA-files.
In frontEntInitialize(), which is the only method which uses Configuration,
there is first a call to the Configuration.ReInit() and then a call to Configuration.load().

This ReInit() method from DRJ is mostly a copy of the one provided by JavaCC,
but it has no parameters and it is altered that way that it uses the static member input,
which is not defined in the first run. That is an error in the current code.
Notice that the stream of the input is not known at this point of time.
Also it calls the ReInit() methods of ASCII_CharStream/SimpleCharStream and ConfigurationTokenManager

Then Configuration.load() is called, which:

- determinates the source of input (the String defaultconfig or the File "config.src"),
  creates a Stream and saves that in the static member input,
- instantiates a new !!! static !!! parser with the correct input stream,
- parses that stream

If there are multiple TLA files, then the static member input contains the input Stream of the run before, which is then used in the next call to Configuration.ReInit().

**Conclusion**:
I hope you see that this handwritten ReInit() method from DRJ is the source of these bugs,
which than required that fixes in the other files.

BTW, in the TLA-parser there is also the same code in SimpleCharStream and TLAplusParserTokenManager which has worked for the same time and did not need the bugfixes there.

**How to proceed**:
If you accept all patches until the ReInit() removal patches,
then I would rebase them, so you can pull these.
Then I would make another branch for the patches which will result in a sane static parser,
the way it is thought by the JavaCC developers.
After that the transition from static to non-static can be done seperately.

Henne:
Thats good, but imho a better check would be to run tla2sany with multiple TLA files.
  `java tla2sany.SANY spec1.tla spec2.tla ... `

Since that would trigger a bug, if one exists.

lemmy:
Why do you think invoking SANY directly on a set of specs will be a better check compared to executing TLC on the same specs?  TLC internally calls SANY and additionally checks the semantics by evaluating the semantic graph.

Henne:
Sorry, checking with tla2sany is not better, since it checks another testcase.

TLC calls the configuration parser only once per run.

Tla2sany calls the configuration parser for each spec.
For each spec the configuration parser needs to be reinitialized,
so bugs or unexpected behavior could occur there which would not be seen with TLC.

I've made another [pull request #254](https://github.com/tlaplus/tlaplus/pull/254) for a small patch, which clarifies the removal of the DRJ-Bugfixes.
If you accept it, I'll rebase the other patches.

lemmy:
Please see https://github.com/tlaplus/tlaplus/pull/254#issuecomment-473422610

PR #254, opened 2019-03-14
Configuration Parser: use the static parser the right way

Henne:
Remove a missuse of the static configuration parser.
Before the parser is called the source of the input must
be specified and passed to the ReInit() method to
(re-)initialize the internal members of the parser.
Thats the way how the JavaCC devs want us to use the parser
in the static version.

This patch removes the handwritten ReInit() method
which was used to do the reinitialization without
knowledge of the input and the two resulting
bugfixes which made that ReInit() method work,
but broke the way it should be used.

The code is now restored the way it is generated by JavaCC.

[Refactor][SANY]

### DIFF
--- a/tlatools/src/tla2sany/configuration/ASCII_CharStream.java
+++ b/tlatools/src/tla2sany/configuration/ASCII_CharStream.java
@@ -257,9 +257,7 @@ public final class ASCII_CharStream
   static public void ReInit(java.io.Reader dstream, int startline,
   int startcolumn, int buffersize)
   {
-    // Bug: The following line used to read "inputStream = dstream;" but this bug was
-    // corrected by DRJ.  The correction should be in the .jcc copy, but isn't.
-    inputStream = null;
+    inputStream = dstream;
     line = startline;
     column = startcolumn - 1;
 
--- a/tlatools/src/tla2sany/configuration/Configuration.java
+++ b/tlatools/src/tla2sany/configuration/Configuration.java
@@ -4,8 +4,7 @@
 /***************************************************************************
 * This file was originally created by running javacc on the grammar file   *
 * config.jj that specifies the parsing of the string                       *
-* ConfigConstants.defaultConfig.  The current file has since been          *
-* modified by DRJ (David Jefferson).                                       *
+* ConfigConstants.defaultConfig.                                           *
 *                                                                          *
 * This file was modified by LL on 13 May 2008 to replace calls to          *
 * System.out and System.err by calls to ToolIO.out and ToolIO.err.         *
@@ -42,7 +41,7 @@ public final class Configuration implements ConfigConstants {
     /***********************************************************************
     * Called from drivers/SANY.java                                        *
     ***********************************************************************/
-    Configuration Parser;
+
     try {
       errors = errs;
       File source = new File( "config.src" );
@@ -56,10 +55,10 @@ public final class Configuration implements ConfigConstants {
         input = new java.io.StringReader( defaultConfig );
         origin = " from defaults.";
       }
-      Parser = new Configuration( input );
+     Configuration.ReInit(input);
 
       try {
-        Parser.ConfigurationUnit();
+        Configuration.ConfigurationUnit();
 //      Operators.printTable();
       } catch (ParseException e) {
         errors.addAbort(Location.nullLoc,"\nConfiguration Parser:  Encountered errors during parse.  " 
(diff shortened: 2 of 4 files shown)


lemmy:
@Henne What's the overarching goal of your work on the config parser?

Henne:
The overarching goal is **maintainability** .

It is hardly replicable what happened exactly in the time before TLATools were in a git repository,
but but the comments and the documentation say that the Configuration parser and the TLAPlusParser were generated and manually modified.

This means exactly **forking generated code** which is bad idea.

The config parser is more than 2% and the TLAPlus parser is more than 16% ,
so together nearly **20% of the TLAPlus code** is generated code by JavaCC.
To maintain this and keep control over 1/5 of the TLATools code requires a good strategy.

My offer is:
- use the parsers the way it was meant by the JavaCC developers,
- bring the input files config.jj and tla+.jj in sync again with the used generated files without breaking anything,
- let the JavaCC developers handle the bug fixing in the parsers by using the input files on newer versions of JavaCC and apply the changes to TLATools

This would also clear some problems addressed by Lamport in this [Note](https://github.com/tlaplus/tlaplus/blob/master/tlatools/javacc/README/methods-sany1.txt) which is from 2009.



**Now to some arguments for the patch:**

----

The ReInit() method in SimpleCharStream (which is ASCII_CharStream in the old configuration parser)
is generated by JavaCC from that [Template](https://github.com/javacc/javacc/blob/e32a08629519cd7758cced211fcadfa43a6aeac5/src/main/resources/templates/SimpleCharStream.template#L335). Here dstream is assigned to inputStream.

The ReInit() method in ConfigurationTokenManager is generated by JavaCC from that [Template](https://github.com/javacc/javacc/blob/e32a08629519cd7758cced211fcadfa43a6aeac5/src/main/resources/templates/TokenManagerBoilerPlateMethods.template#L58). Here dstream is also assigned to inputStream.

**Heuristic Argument :** If this would have been a bug in the JavaCC templates, then it should have been fixed there now. But it is not fixed, so it seems that somethings wrong with the TLATools code.

----

**Wrong Suggestion:**
When generating a parser with JavaCC, the code from the input file is copied to Configuration.java
in our case. The other files are generated from templates.
It is not possible to change code in SimpleCharStream and ConfigurationTokenManager by modifying the input file. This is a contradiction to the comments of DRJ in SimpleCharStream and Configuration TokenManager.

----

**My interpretation of what happened in the past:**
- the code was generated,
- DRJ has not understood or known that the source of the input must be known when
  the Configuration-object (the parser) is initialized,
- DRJ wrote that ReInit() method in Configuration by copying the code of the generated ReInit() method above in the sourcecode (which is not good style),
- DRJ noticed it doesn't work and changed the other two methods in SimpleCharStream and ConfigurationTokenManager claiming they are buggy.

-------

Another argument that this ReInit() from DRJ is buggy is the following fact:
**Init**:
- a Configuration object is created at starttime of the program and I assume the static member _input_ = _null_
- then for every tla-file in set of files first Configuration.ReInit() is run and then Configuration.load(),
  See [frontEndInitialize()](https://github.com/tlaplus/tlaplus/blob/b7830732edeaecb6c30a4f1589bdef101e541aad/tlatools/src/tla2sany/drivers/SANY.java#L160).

**Run on the first tla-file**:
- the first ReInit() uses the static member _input_, which is null as I assume, and calls the ReInit(input)-method of the SimpleCharStream object,

- then load() is called, which determines the source of the input and sets the static member _input_ to the input stream
- calls the constructor with _input_ as argument,  which initializes the static parser to the
  correct input stream

**Run on the second tla-file**:
- ReInit() is called and the member _input_ contains a reference to the input stream of the **first** tla-file,
  which is passed with the ReInit(input) to SimpleCharStream.
  **That is absurd.**

-------

**Heuristic Argument 2:**
If it is not intuitive how the configuration parser works together with the TLATools,
then something is wrong and should be cleared.

-------

In case of question please feel free to ask.

lemmy:
What you write makes sense and I'm inclined to agree with you.  However, SANY is part of a formal verification tool whose users write safety critical software.  As such, the general rule is: Don't fix it if it ain't broken.  Maintainability is not a goal in itself but a means to achieve actual goals.  Improving maintainability for a piece of code which hasn't changed in a decade provides very little benefit compared to the risks of such a refactoring (especially in the light of the fact that the parser is central to the tools).

By the way, you shouldn't speculate about DJR's motifs.  There might have been good reasons for the fix (which unfortunately have not been documented).

This rejection has probably discouraged you from future contributions, but please believe me that this wasn't my intent.  There are [countless areas](https://github.com/tlaplus/tlaplus/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) that desperately needs improvement.  Please [discuss future ideas first](https://github.com/tlaplus/tlaplus/blob/master/CONTRIBUTING.md) before starting the actual work.

-----

If code gets changed/fixed, please always update all corresponding documentation (methods.txt & methods-sany1.txt).  Removing DJR's bugfix breaks the following JUnit tests: 

tlc2.tool.simulation.SimulationWorkerTest
tlc2.tool.simulation.SimulatorMultiThreadTest
tlc2.tool.simulation.SimulatorTest
pcal.OptionalSemicolonTest

Additionally, the Toolbox pops up an error message on "Parse Spec" without DJR's bugfix.  FWIW this commit does not introduce any regressions. I took me approximately 3h to understand the existing code, review parts of this commit and the larger PC https://github.com/tlaplus/tlaplus/pull/248

lemmy:
One more remark:  If we find that we in the future have to work on the configuration parser to implement a feature, we would be more than happy to revisit these changes.

PR #307, opened 2019-05-18
Add DEVELOPING.md

japgolly:


lemmy:
Two commits look good.  I'd suggest to indicate that 'entr' is Linux only (if it is).  The maven will also check the Java dependency and complain unless Java 11 (not duplicated in customBuild.xml).  The maven build can skip the TLC but not the Toolbox tests with '-Dtest.skip=true'.

@quaeler Anything else?

quaeler:
`entr` has a macOS Homebrew port - maybe state "non-Windows"?

Otherwise, nothing to add beyond what you've mentioned. Thanks @japgolly

japgolly:
Alrighty, I've made two changes as per feedback:

* move Java11 note into tlatools section as the main Maven build asserts this itself
* note `entr` is for LInux & Mac

lemmy:
Squashed and slightly amended: https://github.com/tlaplus/tlaplus/commit/a616df45578570395dbae3abf1ae7680ca1a403e

Thanks.

Issue #191, opened 2018-09-13
Animation Support for Traces


will62794:
## Goals and Motivation
The [Animation](https://github.com/will62794/tlaplus_animation) module is a TLA+ module that provides operators that serve as the basis of a simple vector graphics description language, as well as the ability to animate TLC execution traces. The current instructions for using this module are described in the linked Github repository. The goal is to integrate the functionality of the Animation module with TLC and the Toolbox, so that animation traces can be generated without impacting model checking performance, and so that animations can be easily viewed directly inside the toolbox. An animation trace would be automatically generated by TLC and opened in a web viewer pane in the Toolbox when a trace violates an invariant, analogous to how the Error Trace Explorer works. Using TLA+ itself as a graphics description language will be easier and more natural for existing users of TLA+. They will not need to learn a new language or framework to create animations of their specifications.

## Design Ideas

We can add a new flag (e.g. `--animationSpec`) to TLC that takes as a parameter a user's "animated spec" expression, call it _AnimSpec_. _AnimSpec_ contains the same variables as an original spec, but adds auxiliary variables that track animation related state. It is not necessary for this animation related state to be tracked during the entire model checking procedure. Currently, when TLC finds a violating error trace, it does not have access to all states of that trace. It only has the fingerprints of each state. For error trace exploration, TLC already re-constructs the full state of each step in an error trace. If a user wants to view the animated error trace, then during this re-construction process, TLC can use _AnimSpec_ instead of the original specification. This will produce the animation for that specific error trace. Since _AnimSpec_ adds auxiliary variables, we will want these variables to be ignored by using a "view", so that the state fingerprints generated during animated trace generation are identical to the original fingerprints. Once TLC generates the animated trace, the generated HTML/SVG string can be copied into a standard HTML template, which can then be opened in a web viewer pane in the toolbox.

Note that If the evaluation of _AnimSpec_ fails during trace re-construction, TLC can always fall back to the standard _Spec_, print an error message, and not produce or open the animation.

will62794:
[...]

Issue #265, opened 2019-03-28
Support named expressions in Error Trace Exploration


lemmy:
The actual expression as "Name" in the error trace is hard to read:

![Screenshot from 2019-03-27 22-29-54](https://user-images.githubusercontent.com/88777/55132498-dbf97a80-50df-11e9-9b2b-f21a061a8e00.png)

One possibility is to explicitly name expressions, s.t. "foo" gets used as a name:
```tla
foo == IF variable = 42 THEN "blah" ELSE "blub"
```

Workaround is obviously to define the operator in the spec and references the op in the trace explorer.  This falls short however, if op references another operator defined solely in the trace explorer.

lemmy:
![TLAGitHubIssue265](https://user-images.githubusercontent.com/88777/55184925-88734500-5150-11e9-9c1b-35a91757c21b.gif)

lemmy:
TODO: Visually separating by using italics for named expressions in Error-Trace table.

lemmy:
Map SANY error locations back to Trace Explorer locations (9ae65abe9131abb351b28bd84d4965e8d933889b):

![SanyLocation2TraceExplorerLocation](https://user-images.githubusercontent.com/88777/55581262-4195ca00-56d1-11e9-9dad-b69f3f3809d1.png)

PR #310, opened 2019-05-19
Initial implementation of the trace animation feature

will62794:
This patch (for issue #191) adds initial support for animating TLC traces by providing a 'view' expression that produces a visual representation of a TLA+ state. The trace animator re-uses most of the same functionality from the trace explorer, except that it evaluates the 'view' expression at each state of a trace instead of user defined expressions. Each frame of an animation is an SVG element. These are then stitched together into an interactive animation and displayed in an HTML viewer in the Toolbox.

This patch lays the groundwork for the animation feature and I think it should be fairly complete. Having said that, it took me a while to understand how the Toolbox deals with model launching, TLC process output, and wiring all of these things together, so there may be issues. I will note that [this](https://github.com/will62794/tlaplus/blob/a579464131a087e0643e6cb48b57f04f436eda95/general/docs/TLAToolboxTLCDocumentation.txt) documentation was quite well-written and helpful. 

It would be reasonable to consider this a beta feature at this point, especially since this patch noticeably omits additional test coverage for the animation feature. I can add that but right now I am constrained in how much more time I will have to work on this in the very near future. I may be able to do follow up work starting in mid or late June. Based on what you think about the state of this patch, we can decide whether it makes sense to include the animation feature in beta as a part of the upcoming Toolbox release (1.5.8), or just hold off until the next release. 

For a quick demo of how to use the feature, here is a [link](https://www.dropbox.com/s/azyhzxtcjoj3f55/tla_animation_demo_trimmed_1080p.mov?dl=0) to the screencast you requested. I'm sure there may be remaining open questions about this feature, like whether or not we want an "Animate" button to be the way users launch trace animation, and how we eventually want users to import the "Animation.tla" module, but all of these questions will hopefully be easier to settle once the foundational implementation is complete, which I think this patch achieves.

### DIFF
--- a/org.lamport.tla.toolbox.tool.tlc.ui/plugin.xml
+++ b/org.lamport.tla.toolbox.tool.tlc.ui/plugin.xml
@@ -827,6 +827,15 @@ Console Facory
             name="TLC Errors"
             restorable="true">
       </view>
+      <!-- TODO: Add a custom icon for the TraceAnimationView. -->
+      <view
+            allowMultiple="true"
+            category="org.lamport.tla.toolbox.category"
+            class="org.lamport.tla.toolbox.tool.tlc.ui.view.TraceAnimationView"
+            id="toolbox.tool.tlc.view.TraceAnimationView"
+            name="Trace Animation"
+            restorable="true">
+      </view>
    </extension>
    <extension
          point="org.eclipse.help.contexts">
--- a/org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/output/LogFileReader.java
+++ b/org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/output/LogFileReader.java
@@ -54,7 +54,7 @@ public class LogFileReader {
 	public LogFileReader(Model model, IFile aLogFile, boolean isTraceExplorerLogFile) {
 		this.logFile = new File(aLogFile.getLocation().toOSString());
 		this.parser = new TagBasedTLCOutputIncrementalParser(model, ITLCOutputSource.PRIO_LOW, isTraceExplorerLogFile,
-				TagBasedTLCOutputIncrementalParser.Mode.BATCH, this.logFile.length());
+				false /* isTraceAnimation */, TagBasedTLCOutputIncrementalParser.Mode.BATCH, this.logFile.length());
 	}
 
     /**
--- a/org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/output/ParsingTLCOutputSink.java
+++ b/org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/output/ParsingTLCOutputSink.java
@@ -3,6 +3,7 @@ package org.lamport.tla.toolbox.tool.tlc.output;
 import org.eclipse.jface.text.BadLocationException;
 import org.lamport.tla.toolbox.tool.tlc.model.Model;
 import org.lamport.tla.toolbox.tool.tlc.output.source.ITLCOutputSource;
+import org.lamport.tla.toolbox.tool.tlc.output.source.TLCOutputSourceRegistry;
 import org.lamport.tla.toolbox.tool.tlc.output.source.TagBasedTLCOutputIncrementalParser;
 import org.lamport.tla.toolbox.tool.tlc.ui.TLCUIActivator;
 
@@ -58,8 +59,16 @@ public class ParsingTLCOutputSink implements IProcessOutputSink
     public void initializeSink(Model model, int sinkType)
     {
         boolean isTraceExploration = sinkType == IProcessOutputSink.TYPE_TRACE_EXPLORE;
+        boolean isTraceAnimation = sinkType == IProcessOutputSink.TYPE_TRACE_ANIMATE;
         // parser = new TLCOutputIncrementalParser(processName, ITLCOutputSource.PRIO_HIGH);
-        parser = new TagBasedTLCOutputIncrementalParser(model, ITLCOutputSource.PRIO_HIGH, isTraceExploration);
+        if(isTraceAnimation) {
+        	// Adding this line constructs a new TraceAnimatorDataProvider and registers in the map of providers in
+        	// the source registry. Adding it appears to fix the issue where the TraceAnimatorDataProvider was not getting alerted
+        	// when the TraceExplorerJob completed.
+        	TLCOutputSourceRegistry.getTraceAnimateSourceRegistry().getProvider(model);
+        }
+        
+        parser = new TagBasedTLCOutputIncrementalParser(model, ITLCOutputSource.PRIO_LOW, isTraceExploration, isTraceAnimation);
     }
 
     /* (non-Javadoc)
--- a/org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/output/data/TraceAnimatorDataProvider.java
+++ b/org.lamport.tla.toolbox.tool.tlc.ui/src/org/lamport/tla/toolbox/tool/tlc/output/data/TraceAnimatorDataProvider.java
@@ -26,8 +26,10 @@
 package org.lamport.tla.toolbox.tool.tlc.output.data;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.runtime.Assert;
 import org.lamport.tla.toolbox.tool.tlc.model.Model;
 import org.lamport.tla.toolbox.tool.tlc.output.source.TLCOutputSourceRegistry;
 import org.lamport.tla.toolbox.tool.tlc.ui.view.TraceAnimationView;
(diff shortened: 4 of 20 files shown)


lemmy:
Some observations:

* Is ```Animation.tla``` supposed to be shipped as part of the Toolbox?
* If yes, ```Animation``` should adopt operator naming pattern (prefix of ```_TAGroup```, ```_TARect```, ... similar to ```_TEPosition``` and ```_TETrace``` to avoid name clashes
* Do you want to let users eventually specify the ```View````as a trace explorer expression?
  * Would also free users from EXTENDing Animation (TA.tla (?) can extend it)
* The "Animate" button does not follow the new enabledment paradigm of the trace explorer (@quaeler)
* I remember seeing controls to step through an animation in your 2018 presentation. Do you intend to add them later?

will62794:
Hello @lemmy. I responsed to your observations below. I also have some more thoughts about the state of this patch and would like to get your feedback. This patch is quite barebones i.e. as I mentioned before, a minimum viable product, for a fairly weak definition of "minimum". There are several things that are rough around the edges that still need to be cleaned up. These include:

1. Adequate test coverage of the animation feature
2. Integrating the `Animation.tla` module directly into the Toolbox to obviate the need for a manual download
3. The animation viewer "flickers" right now when stepping to a new frame, which should be fixed.
4. For some reason, evaluating the view expression is extremely slow in TLC, to the point of making the feature almost unusable. I have not investigated this thoroughly but I think it may have something do with the way the trace explorer builds up the next state relation. I think this will be fixable, but it will require a bit of exploration.
5. Enhancing the UI of the trace animation view
	- Add info text that indicated what frame number is currently being viewed
	- Add a custom icon to the trace animation view (an icon [idea](https://www.flaticon.com/premium-icon/film_1146203#term=film&page=1&position=2)).
	- Proper control buttons in the animation viewer for stepping through the trace

I am bringing this up because I will not have significant time to invest on this in the next few weeks. I am leaving the country at the end of this week and will be away until mid June. Since there seems to be a deadline approaching for the new Toolbox release, I feel that we should make a firm decision about whether to include any parts of this patch in the next release (1.5.8), or just put it on hold until I can work on it in more depth and put it into the subsequent release (1.5.9). 

One option is to include this foundational logic for the feature, but not expose it at all in the UI yet e.g. hide or disable the "Animate" button in the trace explorer interface for now. Then, when I come back to continue work on it, at least a large chunk of the foundation will be in place, and I can work on polishing up the user interface and experience and adding proper test coverage. What are your thoughts on this? I have no problem pushing this off until the next Toolbox release, if you are OK with it.

---

In response to your observations:

> Is Animation.tla supposed to be shipped as part of the Toolbox?

Eventually, yes, since I think that would make for the nicest user experience. I did not, however, consider that a necessary requirement for the MVP version of the feature. I imagine it could ultimately become a part of the TLA+ "standard library", analogous to the `TLC` module.

> If yes, Animation should adopt operator naming pattern (prefix of _TAGroup, _TARect, ... similar to _TEPosition and _TETrace to avoid name clashes

Acknowledged. That should be easy to address.

> Do you want to let users eventually specify the ```View````as a trace explorer expression?

This was not an explicit goal. I'm not really sure it would be a very ergonomic feature, since I imagine the "View" expressions for most non-trivial specs will be somewhat lengthy, and will require the definition of helper operators in a full TLA+ spec. If people are going to define helper operators for a "View" expression in a spec, they would need to extend the Animation module anyway, so I don't really see a particular need to make this a primary goal; or at least not in the initial version of the feature.

> The "Animate" button does not follow the new enabledment paradigm of the trace explorer (@quaeler)

I will look into that. If we decide to commit this code but disable the feature temporarily, then this can likely be addressed in the follow up work.

> I remember seeing controls to step through an animation in your 2018 presentation. Do you intend to add them later?

Yes, along with potentially other UI enhancements mentioned above. Again, I simply did not consider "clickable" controls a requirement for the MVP. Currently you can only use the left/right arrow keys to step through the trace.

lemmy:
@will62794 Where do I find Animation.tla?

will62794:
For testing, you can use this version: [Animation.tla](https://gist.github.com/will62794/69efad335d2c752649ab34f9b41ebffe)

lemmy:
The massive slowdown stems from the ```Join``` operator.  I replaced it (and SetToSeq) locally with a Java module overwrite:

```java
import tlc2.value.impl.FcnRcdValue;
import tlc2.value.impl.SetEnumValue;
import tlc2.value.impl.StringValue;
import tlc2.value.impl.TupleValue;
import tlc2.value.impl.Value;

public class Animation {

//	    IF Len(seq) = 0 THEN ""
//	    	    ELSE IF Len(seq) = 1 THEN Head(seq) 
//	    	    ELSE (Head(seq) \o delim \o Join(Tail(seq), delim))
	public static Value Join(Value seq, Value del) {
		if (seq == null || seq.size() == 0) {
			return new StringValue("");
		} else if (seq instanceof FcnRcdValue) {
			FcnRcdValue frv = (FcnRcdValue) seq;
			Value tuple = frv.toTuple();
			return Join(tuple, del);
		} else if (seq instanceof TupleValue) {
			String delim = ((StringValue) del).val.toString(); 
			final TupleValue tv = (TupleValue) seq;
			StringBuffer sb = new StringBuffer(tv.size());
			for (int i = 0; i < tv.elems.length; i++) {
				Value value = tv.elems[i];
				if (value instanceof StringValue) {
					StringValue sv = (StringValue) value;
					sb.append(sv.val);
				} else {
					sb.append(value.toString());
				}
				if (i + 1 < tv.elems.length) {
					sb.append(delim);
				}
			}
			return new StringValue(sb.toString());
		}
		return null;
	}

//	RECURSIVE SetToSeq(_)
//	SetToSeq(S) == IF S = {} THEN <<>> 
//	               ELSE LET v == Pick(S) IN <<v>> \o SetToSeq(S \ {v})
	public static Value SetToSeq(Value seq) {
		if (seq == null || seq.size() == 0) {
			return TupleValue.EmptyTuple;
		} else if (seq instanceof SetEnumValue) {
			final SetEnumValue sev = (SetEnumValue) seq;
			return new TupleValue(sev.elems.toArray());
		}
		return seq.toTuple();
	}
}
```

You cannot copy it verbatim because the string quoting in Join appears broken.  I will leave it at that because I don't know what quoting is expected.

lemmy:
> This was not an explicit goal. I'm not really sure it would be a very ergonomic feature, since I imagine the "View" expressions for most non-trivial specs will be somewhat lengthy, and will require the definition of helper operators in a full TLA+ spec. If people are going to define helper operators for a "View" expression in a spec, they would need to extend the Animation module anyway, so I don't really see a particular need to make this a primary goal; or at least not in the initial version of the feature.

#265 introduced support for named expressions with which helper operators are possible in the trace explorer.

will62794:
@lemmy I pushed a new patch that addresses some of the issues mentioned above. To avoid the hassle of dealing with strings in TLC, I decided to parse the view expression from within the TraceAnimationDataProvider and convert it to SVG in Java. This speeds things up, similar to your suggestion of overriding SVGElemToString directly. I also added next/previous button controls and a frame counter to the trace animation viewer. 

I think it's reasonable to not ship the `Animation.tla` module in the toolbox directly for the MVP version of the feature. [Here](https://gist.github.com/will62794/699490bea15ab3cd6aafc16c09259a15) is the updated version of the module that can be downloaded and used for creating animations in the toolbox. Perhaps you have thoughts on what would be the best way to distribute that to users i.e. place it in some central download location.

As far as testing goes, maybe you could provide some suggestions on the best kinds of tests to write for these changes. I tried to play around with the RCPTT test cases but encountered an error when launching the Eclipse application under test on my Mac:
```
An internal error occurred during: "Launching org.lamport.tla.toolbox.product.product.product (rcptt)".

org.eclipse.pde.internal.core.PDEState.<init>([Ljava/net/URL;ZZLorg/eclipse/core/runtime/IProgressMonitor;)V
```
I don't know what that's about. I also looked at some of the tests under `org.lamport.tla.toolbox.tool.tlc.ui.test` which seemed to test some of the UI elements, but I was not entirely sure if those would be the right template to follow.

lemmy:
@will62794 Why do you think it is better to handle the string munging in the Java layer?  What is the reasoning to not include ```Animation.tla```?

The RCPTT related exception is usually because of a mismatch between RCPTT and Eclipse.  Is Eclipse 2018-12 installed?

lemmy:
@will62794 Are you going to stay in St. Louis after TLA+ conf for Strange Loop?  If yes, I'd like to discuss/work on this PR in person.

will62794:
@lemmy I am attending the full Strangeloop conference, so yes, I will be around. I also have some free time to get back to this patch in the next couple of weeks, if you want to discuss the next steps sooner.

lemmy:
The new [CommunityModules repository](https://github.com/tlaplus/CommunityModules) would be one way to distributed ```Animation.tla```.

lemmy:
@will62794  Below are a few items I'd like to talk about tomorrow. If they don't make sense to you today, I'm happy to elaborate during the call. In other words, just ignore these ramblings for now.

Pure Spec
--------------
- Spec has to extend Animation.tla (related to #342)
- ```View``` definitions taint spec (better Trace Expressions?)

Processing
-----

- Hard-coded ```View``` term ```TraceAnimatorDataProvider.constructTraceAnimation``` will likely lead to name clashes 
  - ```TraceAnimatorDataProvider.svgElemToString(TLCVariableValue)``` will throw assertion error if ```View``` expression not a record

=> Splitting the evaluation into a TLA+ part (```View```) and a Toolbox-internal one (```TraceAnimatorDataProvider.constructTraceAnimation```) will probably lead to a poor user experience.  E.g. it will be impossible to diagnose why a ```View``` definition is incorrect.

=> Why hard-code it to render with SVG at all? Lets be open to extensions.

UI
--
- Do not process trace in UI/main thread in ```TraceAnimatorDataProvider.postProcess```
- ```Animate``` button misplaced
- No Play button Trace Animation view
- Trace Animation does not scale visualization

lemmy:
A few more thoughts:

1. The Trace Explorer could provide generic functionality (a button on the view) to copy the (filtered) trace to the clipboard.  For trace animation a user then pastes the SVG frames into the javascript tool
2. A more matching name for ```Animator.tla``` is ```Shapes.tla```.  The SVG specific operators in Shapes.tla should be moved to SVG.tla.  Both modules can be distributed via [CommunityModules]()

will62794:
To clarify, do you think it is still a good idea to embed the trace animation viewer inside the Toolbox itself, even if it is mainly a wrapper around an HTML/Javascript webpage? One alternative would be to have people open a separate webpage in their browser and paste in the appropriate data for the animation frames. The latter solution obviously doesn't really allow for automatic syncing between states in the error trace explorer and the animation. My impression was that we would embed the HTML/JS viewer inside the Toolbox itself(to allow for error trace/animation syncing), but just wanted to make sure you're on board with that.

Btw, I just tried out the CommunityModules and it's very easy to use. Implementing my own Java override, re-compiling the archive file, and loading it into the Toolbox was no trouble at all. Really useful feature.

lemmy:
I think it makes sense to initially implement the HTML/Javascript webpage separately and - upon success - integrate it into the Toolbox at a later point.  The API of the HTML/Javascript webpage is simple enough that it shouldn't be too difficult to come up with a Toolbox <> webpage API even late in the game.

lemmy:
> Btw, I just tried out the CommunityModules and it's very easy to use. Implementing my own Java override, re-compiling the archive file, and loading it into the Toolbox was no trouble at all. Really useful feature.

Please consider contributing your overrides to CommunityModules too. Once a PR is merged, a new build will be created automatically.

lemmy:
Copy&Paste to Clipboard added with https://github.com/tlaplus/tlaplus/commit/c7242cdc850baeee3ef8af1fe50ad0921d4fd49b

will62794:
Hi @lemmy thanks for the updates on your work. 

I am close to having the whole feature up and running. Based on your recent work, I don't think I will need to make any code changes to the Toolbox. Instead, I will open a PR against the CommunityModules repo where we can put the Animation related modules (SVG, etc.). 

After playing with the latest nightly Toolbox build a bit, I discovered a slightly different workflow for trace animation that I personally prefer. It's basically the same as what we discussed except that you can edit your animation in a full TLA+ file, instead of doing it in the trace expression editor. I recorded a "quick n' dirty" [screencast](https://www.dropbox.com/s/se49rqz9h4x5xvs/tla_animator_copy_paste_web_tool.mov?dl=0) to demo this workflow. I think that using the trace expression editor to write the full animation is still an option if people prefer that, but I thought about using this alternate approach once I realized you could import arbitrary modules into the trace explorer. For now I am planning to host the web tool on my own website so it's easily accessible. People can just go to the link and paste the text they copied from the Trace Explorer, similar to how you would use ShiViz.

I should be able to put up a PR for the community module changes in the next couple of days. If I have enough time, I am also planning to record some demos of the animation feature with various algorithms/specs that would be interesting to visualize.

lemmy:
> After playing with the latest nightly Toolbox build a bit, I discovered a slightly different workflow for trace animation that I personally prefer. It's basically the same as what we discussed except that you can edit your animation in a full TLA+ file instead of doing it in the trace expression editor. I recorded a "quick n' dirty" [screencast](https://www.dropbox.com/s/se49rqz9h4x5xvs/tla_animator_copy_paste_web_tool.mov?dl=0) to demo this workflow. I think that using the trace expression editor to write the full animation is still an option if people prefer that, but I thought about using this alternate approach once I realized you could import arbitrary modules into the trace explorer. For now I am planning to host the web tool on my own website so it's easily accessible. People can just go to the link and paste the text they copied from the Trace Explorer, similar to how you would use ShiViz.

Looks good, I actually had ShiViz and the Animator in mind when I implemented #342 whose screencast shows the workflow for ShiViz.  The ShiViz related operators have since been contributed to the CommunityModules though.

lemmy:
Marking closed as per previous comment.

Issue #391, opened 2019-11-13
TLC does not correctly compare integers or integer intervals due to overflow


Calvin-L:
The assertion in this spec should trivially pass:
```
---- MODULE Overflow ----

EXTENDS Integers

x == {1157660672, -989822976}
y == {-989822976, 1157660672}

ASSUME x = y

=========================
```

However, TLC reports that it fails:
```
$ tlc2 Overflow
[...]
Error: Assumption line 18, col 8 to line 18, col 12 of module Overflow is false.
Finished in 00s at (2019-11-12 15:58:54)
```

The root cause is that TLC uses simple subtraction to compare integers, which can overflow when those integers are very far apart. As a result, the two binary sets in the specification do not normalize to an identical internal representation.

I am aware of at least two places where TLC incorrectly uses subtraction to compare integer values:

 - https://github.com/tlaplus/tlaplus/blob/56740fd57b089eb8f85ef722dd0ee60a961b391f/tlatools/src/tlc2/value/impl/IntValue.java#L73
 - https://github.com/tlaplus/tlaplus/blob/56740fd57b089eb8f85ef722dd0ee60a961b391f/tlatools/src/tlc2/value/impl/IntervalValue.java#L40

There may be other locations as well.

A simple fix would be to use the library function [Integer.compare(int, int)](https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#compare(int,%20int)), instead of subtraction. The library function correctly handles this case.

lemmy:
Has this bug shown up in a real-world spec?

Calvin-L:
Not to my knowledge. I noticed the problem while reading through the TLC source code, and constructed the artificial spec above to exhibit it.

Calvin-L:
@lemmy Sorry to bug you again, but the fix is not quite right.

1. The computation `size = this.high - this.low + 1` returns zero when `high` = `Integer.MAX_VALUE` and `low` = `Integer.MIN_VALUE`. Therefore the subsequent test `size < 0` does not actually detect overflow.
2. The new `compareTo` method for intervals might not return zero ("equal") when the two intervals have different `low` values but are both empty because `high < low`.

Here's an spec that demonstrates the two problems:

```
---- MODULE Overflow2 ----

EXTENDS Integers, FiniteSets

MaxInt == 2147483647
MinInt == -MaxInt - 1

ASSUME Cardinality(MinInt .. MaxInt) > 0      \* FAILS due to (1)
[...]

PR #416, opened 2020-01-17
Fix some issues in IntervalValue

Calvin-L:
This fixes the new cases I found today in #391.

### DIFF
--- a/tlatools/src/tlc2/value/impl/IntervalValue.java
+++ b/tlatools/src/tlc2/value/impl/IntervalValue.java
@@ -38,6 +38,10 @@ implements Enumerable, Reducible {
 		        if (cmp != 0) {
 					return cmp;
 				}
+				if (this.size() == 0) {
+					// empty intervals are equal, regardless of the low value
+					return 0;
+				}
                 return Integer.compare(this.low, intv.low);
 			}
       // Well, we have to convert them to sets and compare.
--- a/tlatools/src/tlc2/value/impl/IntervalValue.java
+++ b/tlatools/src/tlc2/value/impl/IntervalValue.java
@@ -116,12 +116,13 @@ implements Enumerable, Reducible {
 		if (this.high < this.low) {
 			return 0;
 		}
-		final int size = this.high - this.low + 1;
-		if (size < 0) {
-			Assert.fail("Size of interval value exceeds the maximum representable size (32bits)"
-					+ Values.ppr(this.toString()) + ".");
+		try {
+			return Math.addExact(Math.subtractExact(this.high, this.low), 1);
+		} catch (ArithmeticException e) {
+			Assert.fail("Size of interval value exceeds the maximum representable size (32bits): "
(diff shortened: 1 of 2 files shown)


tlatools/src/tlc2/value/impl/IntervalValue.java
```diff
@@ -116,12 +116,13 @@ public final int size() {
 		if (this.high < this.low) {
 			return 0;
 		}
-		final int size = this.high - this.low + 1;
-		if (size < 0) {
-			Assert.fail("Size of interval value exceeds the maximum representable size (32bits)"
-					+ Values.ppr(this.toString()) + ".");
+		try {
+			return Math.addExact(Math.subtractExact(this.high, this.low), 1);
```
lemmy: @Calvin-L Do you have an estimate for how much slower this is going to be?

(PageQueue)

Calvin-L: Mmm, it's always hard to say anything definitive about small changes like this one. I ran the JMH benchmarks on my laptop for commits 293b755 (this PR) and a6a4c7f (current master). The results are here:

https://gist.github.com/Calvin-L/2d4163c1128a75b7d292158800b01f68

I can't claim to understand how to interpret the results. The average of the "score" fields differs by only about 5% between the two. My intuition says that's just noise, not anything meaningful.

I put together a [microbenchmark](https://gist.github.com/Calvin-L/985f8c442a77b95e076d012140628458) for just the `IntervalValue.size()` method, which suggests that my proposed implementation is about 3-4x slower than the current one. I highly doubt the `size()` method will ever be a bottleneck, but if 3-4x seems steep then we can consider devising some clever-er code to check for overflow.

lemmy: https://jmh.morethan.io/?source=https://gist.githubusercontent.com/Calvin-L/2d4163c1128a75b7d292158800b01f68/raw/617a9f988b1094a58f76312888e40b1a8ee20b94/DiskQueueBenachmark-1579313387-293b755.json is a visualization that is easier to interpret.  I would however suggest to write a dedicated JMH benchmark for IntervalValue#size (i.e. extend tlc2.value.impl.IntervalValueBenchmark), because JMH makes sure to minimize noise created from JVM optimizations.

lemmy: No need to bother (overhead is negligible): https://jmh.morethan.io/?source=https://gist.githubusercontent.com/lemmy/46a8c2ff3bb7c1718ebbcd6c2404230b/raw/0d1a2a92f31618021bb79283534b31380bd24b32/jmh.json

https://gist.github.com/lemmy/46a8c2ff3bb7c1718ebbcd6c2404230b

Will merge PR as is later.

lemmy:
Thanks, https://github.com/tlaplus/tlaplus/commit/4ca4c61aaabaf677b5084e02c9657dde3bdadc97

PR #425, opened 2020-02-04
Fix TLC rerun at same JVM for some temporal property

pfeodrippe:
I'm in a clojure REPL (but it's only a technicality). When running
the TLC model checker for some TLA file for the first time, it shows
me the correct error for a temporal property. But for a rerun of the
TLC model checker, it does not show me this error.

```clojure
;; Code which is run twice
(doto (tlc2.TLC.)
  (.handleParameters (into-array
                      ["/Users/feodrippe/dev/my-practical-tlaplus/ch1/wire.tla"]))
  (.process))
```

I was debugging it and found that there are some assumptions for the
initial values of the `LiveWorker` class. Maybe it should reset all the
static variables to their initial values when the method `process` is run,
but for now I've just created a public function to reset `errFoundByThread`
whenever someone needs it, I don't want to break anyone's code.

To give a more accurate time, you can also reset startTime at TLC class.

### DIFF
--- a/tlatools/src/tlc2/TLC.java
+++ b/tlatools/src/tlc2/TLC.java
@@ -990,17 +990,21 @@ public class TLC {
 		return dumpFile + suffix;
 	}
 
+    public void resetStartTime() {
+        startTime = System.currentTimeMillis();
+    }
+
 	/**
      * The processing method
      */
     public int process()
     {
         // UniqueString.initialize();
-        
-        // a JMX wrapper that exposes runtime statistics 
+
+        // a JMX wrapper that exposes runtime statistics
         TLCStandardMBean modelCheckerMXWrapper = TLCStandardMBean.getNullTLCStandardMBean();
-        
-		// SZ Feb 20, 2009: extracted this method to separate the 
+
+		// SZ Feb 20, 2009: extracted this method to separate the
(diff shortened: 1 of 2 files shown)


lemmy:
@pfeodrippe Thanks for the patch.  Unfortunately, it only fixes one of the many places where TLC does not re-initialize values. Years ago, a postdoc spend many weeks in an attempt to find and change all places, but ultimately gave up.  Since then, we accepted that TLC cannot check two different specs in the same JVM instance.  However, there is a [hack to reload all of TLC's classes](https://github.com/tlaplus/tlaplus/blob/master/tlatools/test/util/IsolatedTestCaseRunner.java) before model checking runs.  The same approach worked in prototype of mine where I wrapped tlc2.tool.TLC.

pfeodrippe:
I understand, @lemmy, I'm not so ambitious :)

But it could help other tools if they want to reset these values, it would not change one bit of what happens right now at TLC.

It's better than rely on setAcessible to set these fields, isn't? If it's mutable and static, it should have some way to set it.

Great that you have a isolated TLC runner, I will try it, but please consider these small "exposures" for these mutable fields o/

pfeodrippe:
Thank you for your explanation :)

PR #426, opened 2020-02-09
tlc2: disable deadlock check from configuration

### LINKED ISSUE #408: Define deadlock checking in model config file
TLC by default checks for deadlocks unless "-deadlock" is passed on the command-line.  In addition to the command-line, deadlock checking should be a configuration option of the config file.

(see https://github.com/alygin/vscode-tlaplus/issues/135 for inspiration)

BusyJay:
deadlock check will be disabled if either configuration file or command line disables it.

Close #408.


lemmy:
@BusyJay Thanks for the PR.  Can you outline the behavior if conflicting choices are made on the command-line and in the cfg file?

BusyJay:
To my understanding there is no such way to enable deadlock check from command line. So the behavior becomes:

Command line |  configuration |   Actual behavior
-------------| --- | --- 
enable implicitly | enable implicitly or explicitly | enable
enable implicitly | disable explicitly | disable
disable explicitly | enable implicitly or explicitly | disable
disable explicitly | disable explicitly | disable

In short, explicit command line choice is preferred, otherwise just follow the configuration.

lemmy:
It seems as if adding deadlock to the config *without* removing the command-line flag might not be such a good idea.

BusyJay:
I keep the command-line flag just for compatibility purpose. I can remove it if you like.

lemmy:
Let's not rush it, removing a flag is a big change.

tozaicevas:
Is this gonna get merged? I am coming from vscode extension, by the way.

lemmy:
@tozaicevas What makes more sense to be pushed to the VSCode extension?

tozaicevas:
Leaving `-deadlock` in TLC as it is, but updating vscode-tlaplus extension to include a command for passing `-deadlock` for opened spec or adding a checkbox in model checking panel, rather than having a global extension setting.
That's because, as I can see, this feature PR started from https://github.com/alygin/vscode-tlaplus/issues/135

andylokandy:
@tozaicevas Leaving the deadlock setting to the extension means that the author cannot assign the deadlock setting for each individual model. I think this PR is the right way.

tozaicevas:
> @tozaicevas Leaving the deadlock setting to the extension means that the author cannot assign the deadlock setting for each individual model. I think this PR is on the right way.

Well, in my case you'd have the advantage of tla toolbox and vscode extension being more similar.
vscode_tla

lemmy:
Who has the capacity to write [tests](https://github.com/tlaplus/tlaplus/tree/master/tlatools/org.lamport.tlatools/test/tlc2/tool) and [documentation](https://github.com/tlaplus/tlaplus/blob/a9cc2765396b27713af39d2b1b8a706f8283c584/tlatools/org.lamport.tlatools/src/tlc2/TLC.java#L1356-L1487) for this PR?  We are currently working towards the 1.6.1 release and if tests and doc can be pulled off soon, this PR could be part of 1.6.1.

BusyJay:
If you are not in a hurry, I can do that in the weekend.

BusyJay:
@lemmy Help messages and test case are added.

The ident configuration of the project is very weird. Even in the same file, like TLC.java, some lines use pure tabs, some lines use four spaces first and then use tabs for the left blanks. I'm not sure which is the preferred style, so I just keep using tab for the diff.

lemmy:
https://github.com/tlaplus/tlaplus/commit/6b4ed51d71d8920e7d46cc6082d74a49ddd9c4a2

vimalk78:
which version was this merged in? from vscode marketplace i only get version `v1.5.4`, and this option is not there

lemmy:
> which version was this merged in? from vscode marketplace i only get version `v1.5.4`, and this option is not there

https://marketplace.visualstudio.com/items?itemName=alygin.vscode-tlaplus-nightly

vimalk78:
I tried nightly  `v2022.11.806` already, but i dont see any checkbox for enabling/disabling deadlock

lemmy:
Add `CHECK_DEADLOCK FALSE` to TLC's config file or pass `-deadlock` in the VSCode command emmet.  The [checkbox proposal](https://github.com/tlaplus/tlaplus/pull/426#issuecomment-604635448) has not been implemented in the VSCode extension.

PR #440, opened 2020-04-10
Allow "-workers auto" to automatically detect available parallelism

Calvin-L:
I often forget how many cores my laptop has. (It's more than one but less than eight, I think.)

In any case, now I don't need to know. The toolbox GUI has had this feature forever, but I have wanted it in the command-line tools for a while.

As a bonus, this change also fixes a typo in another error message.


lemmy:
Thanks!
6c28e8538e0752f6fa26632add5e126e644d4068
4111431a169b96d0b2eb850fd6fb4416cf5af4ff

PR #465, opened 2020-05-18
[feature] add TLC constructor for custom state writer

pfeodrippe:
## Changes

- Add a custom TLC constructor to receive a `IStateWriter` and use it.
- Add `IStateWrite.writeState` method with 8 args (receiving a `Action` at the very end), it is used at `DotStateWriter.java` as if it was already at the interface.
- Fix some typos (well, I've assumed they were typos).

I could add tests for it, maybe at `TLCTest.java`? 

## Reason
There may be applications (or tooling) using TLA+ which could benefit of a fined control of the State Writer. One use case is to print the states in real time for custom debugging (could be interactive or not, it's up to the application). Another use case is to check the trace in a override operator without having to read from a file (I had no success using `TLCTrace` while in runtime, maybe someone could point me out other ways to do it?).

I have the second case. I'm testing some real system code from a simple TLA+ Spec (there were some other changes I've needed to make to at tlaplus code to know the current state when calling some operator, but it's very rough and need more think time).

If this is not desired, please tell me.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/TLC.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/TLC.java
@@ -162,10 +162,15 @@ public class TLC {
         traceDepth = 100;
 
         fpSetConfiguration = new FPSetConfiguration();
-        
+
         avoidMonolithSpecTECreation = false;
         waitingOnGenerationCompletion = new AtomicBoolean(false);
 	}
+
+    public TLC(IStateWriter stateWriter) {
+        this();
+        this.stateWriter = stateWriter;
+    }
 
     /*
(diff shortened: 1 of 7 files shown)


lemmy:
I believe that all three issues have been addressed by my recent commits.  With regards to debugging, the [DotStateWriter can already snapshot the current/partial state graph](https://www.reddit.com/r/tlaplus/comments/e5fkyr/debug_a_tlapluscal_spec_by_interactively/).

It is not clear to me what your two use cases are about.  Please open issues with *more/detailed* information.

pfeodrippe:
Great! Thank you o/

Issue #124, opened 2017-12-02
TLA+ Toolbox should have a REPL


hwayne:
From my own experiences with Alloy, a REPL is really nice. It makes learning the language easier, as you don't have to add anything to your core spec to explore. It makes it possible to query information about a given trace to learn more about it. Andrew Helwer [cites a REPL as a key advantage of Runway](https://groups.google.com/forum/#!topic/tlaplus/5Xd8kv288jE). 

This has been discussed [before](https://groups.google.com/forum/#!topic/tlaplus/WdDT5sKY7rI). Here were some alternatives raised:

* _Using "Evaluate Constant Expression":_ This only works on the initial state of a behavior. It takes time to run. It does not (easily) allow defining new operators without using `let`, which is awkward. No history of expressions.
* _Using multiple `ASSUME`:_ Limitations covered by Chris Newcombe [here](https://groups.google.com/d/msg/tlaplus/WdDT5sKY7rI/OcO7FUEtW-kJ).
* _Using Error Traces:_ Evaluated at every step of the trace, cluttery, specifically only for debugging while specs fail.

Lamport asked for [specific scenarios](https://groups.google.com/d/msg/tlaplus/WdDT5sKY7rI/v1feH95tiikJ) for where a REPL would be useful. Here are some I've run into:
* I'm not sure a particular part of the language works. Can I do `CHOOSE x \in A, b \in Y`? With a REPL, I can test that quickly and see it does not work.
* I'm not sure what a particular logical expression does. Is `\E foo \in F: \A bar \in B` what I want? With a REPL, I can try a few different logical expressions and see what I want.
* A spec is failing and I don't understand why. I could use Error Traces, but I don't know what I'm looking for yet. I want be able to take the state one step before the error and try several different expressions to see if I can figure out what it is I don't know.
* I want to use some debugging operators that shouldn't be part of the spec.

I think all of these could be resolved with a REPL that would allow the following:
1. Load at the initial state of a spec, or a given state of a failing behavior
1. Definition of new operators and actions
1. Evaluation of arbitrary expressions, including with the new operators and actions
1. Fast (May be untenable if requires heavy TLC refactors)

To make this simpler, the REPL should **not** be able to change the trace. If `x = 1` and `x' = 2`, `x' = 3` would be `FALSE`. This means that you cannot use the REPL to test a change to a spec. You can only use it to explore information about an existing spec.

lemmy:
[...]

PR #467, opened 2020-05-25
Add an interactive TLA+ REPL

will62794:
Initial implementation for #124.

As you suggested, I implemented the REPL in its own class, which you can specify at the command line. Here's a [screencast](https://asciinema.org/a/J3hcEGQBFaaZO1UbZyTG86wRv) demoing the usage. Right now you can't load an existing spec into the REPL context, but that could be an additional feature we implement once we have this basic REPL pushed. I already added some of the functionality to [load a spec](https://github.com/tlaplus/tlaplus/commit/a78b5111e528e3a8bee58834d0f4462d285d26c4#diff-21f38576ee05d6a9897410ac904f3373R59-R62) into the REPL in this patch, but it's not enabled via any command line options yet. Also, we could probably add a way to add new definitions in the REPL context in addition to just evaluating expressions.

The latency of evaluation is pretty good. Here's a set of latency measurements (i.e. "Duration") for a series of evaluations:
```
Welcome to the TLA+ REPL!
Enter any TLA+ expression.
(tla+) {1,2,3}
Duration: 612ms
{1, 2, 3}
(tla+) {1,2,3} \X {4,5,6}
Duration: 137ms
{<<1, 4>>, <<1, 5>>, <<1, 6>>, <<2, 4>>, <<2, 5>>, <<2, 6>>, <<3, 4>>, <<3, 5>>, <<3, 6>>}
(tla+) {1,2,3} \X {4,5,6}
Duration: 155ms
{<<1, 4>>, <<1, 5>>, <<1, 6>>, <<2, 4>>, <<2, 5>>, <<2, 6>>, <<3, 4>>, <<3, 5>>, <<3, 6>>}
(tla+) {1,2,3} \X {4,5,6}
Duration: 113ms
{<<1, 4>>, <<1, 5>>, <<1, 6>>, <<2, 4>>, <<2, 5>>, <<2, 6>>, <<3, 4>>, <<3, 5>>, <<3, 6>>}
(tla+) {1,2,3} \X {4,5,6}
Duration: 127ms
{<<1, 4>>, <<1, 5>>, <<1, 6>>, <<2, 4>>, <<2, 5>>, <<2, 6>>, <<3, 4>>, <<3, 5>>, <<3, 6>>}
(tla+) {1,2,3} \X {4,5,6,8}
Duration: 75ms
{<<1, 4>>, <<1, 5>>, <<1, 6>>, <<1, 8>>, <<2, 4>>, <<2, 5>>, <<2, 6>>, <<2, 8>>, <<3, 4>>, <<3, 5>>, <<3, 6>>, <<3, 8>>}
(tla+) {1,2,3} \X {4,5,6,8}
Duration: 77ms
{<<1, 4>>, <<1, 5>>, <<1, 6>>, <<1, 8>>, <<2, 4>>, <<2, 5>>, <<2, 6>>, <<2, 8>>, <<3, 4>>, <<3, 5>>, <<3, 6>>, <<3, 8>>}
(tla+) {1,2,3} \X {4,5,6,8}
Duration: 51ms
{<<1, 4>>, <<1, 5>>, <<1, 6>>, <<1, 8>>, <<2, 4>>, <<2, 5>>, <<2, 6>>, <<2, 8>>, <<3, 4>>, <<3, 5>>, <<3, 6>>, <<3, 8>>}
(tla+) <<1,2>> \o <<4,5>>
Duration: 66ms
<<1, 2, 4, 5>>
(tla+) {x *2 : x \in {1,2,3,4}}
Duration: 65ms
{2, 4, 6, 8}
(tla+) {1,2,3} \cap {3,4,5}
Duration: 66ms
{3}
```
Most are well under 100ms. 

I don't know if this implementation approach will satisfy all use cases, but I think it's probably a useful feature to have built into TLC.

Note: I added a [jline3](https://github.com/jline/jline3) dependency in the patch. Let me know if there's anything else I need to do to make sure that is set up correctly.


will62794:
Apparently I had the jline JAR present locally when I was testing, but never added it to version control. The latest commit fixes that.

tlatools/org.lamport.tlatools/src/tlc2/REPL.java
```diff
@@ -0,0 +1,166 @@
...
+
+            // Avoid sending log messages to stdout and reset the messages recording.
+            ToolIO.setMode(ToolIO.TOOL);
+            ToolIO.reset();
+
+            try {
+                // We placed the REPL spec files into a temporary directory, so, we add this temp directory
+                // path to the filename resolver used by the Tool.
+                SimpleFilenameToStream resolver = new SimpleFilenameToStream(replTempDir.toAbsolutePath().toString());
+                Tool tool = new FastTool(REPL_SPEC_NAME, REPL_SPEC_NAME, resolver);
```
lemmy, line 108: What makes `FastTool` faster is that it doesn't collect information for an error stack if an expression fails to evaluate.  I assume a REPL benefits from tlc2.tool.impl.CallStackTool that collects an error stack.

tlatools/org.lamport.tlatools/test/tlc2/REPLTest.java
```diff
@@ -0,0 +1,54 @@
...
+        // Set expressions.
+        res = repl.processInput("{1,2} \\X {3,4}");
+        assertEquals("{<<1, 3>>, <<1, 4>>, <<2, 3>>, <<2, 4>>}", res);
+        res = repl.processInput("{1,2} \\cup {3,4}");
+        assertEquals("{1, 2, 3, 4}", res);
+        res = repl.processInput("{1,2} \\cap {2,3}");
+        assertEquals("{2}", res);
+
+        // Tuple expressions.
+        res = repl.processInput("Append(<<1,2>>, 3)");
```
lemmy, line 36: Does the REPL by default extend all standard modules?  Is it possible to extend user modules?

tlatools/org.lamport.tlatools/src/tlc2/REPL.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    public static void main(String[] args) {
+        try {
+            Path tempDir = Files.createTempDirectory(TEMP_DIR_PREFIX);
+            REPL repl = new REPL(tempDir);
+            // TODO: Allow external spec file to be loaded into REPL context.
+            System.out.println("Welcome to the TLA+ REPL!");
+            System.out.println("Enter any TLA+ expression.");
```
lemmy, line 160: "Enter a constant-level TLA+ expression"?

tlatools/org.lamport.tlatools/src/tlc2/REPL.java
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Evaluate the given string input as a TLA+ expression.
+     *
+     * @return the pretty printed result of the evaluation or an empty string if there was an error.
+     */
+    public String processInput(String evalExpr) {
+
+        // The modules we will extend in the REPL environment.
+        String moduleExtends = "Naturals,Reals,Sequences,Bags,FiniteSets,TLC";
```
lemmy, line 58: * `Reals` extends `Integers` extends `Naturals`
* What about `Randomization`

tlatools/org.lamport.tlatools/customBuild.xml
```diff
@@ -249,7 +250,12 @@
 			</patternset>
 		</unzip>
 		<touch file="target/classes/META-INF/javamail.default.address.map"/>
-		
+
+		<unzip src="lib/jline-3.14.1.jar" dest="${class.dir}">
```
lemmy, line 254: BSD-3 (https://github.com/jline/jline3/blob/master/LICENSE.txt):

"Redistributions in binary form must reproduce the above copyright
notice, this list of conditions and the following disclaimer
in the documentation and/or other materials provided with
the distribution."

I assume `*.*` satisfies this requirement.

lemmy:
FWFI: jline library has to be added to MANIFEST.MF of the tla2tools OSGi bundle (compiler error in Eclipse without it).

tlatools/org.lamport.tlatools/src/tlc2/REPL.java
will62794: Ok, good to know. I briefly tried out `CallStackTool` in place of `FastTool` here and it didn't seem noticeably slower in terms of REPL latency. Using `CallStackTool` might be necessary to improve error reporting, but I was hoping to defer that work to follow up patches. Is that acceptable to you?

tlatools/org.lamport.tlatools/src/tlc2/REPL.java
will62794: Done.

lemmy:
@will62794 Do you want me to merge it now?

tlatools/org.lamport.tlatools/src/tlc2/REPL.java
```diff
@@ -0,0 +1,166 @@
...
+                return;
+            }
+        }
+    }
+
+    public static void main(String[] args) {
+        try {
+            Path tempDir = Files.createTempDirectory(TEMP_DIR_PREFIX);
+            REPL repl = new REPL(tempDir);
+            // TODO: Allow external spec file to be loaded into REPL context.
```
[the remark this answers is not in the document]
lemmy: Is the REPL supposed to replace the Toolbox's Evaluate Constant Expression functionality for command-line users (this is when users would want to load a spec)?  If yes, how does a user define constants?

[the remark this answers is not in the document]
will62794: That's a good point. For now, I was viewing this initial REPL implementation as mostly a built in TLC replacement for the functionality of https://github.com/will62794/tlaplus_repl, since that seems have to been useful for others when first exploring the language. That only provides the ability to evaluate expressions without any loaded TLA+ module. To support loading modules, I guess we would need to require the user give a config, same as they would when checking their spec `tlc2.TLC`. I forgot that in my comment above.

Ideally this REPL implementation can serve as a foundation for both CLI use cases and integration into other tools, including the Toolbox and/or VS Code. My thought was that for users who prefer to play around in a REPL from the CLI, they can use this tool, but that we could also take advantage of the low latency evaluation capability for improving things like the "Evaluate Constant Expression" feature of the Toolbox, and the expression evaluation feature of the VS Code extension. That might require having some sort of simple interprocess communication mechanism (plaintext?), if we wanted to have the Toolbox, for example, keep a REPL process running in the background that it could use for quick evaluation of an expression in context of a spec/model.

What are your thoughts on this?

will62794:
> @will62794 Do you want me to merge it now?

I still want to check to make sure I handled the licensing correctly. I will do that and then post here. 

Also, before we merge, I wonder if it would be helpful to see if anyone else in the community had thoughts on the implementation to see whether it satisfies some of their use cases. Perhaps @hwayne might have opinions on the feature as currently implemented and its future direction, since he created the original Github issue.

lemmy:
@ahelwer might also have some input.

lemmy:
@will62794 Time to merge?

will62794:
I mentioned it to @hwayne so was waiting to see if he had any comments.

hwayne:
I'm sorry, I've been overwhelmed in personal matters the past few days. I
put it on my workload for tomorrow to check this in detail.

will62794:
@hwayne No worries! There's no rush.

lemmy:
@alygin Will's REPL might be useful in the scope of https://github.com/alygin/vscode-tlaplus/issues/156#issuecomment-643647073 and your VSCode extension in general.

lemmy:
@hwayne Have you had a chance to look at this?

will62794:
@lemmy I'm ready to merge this, but I need to push one last commit first. The JARs I added don't seem to include this [LICENSE](https://github.com/jline/jline3/blob/master/LICENSE.txt) so I need to include it explicitly when we create `tla2tools.jar`. What would be a good place to put it in the Git repo, perhaps just inside `tlatools/org.lamport.tlatools/lib`? Not sure if there's a precedent here.

lemmy:
For the bits related to mail/smtp, there is a dedicated folder beneath `tlatools/org.lamport.tlatools/lib`.  I suggest to create another folder for the repl what then also contains any license files.

will62794:
Ok, added the license. We should be good to go. Feel free to squash and merge if everything looks good to you.

lemmy:
97afa3c6952e343ee2409366a668ba12afceeef4

PR #483, opened 2020-07-04
return non-zero error codes from SANY on more errors

zmatti:
This is intended as useful when using SANY as part of shell-based
tooling. The command line flag is introduced to keep it backwards
compatible so that no one breaks.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
@@ -77,6 +77,10 @@ public class SANY {
     // true <=> statistics about builtin operator usage 
     //          should be reported
 
+  private static boolean strictErrorCodes   = false;
+    // true <=> error level should be reported as the tools'
+    //          return value
+
   /**
    * The SANY.frontEndMain method Processes an entire TLA+ spec
    * that starts in the file named in "filename", including all files
@@ -143,7 +147,11 @@ public class SANY {
       syserr.println(e.toString());
       throw new FrontEndException(e);
     }
-    return 0;
+    if (strictErrorCodes) {
+      return spec.errorLevel;
+    } else {
+      return 0;
+    }
   }
 
   /** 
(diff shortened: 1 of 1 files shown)


tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -77,6 +77,10 @@
     // true <=> statistics about builtin operator usage 
     //          should be reported
 
+  private static boolean strictErrorCodes   = false;
```
lemmy, line 80: For consistency with the rest: Please rename to `doStrictErrorCodes`

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -410,6 +418,8 @@ else if (args[i].equals("-D") || args[i].equals("-d"))
            doDebugging        = !doDebugging;
       else if (args[i].equals("-STAT") || args[i].equals("-stat")) 
            doStats      = !doStats;      
+      else if (args[i].equals("-error-codes"))
```
lemmy, line 421: I suggest to call `toLowerCase` on `args[i]` to ignore casing.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -77,6 +77,10 @@
     // true <=> statistics about builtin operator usage 
     //          should be reported
 
+  private static boolean strictErrorCodes   = false;
+    // true <=> error level should be reported as the tools'
```
lemmy, line 81: Please bump SANY's version number and moddate.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
zmatti: Done

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
zmatti: Done

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
zmatti: Done. I believe you meant the constants towards the top of the file. Let me know if I misunderstood.

zmatti:
Thanks for the review! Made changes as requested, please take a look

PR #493, opened 2020-07-28
Closes #490: look up for modules in the directories that are defined in the environment variable TLA_PATH

### LINKED ISSUE #490: An environment variable for the standard modules?
> Run TLC with -DTLA-Library=/path/to/lib/archive or add the library archive to the Toolbox (File > Preferences > TLA+ Preferences > TLA+ library path locations).

It would be great to have a standard environment variable, e.g., called `TLA_PATH`. Similar to `PATH`, `PYTHON_PATH`, `JAVA_LIBRARY_PATH`, etc., it would list directories, from which the TLA+ tools would pull the library modules.

konnov:
This PR closes issue #490. The paths defined in the environment variable `TLA_PATH` are joined with the paths defined in the Java system variable `TLA-Library`. The latter is having priority over the former.

As it is not clear, how to test this behavior with unit tests, I only did rudimentary testing:

```bash
$ pwd
/Users/igor/tmp/tlapath
$ export TLA_PATH=`pwd`/a:`pwd`/b
$ echo $TLA_PATH 
/Users/igor/tmp/tlapath/a:/Users/igor/tmp/tlapath/b
$ cat a/A.tla
--------- MODULE A -----------
EXTENDS Integers
DO_A(x) == x + x
=======================
$ cat b/B.tla                 
--------- MODULE B -----------
EXTENDS A
DO_B(x) == DO_A(x) + 1
=======================
$ cat c/C.tla                                           
---------- MODULE C ---------
EXTENDS B
DO_C(x) == DO_B(4)
=======================
$ cd c
$ java -cp $HOME/tmp/tla2tools.jar tla2sany.SANY C.tla

****** SANY2 Version 2.2 created 08 July 2020

Parsing file /Users/igor/tmp/tlapath/c/C.tla
Parsing file /Users/igor/tmp/tlapath/b/B.tla
Parsing file /Users/igor/tmp/tlapath/a/A.tla
Parsing file /private/var/folders/42/77j16l715bd5k68fffptzgvw0000gn/T/Integers.tla
Parsing file /private/var/folders/42/77j16l715bd5k68fffptzgvw0000gn/T/Naturals.tla
Semantic processing of module Naturals
Semantic processing of module Integers
Semantic processing of module A
Semantic processing of module B
Semantic processing of module C
```

### DIFF
--- a/tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
+++ b/tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
@@ -31,6 +31,16 @@ import java.net.URL;
 public class SimpleFilenameToStream implements FilenameToStream {
 
 	public static final String TLA_LIBRARY = "TLA-Library";
+	
+	/**
+	 * The environment variable that enumerates the directories, in which
+	 * the modules should be looked up (following the syntax of PATH).
+	 * The Java system variable TLA_LIBRARY is joined with TLA_PATH.
+	 * The paths defined in TLA_LIBRARY have priority over the paths defined in TLA_PATH.
+	 * 
+	 * July 2020 - Igor Konnov, issue #490.
+	 */
+	public static final String TLA_PATH = "TLA_PATH";
 
 	public static final String STANDARD_MODULES_FOLDER = "StandardModules";
 
@@ -57,8 +67,8 @@ public class SimpleFilenameToStream implements FilenameToStream {
 /**
  * August 2014 - TL
  * This constructor was on the interface but was not implemented.
- * Now one can pass additiona libraries which will be added to the
- * installtion path and instead the path supplied using the system property TLA_LIBRARY
+ * Now one can pass additional libraries which will be added to the
+ * installation path and instead the path supplied using the system property TLA_LIBRARY
  * (which is being used in the default constructor).
  */
   public SimpleFilenameToStream(String[] anLibraryPaths) {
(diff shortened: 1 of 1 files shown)


lemmy:
A few ad-hoc questions/comments:. 

What takes precedence (env var or -D    property)?

I believe that the changes have to be ported to RCPFilename... for the env var to be supported by the Toolbox

konnov:
`-DTLA-Library` has priority over `TLA_PATH`. I will have a look at `RCPFilename`.

konnov:
It looks like `RCPNameToFileIStream` does not respect `-DTLA-Library` either. Is that true?

konnov:
I just googled a bit. Perhaps, environment variables do not fit in the eclipse ideology. They seem to be using URIs everywhere.

lemmy:
The Toolbox has a preference to configure Toolbox-wide includes, but I personally would like to also have a system-wide include configured (via TLA_PATH) for e.g. the community modules.

lemmy:
Btw. the test suite is executed with `ant -f tlatools/org.lamport.tlatools/customBuild.xml test` assuming the classes have been compiled.

konnov:
Right. The problem is not in running the unit tests, but in writing a unit test that would test the new code using the environment variable. I even don't know how to do it JUnit.

lemmy:
1. Derive a new JUnit test from `ModelCheckerTestCase` (see other subclasses for examples)
2. New test case checks a spec in `test-module` that extends a module in `TLA_PATH`
3. Add a new [`batchtest` or `junit` block](https://ant.apache.org/manual/Tasks/junit.html) to `test` target in `customBuild.xml` to set `TLA_PATH`

lemmy:
@konnov I'll be back from vacation tomorrow.  How much of a blocker is this for you (I see that you have a workaround in apalache)?  If this gets merged, are you going to wait until the next release anyway or will you consume a nightly build?

konnov:
This is not a blocker for us at all. We will use the workaround until tla2tools gets released.

lemmy:
This feature will also need proper documentation.  I suggest https://github.com/tlaplus/tlaplus/blob/e434e139adebf73ed8f9470117031f1ad4b749df/tlatools/org.lamport.tlatools/src/tlc2/TLC.java#L1335-L1471 because standalone SANY doesn't have a help page.

konnov:
@lemmy I have found out that I've created this PR right off the master branch in my fork. Do you know how to retarget it against https://github.com/konnov/tlaplus/tree/igor/tlapath? I would like to advance master in my fork.

lemmy:
@konnov I would try to cherry-pick the commits related to this PR from your `master` into your `tlapath` branch, followed by a reset of `master`.

lemmy:
I'm a bit cautious because everything around reading resources from disk is extremely messy in TLC and in the Toolbox (see `RCPFilenameToStream` vs. `SimpleFilenameToStream.java`), and adding support for env vars would just add to the complexity.  That said, having the functionality would be convenient, especially if users could add `CommunityModules-deps.jar` to a system-wide `TLA_PATH` to globally include it.

lemmy:
Also, are we sure that TLAPS supports `TLA_PATH`?

Calvin-L:
If I had to guess, I would guess TLAPS does not honor `TLA_PATH`. I believe it has its own logic for locating modules on the filesystem.

lemmy:
> If I had to guess, I would guess TLAPS does not honor `TLA_PATH`. I believe it has its own logic for locating modules on the filesystem.

You can pass it multiple input paths with `tlapm -I /path/to/TLA/modules -I ...`, which is at least consistent with TLC in the sense that they both only support parameters and not env vars.

PR #509, opened 2020-09-23
Refactored parsing of command line arguments in TLC class

ahelwer:
- Separated parsing, validation, transformation, and consumption of command line options
- Wrote many regression tests before transition to give high assurance of backwards compatibility
- Introduced Either<L,R> and OneOf<T1,T2,T3> types to util module
- Changed TLC class fields to be initialized in place instead of in constructor


ahelwer:
@lemmy this is now good to go

lemmy:
What is the high-level description of this change that you would want to read in a git commit (assume commits squashed) one year from now?  Especially, why was it necessary?

ahelwer:
Benefits as I see them:
- Enabled unit testing of all command line options
- Added many regression tests to ensure CL semantics are preserved going forward
- Greatly simplified process of adding new CL parameters
- Made TLC class much smaller & easier to understand
- Good first step toward moving away from static global variables - can eventually use non-static CommandLineOptions object to store all settings

ahelwer:
Closing this PR because I was able to salvage many of the UTs in TLCTest.java without requiring large-scale modification of TLC.java; see https://github.com/tlaplus/tlaplus/pull/554

Issue #221, opened 2019-01-13
using unix line endings on windows gives baffling errors


dimbleby:
I'm using version 1.57 of the toolbox.

Although I'm working on Windows, I generally set up my git repositories to use unix line endings.  

Trying to translate the following from pluscal to tlaplus:

```pluscal
--------------------------- MODULE foo ---------------------------
(*--algorithm upgrade
begin
    skip;
end algorithm;*)
=============================================================================
```

gives "Unrecoverable error: Text on same line following `*)` that ends the comment containing the algorithm".  Which is puzzling, partly because it's rather awkwardly phrased so that I'm not completely sure what it's trying to tell me - but mostly because I think that whatever it is trying to tell me is not true.

As I say, I set up my git to commit with LF endings.  So I've also had fun where I write a spec that's valid, commit it, and then find that it doesn't work any more.  The error that I see then is usually "Unrecoverable error: No line containing `END TRANSLATION".  But I can see quite clearly that there **is** such a line.

This has all been very confusing to debug.  

It would be great if the toolbox was agnostic about line endings; if that's not possible, it would be helpful if it could give more useful errors in such cases.

lemmy:
Related to #157

lemmy:
@dimbleby Can you please check if the https://github.com/tlaplus/tlaplus/releases/tag/v1.6.0 release fixes this issue?

dimbleby:
It looks as though the toolbox now re-dos-es the file before performing translation - is that what you're expecting?

While that's not completely ideal - I'd prefer that the toolbox was truly agnostic about the line endings and would leave my file using whichever format it found - it's certainly a huge improvement on the cryptic errors that I was seeing when I reported this.

Cheers.

lemmy:
Please reopen if this remains an issue (with latest and greatest).

PR #551, opened 2020-12-12
Test tlc2.tool.Github407Test now passes on Windows

ahelwer:
Failure was due to line ending differences

### DIFF
--- a/tlatools/org.lamport.tlatools/test/tlc2/tool/Github407Test.java
+++ b/tlatools/org.lamport.tlatools/test/tlc2/tool/Github407Test.java
@@ -25,14 +25,21 @@
  ******************************************************************************/
 package tlc2.tool;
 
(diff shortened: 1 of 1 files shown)


lemmy:
It is probably good to make the test more robust, but the underlying issue is that git on windows converts the line endings, no?

ahelwer:
@lemmy I fixed this a few weeks back (decided to spin it out into its own PR) but I think the issue is that TLC writes files using windows line endings, but the file used for comparison uses linux line endings.

lemmy:
Perhaps, it's better to add a system property or even a command-line parameter to force a line ending. 

Related: https://github.com/tlaplus/tlaplus/issues/221

lemmy:
Merged with https://github.com/tlaplus/tlaplus/commit/c730bc152dd6439d514c17a045d5e5cdb0dba7ed

PR #554, opened 2020-12-17
TLC now generates TE specs by default with stub ALIAS

ahelwer:
- TE spec generated by default instead of on request
- TE spec not generated when TLC is given parameters `-noGenerateSpecTE`, `-continue`, or `-tool`
- TE spec has filename pattern `SpecTE_ModuleName_Timestamp`
- Can control TE spec output directory with `-teSpecOutDir`
- TE spec includes stub trace expression function applying identity function to existing variables, with directions for adding additional custom trace expressions


lemmy:
Merged with https://github.com/tlaplus/tlaplus/commit/7c9e44327242bc4f3eba08780c453fd1264f2b91

Issue #362, opened 2019-09-12
Name clash between variable in refined spec and operator in instantiated spec


Calvin-L:
If specification `A` contains a variable with the same name as a zero-argument operator in a specification it instantiates, TLC seems to get confused about whether to use the variable or the operator when evaluating operators defined in the instantiated specification. (Wow, what a mouthful!)

Here is an example:

A.tla
```
---- MODULE A ----

EXTENDS TLC

VARIABLES overloadedName

Abstract == INSTANCE B WITH
    x <- "x"

Init ==
    /\ overloadedName = "fizzbuzz"
    /\ Print(<<"Evaluated initial state in A; overloadedName is: ", overloadedName>>, TRUE)
    /\ Print(<<"From A's perspective, B's overloadedName is: ", Abstract!overloadedName>>, TRUE)

Next == UNCHANGED overloadedName

Correct == Abstract!Spec

==================
```

B.tla
```
---- MODULE B ----

EXTENDS TLC

VARIABLES x

overloadedName == x

Init ==
    /\ Print(<<"Evaluating initial state in B; overloadedName is ", overloadedName>>, TRUE)
    /\ overloadedName = "x"
    /\ x = "x"

Next == UNCHANGED x
Spec == Init /\ [][Next]_x

==================
```

A.cfg
```
INIT Init
NEXT Next
PROPERTY Correct
```

TLC reports:

```
$ tlc2 A
TLC2 Version 2.14 of 10 July 2019 (rev: ${git.shortRevision})
Running breadth-first search Model-Checking with fp 120 and seed -518597667049658357 with 1 worker on 4 cores with 3641MB heap and 64MB offheap memory [pid: 60619] (Mac OS X 10.14.6 x86_64, Azul Systems, Inc. 11.0.1 x86_64, MSBDiskFPSet, DiskStateQueue).
Parsing file /Users/cloncari/sw/src/plsql/tmp/A.tla
Parsing file /private/var/folders/k0/tw__ldzn7r19gby9d93tffh40000gn/T/TLC.tla
Parsing file /Users/cloncari/sw/src/plsql/tmp/B.tla
Parsing file /private/var/folders/k0/tw__ldzn7r19gby9d93tffh40000gn/T/Naturals.tla
Parsing file /private/var/folders/k0/tw__ldzn7r19gby9d93tffh40000gn/T/Sequences.tla
[...]

PR #569, opened 2021-01-23
Fix issue #362: only variables should be looked up as variables

Calvin-L:
By poking around in a debugger, I determined that this lookup method is the source of the problem. Most of the time it is true that a symbol with the same name as a variable is indeed itself a variable---but that is not always true, as the test case demonstrates.


lemmy:
@Calvin-L Do you see anything wrong with https://github.com/lemmy/tlaplus/pull/new/mku-issue362  ?  It extends the test case to include constants and only excludes `UserDefinedOpKind` from the lookup.

Calvin-L:
Seems great 👍

lemmy:
Superseded by https://github.com/tlaplus/tlaplus/pull/571

PR #571, opened 2021-01-25
Fix issue #362: only variables should be looked up as variables

lemmy:


lemmy:
Merged with https://github.com/tlaplus/tlaplus/commit/512992766bd5ac45d3fb6dd7d4b13fb8bca4bdaa

Issue #624, opened 2021-05-14
PlusCal translator ignores dangling '}' at the end of block


ztlpn:
Hi! I think the bug is best explained with a simple repro.

```
(* --algorithm test {

variables x = FALSE, y = FALSE;

process (a = "a") {
La: x := TRUE;
    }
}

process (b = "b") {
Lb: y := TRUE;
}

}
```

Note the dangling '}' in the definition of process `a`. Not only it is not reported, but it seems that all that follows it is ignored (including the definition of process `b` leading to its absence from the generated spec).

Toolbox version:
```
This is Version 1.7.1 of 31 December 2020 and includes:
  - SANY Version 2.2 of 20 April 2020
  - TLC Version 2.16 of 31 December 2020
  - PlusCal Version 1.11 of 31 December 2020
  - TLATeX Version 1.0 of 20 September 2017
```

lemmy:
@ztlpn The `}` is not dangling but closing the algorithm definition.  What's actually dangling is here the complete definition of `process b`.

ztlpn:
Ah, so the comment containing the PlusCal algorithm description can continue after the algorithm description! Thanks, it makes sense then, even though it leads to counterintuitive behavior. Looks like there is nothing a translator could do apart from perhaps issuing a warning.

Also looks like p-syntax is preferable as it is more robust in this regard.

lemmy:
Can you provide a patch that makes the translator issue a warning?

ztlpn:
So I've finally managed to spend some time on it. There was a small problem that PlusCal translator warnings were not displayed in the toolbox interface (fixed in #631).

Another problem is to specify the precise rule for the warning. It appears that there are almost no constraints on what a comment can contain after the PlusCal code so it will be difficult to devise a rule with a good balance of false positives and negatives. Some possibilities that I can think of:
* Warn if there are non-whitespace characters in the comment after PCal code. Looks too restrictive.
* Warn if the c-syntax is enabled and there is `}` in the comment after the algorithm. Better, but I can still imagine comments where `}` is legitimately used (for example for set literals).
* Warn if the c-syntax is enabled and there are mismatched `{` and `}`  in the comment after the algorithm. Looks complicated, but I can't think of any false positives.

Honestly I don't particularly like any one of them. @lemmy what do you think?

lemmy:
I also don't see a good solution to dangling blocks of PlusCal.

[...]

PR #631, opened 2021-05-27
Display PlusCal translator warnings in the toolbox interface

ztlpn:
This PR is in preparation for #624 

Currently PCal translator doesn't emit many warnings (namely two: "symbols were renamed" and a warning about conflicting fairness options if `--fair algorithm` has been specified) but I think it still makes sense to display them (and possible future warnings) in the toolbox interface.

This also fixes a small bug where PCal warning messages didn't get their own message in ToolIO and instead got coalesced the next message.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/pcal/PCalTLAGenerator.java
+++ b/tlatools/org.lamport.tlatools/src/pcal/PCalTLAGenerator.java
@@ -45,9 +45,6 @@ public class PCalTLAGenerator
         }
 
         st.Disambiguate();
-        if (st.disambiguateReport.size() > 0)
-            // SZ March 11, 2009. Warning reporting moved to PCalDebug 
-            PcalDebug.reportWarning("symbols were renamed.");
         if (st.errorReport.length() > 0)
             throw new RemoveNameConflictsException(st.errorReport);
         try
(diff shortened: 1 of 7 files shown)


ztlpn:
Re: https://github.com/tlaplus/tlaplus/issues/624#issuecomment-849737712

By "the new warning" do you mean the warning about the dangling blocks or the warning about conflicting fairness options? The latter one isn't new :) anyway it is triggered by
```
------------------------------ MODULE warning ------------------------------

(*
PlusCal options (-nof)

--fair algorithm test
variables x = FALSE;
begin
    x := TRUE;
end algorithm;
*)

```

If you wish, I can add it to test cases – at the very least it would test that warnings are properly registered in ToolIO.

Also there is a warning generated when identical labels in different processes are renamed. Which is a bit superfluous in my opinion - info about label renames also is written into the translation header. But I can leave it as-is.

As far as I can see, there are no other warnings in the existing code.

lemmy:
Yes, please add the test you're proposing.  I've always found the warning at the translation header sufficient. Perhaps, create a new issue once the current one is addressed?

tlatools/org.lamport.tlatools/src/pcal/PCalTLAGenerator.java
```diff
@@ -45,9 +45,6 @@ public void removeNameConflicts() throws RemoveNameConflictsException
         }
 
         st.Disambiguate();
-        if (st.disambiguateReport.size() > 0)
```
lemmy, line 48: I'm not sure we should roll this change into this PR for backward compatibility reasons.  Is there a reason for removal besides it being annoying?

ztlpn: No other reason. I'll revert it then.

PR #632, opened 2021-05-30
Augment simulate stats with basic and full options.

pfeodrippe:
Proposed by @lemmy.

Replaces `stats=true` of `-simulate` by `stats=full` and adds a new options, `stats=basic`.

`stats=full` does the same things as `stats=true`, just the filename that now ends with `_full_actions.dot`.

While `stats=full` uses the full context of the actions, it can lead to a very large and hard to understand [picture of the states transition](https://user-images.githubusercontent.com/2124834/120088894-a095d480-c0cb-11eb-9fdf-5e57593cad68.png). `stats=basic` is similar, but it removes the context of the actions and leaves only the nodes and their connections (state transitions) with other nodes, letting the user with a [higher level overview](https://user-images.githubusercontent.com/2124834/120088917-e488d980-c0cb-11eb-804f-75a938b29050.png) of what's happening.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/TLC.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/TLC.java
@@ -159,7 +159,7 @@ public class TLC {
      * Name of the file to which to write state traces.
      */
     private String traceFile = null;
-    private boolean traceActions = false;
+    private String traceActions = null;
     /**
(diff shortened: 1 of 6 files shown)


lemmy:
Thanks, merged with https://github.com/tlaplus/tlaplus/commit/c3812215d76743c82887e2f0d691460b743d5dba

PR #633, opened 2021-05-30
Make variables appear before definitions on pluscal translation

### LINKED ISSUE #617: Feature Request: define process-local variables above PlusCal define block
Currently PlusCal compilation produces something like this:

```tla
VARIABLES global_variables

\* Anything in a define block goes here

VARIABLES local_variables
```

This means that define block operators can't reference any process local variables. If you want to enforce a type invariant, you have to put it after the pluscal translation.

pfeodrippe:
Resolves #617.

Just concatenates all variables before any definition on pluscal
translation.

This may break some existing pluscal specs which define operators in
`define` with arguments having the same name as some of the local
variables.

Example
```tla
(*--algorithm wire
variables
 people = {"alice", "bob"},
 acc = [p \in people |-> 5],
define
 XX(amount) == amount
 NoOverdrafts == \A p \in people: acc[p] >= 0
end define;
process Wire \in 1..2
 ...
 amount \in 1..acc[sender];
 ...
end algorithm;*)
\* BEGIN TRANSLATION (chksum(pcal) = "4a6f14c6" /\ chksum(tla) = "5be7c6b2")
VARIABLES people, acc, pc, sender, receiver, amount

(* define statement *)
XX(amount) == amount \* <------ "Multiply-defined symbol 'amount'" error!
...
```

The changes in `PlusCal.tla` and `PlusCal2.tla` are not needed, but
I've thought that maybe we are trying to maintain the algorithm in
sync with each other?

I can add test for it if we want.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/pcal/PcalTLAGen.java
+++ b/tlatools/org.lamport.tlatools/src/pcal/PcalTLAGen.java
@@ -1920,24 +1920,19 @@ public class PcalTLAGen
         }
         ;
 
-        if (EmptyExpr(defs))
+        /******************************************************************
+        * Generate a single VARIABLES statement and set gVars to vector   *
+        * of all variables.                                               *        
+        ******************************************************************/
+        gVars.addAll(lVars);
+        gVarsSource.addAll(lVarsSource) ;
+        GenVarDecl(gVars, gVarsSource); 
+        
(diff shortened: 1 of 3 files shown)


pfeodrippe:
If we decide to keep with this feature, I will fix the tests o/

PR #634, opened 2021-05-31
Fix invalid disjunction for long procedure names.

### LINKED ISSUE #370: PlusCal translates to invalid TLA+ if procedure identifier is too long
With toolbox 1.6.0 this works:

```
procedure Foo() begin i:
return;
end procedure;

process Bar \in {1} begin j:
while TRUE do
  call Foo();
end while;
end process;
```

but this doesn't :
```
procedure ProcedureWithVeryLongNameIndeedAndSomeExtraExtraExtra() begin i:
return;
end procedure;

process Bar \in {1} begin j:
while TRUE do
  call ProcedureWithVeryLongNameIndeedAndSomeExtraExtraExtra();
end while;
end process;
```

First one get's transpiled to:
```
Next == (\E self \in ProcSet: Foo(self))
           \/ (\E self \in {1}: Bar(self))
```
And latter one is to:
```
Next == (\E self \in ProcSet:  \/ 
                               \/ ProcedureWithVeryLongNameIndeedAndSomeExtraExtraExtra(self))
           \/ (\E self \in {1}: Bar(self))
```
TLA+ cannot parse it and fails with :
image

pfeodrippe:
Fixes #370.

Check if string buffer is bigger than 0 so we don't have bogus
disjunctions with long names.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/pcal/PcalTLAGen.java
+++ b/tlatools/org.lamport.tlatools/src/pcal/PcalTLAGen.java
@@ -2802,7 +2802,7 @@ public class PcalTLAGen
             for (int i = 0; i < st.procs.size(); i++)
             {
                 PcalSymTab.ProcedureEntry p = (PcalSymTab.ProcedureEntry) st.procs.elementAt(i);
-                if ((p.name.length() + "(self) \\/ ".length() + sb.length()) > max)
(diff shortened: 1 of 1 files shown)


pfeodrippe:
Can add a test here if this fix appears correct.

PR #658, opened 2021-08-12
fix: append of int with list

lalo:


tlatools/org.lamport.tlatools/test/tlc2/REPLTest.java
```diff
@@ -35,8 +35,8 @@ public void testProcessInput() throws IOException {
         // Tuple expressions.
         res = repl.processInput("Append(<<1,2>>, 3)");
         assertEquals("<<1, 2, 3>>", res);
-        res = repl.processInput("Append(3, <<1,2>>)"); // error.
-        assertEquals("", res);
+        res = repl.processInput("Append(3, <<1,2>>)");
+        assertEquals("<<1, 2, 3>>", res);
```
lemmy, line 39: This is inconsistent with the definitions of `Append` and `\o` in Sequences.tla:

```tla
s \o t == [i \in 1..(Len(s) + Len(t)) |-> IF i \leq Len(s) THEN s[i]
                                                           ELSE t[i-Len(s)]]


Append(s, e) == s \o <<e>>
```

lalo: Ah originally I assumed a single element is equivalent to a set containing only itself. But even if this is the case, it would still be wrong since it would be << 3, <<1, 2>> >> if not we can throw something more specific. Is an integer consider an 's'?

lemmy: We don't know what the elements of `3` are, thus we don't know what e.g. `Len(3)` is equal to.

PR #693, opened 2021-11-12
Document TLC Java system properties useful for many-core settings

zlox:
As [discussed on the tlaplus Google Group](https://groups.google.com/g/tlaplus/c/pi5B8YaJzcI), there are a couple of TLC options that can be set through Java system properties that significantly improve TLC's peformance on many-core machines. But they are hard to discover, so this PR should help with that.


lemmy:
Thanks, merged with minor rewording: https://github.com/tlaplus/tlaplus/commit/a30019d42b83e97b36a29346b62c964f7eacf943

Issue #439, opened 2020-04-04
Thread safety concerns due to "memory barrier" pattern in value classes


Calvin-L:
Many of the `Value` classes contain instances of a pattern that I think is an incorrect attempt at creating a "memory barrier". This pattern is [known to be incorrect in Java](https://www.cs.umd.edu/~pugh/java/memoryModel/BidirectionalMemoryBarrier.html).

The essence of the incorrect memory barrier pattern is
 - Writers acquire a lock, initialize an object, release the lock, re-acquire the lock, then write the initialized object to a place where readers can see it. Finally, they release the lock again.
 - Readers do not use any synchronization.

This pattern is seductive because it _does_ constrain the order in which writes reach main memory. There is also no issue if all reads and writes happen in the same thread. However, it is incorrect when there are concurrent writers and readers because the readers do not have to observe those writes in order. Readers can observe writes out-of-order if they page-in the reference to the initialized object before they page-in the writes that initialize the fields of the object.

Here is an example from `FcnRcdValue`. The "writer" pathway is in `createIndex()`:

```java
int[] tbl = new int[len];
Arrays.fill(tbl, -1);

synchronized(this) {
  for (int i = 0; i < this.domain.length; i++) {
    int loc = (this.domain[i].hashCode() & 0x7FFFFFFF) % len;
    while (tbl[loc] != -1) {
      loc = (loc + 1) % len;
    }
    tbl[loc] = i;
  }
}

synchronized(this) { this.indexTbl = tbl; }
```

The first `synchronized` block initializes the cells of the local array `tbl`, and the second one writes `tbl` to `this.indexTbl` where it can be picked up by readers.

The reader pathway is in `select(arg)`:

```java
if (this.indexTbl != null) {
  return this.values[this.lookupIndex(arg)];
}
if (this.isNorm) this.createIndex();

if (this.indexTbl != null) {
  return this.values[this.lookupIndex(arg)];
}
```

In this example, the reader's first check (`this.indexTbl != null`) can be true even when the cells of the `indexTbl` array are uninitialized (i.e. all zero). Even if the cells are initialized in main memory, those writes may not be visible to the reader yet. This could lead to readers crashing or returning incorrect values for function applications.

These classes contain instances of this pattern:
 - `FcnRcdValue`
 - `SetCapValue`
 - `SetCupValue`
 - `SetDiffValue`
 - `SetOfFcnsValue`
 - `SetOfRcdsValue`
 - `SetOfTuplesValue`
[...]

PR #695, opened 2021-11-23
Remove bidirectional memory barriers from set value classes

Calvin-L:
Removing the synchronization from these classes is safe because:

 - Bidirectional memory barriers are not safe to begin with. (https://www.cs.umd.edu/~pugh/java/memoryModel/BidirectionalMemoryBarrier.html)
 - All these classes do their synchronization in `convertAndCache()`, which is called during fingerprinting, before the values are shared with other threads.  Subsequent calls to `convertAndCache()` will be no-ops.

For more information, see discussion in GitHub issue #439.

----

Note that this PR does not cover `FcnRcdValue`. There is still some open discussion in #439 about what to do with that one.

I think this PR is worth merging because it (1) shortens the code and (2) removes a misleading and ineffective synchronization pattern.

That said, there are also good arguments _not_ to merge this PR. First, removing synchronization may uncover latent thread-related defects in the TLC code. (I have verified that the tests still pass, but that's no guarantee.) Second, I do not expect there will be a measurable performance impact from this change. The code path appears to be seldom-used.


lemmy:
Thanks for the PR. Independently of the actual change, looking at the copyright headers, would you be able to [sign our CLA](https://github.com/tlaplus/tlaplus/blob/master/general/cla/YOUR-NAME.txt.template)?

Calvin-L:
I think so---but since I'm doing this on work time, I'll need some internal approvals first. I'll get back to you on this sometime after Thanksgiving. 🦃

Calvin-L:
@lemmy CLA signed!

lemmy:
Pushed [671ed8d](https://github.com/tlaplus/tlaplus/pull/695/commits/671ed8dc39ba65ee08a1c535bb46a7a69008aad8) to master to diagnose build failure.

lemmy:
@Calvin-L Can you rebase and push your remove-memory-barrier-from-set-classes branch to trigger the PullRequest builder?  a46599c0a6c8c6a376ab1018e2564e947636b520 should make sure that the PR builder goes through this time.

lemmy:
@Calvin-L Assuming you accepted the Github invite, you should be able squash and merge your commit.

Calvin-L:
Done! Thanks for your comments and help. :)

PR #696, opened 2021-11-25
Thread safety concern due to "memory barrier" in FcnRcdValue#select

lemmy:
Part of https://github.com/tlaplus/tlaplus/issues/439

@Calvin-L @craft095 Please review commits and reasoning in https://github.com/tlaplus/tlaplus/issues/439#issuecomment-977308905


tlatools/org.lamport.tlatools/src/tlc2/value/impl/FcnRcdValue.java
```diff
@@ -381,76 +379,80 @@ public final Value select(Value arg) {
...
+		while (low <= high) {
+			// cast to int is the same as flooring for positive ints.
+			final int mid = low + (high - low) / 2;
+			final int cmp = this.domain[mid].compareTo(arg);
+			if (cmp == 0) {
+				return this.values[mid];
+			} else if (cmp < 0) {
+				low = mid + 1;
+			} else {
+				high = mid;
```
craft095, line 448: It loops forever if all domain elements are greater than arg. E.g. looking for 'a' in ['b']

tlatools/org.lamport.tlatools/test/tlc2/value/impl/FcnRcdValueTest.java
```diff
@@ -0,0 +1,96 @@
...
+	public static void setup() {
+		// Make test repeatable by setting random seed always to same value. 
+		RandomEnumerableValues.setSeed(15041980L);
+		// Needed to insert elements into java.util.Set (because of hashcode) later to
+		// detect duplicates.
+		FP64.Init();
+	}
+
+	@Test
+	public void testSelect() {
```
craft095, line 70: Test cases for missing values would be helpful:
- empty domain
- all domain values are less than the value
- all domain values are greater than the value

tlatools/org.lamport.tlatools/src/tlc2/value/impl/FcnRcdValue.java
```diff
@@ -381,76 +379,80 @@ public final Value select(Value arg) {
...
+			}
+		}
+		return null;
```
Calvin-L, line 451: [`java.util.Arrays.binarySearch()`](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#binarySearch-java.lang.Object:A-java.lang.Object-) already implements binary search over arrays. Is there a good reason to use a homebrewed implementation here?

lemmy:
Switching to `Value#compareTo` for binary search revealed that `ModelValue#compareTo` differs from `ModelValue#equals` (`compareTo` compares model values of different type):

```tla
------------------ MODULE Github696 -------------------
EXTENDS Integers, TLCExt

VARIABLE v

Init == /\ v = {TLCModelValue("A_a"), TLCModelValue("B_a")}
\*        /\ TLCModelValue("B_a") \in v     \* TLC throws an error if this is uncommented. 
 
Next == UNCHANGED v

Inv == TLCModelValue("B_a") \in v           \* No problem in an invariant because v has been normalized and equals doesn't get used by `ValueVec#search`.

====
```

#### Typed Model Values

One way that TLC finds bugs is by reporting an error if it tries to compare two incomparable values&mdash;for example, a string and a set. The use of model values can cause TLC to miss bugs because it will compare a model value to any value without complaining (finding it unequal to anything but itself). Typed model values have been introduced to solve this problem.

For any character &tau;, a model value whose name begins with the two-character string "&tau;\_" is defined to have type &tau;. For example, the model value _x\_1_ has type _x_. Any other model value is untyped. TLC treats untyped model values as before, being willing to compare them to anything. However, it reports an error if it tries to compare a typed model value to anything other than a model value of the same type or an untyped model value. Thus, TLC will find the model value _x_1_ unequal to the model values _x_ab2_ and _none_, but will report an error if it tries to compare _x\_1_ to _a\_1_.

From: https://lamport.azurewebsites.net/tla/current-tools.pdf

lemmy:
Note to self: When done, delete bare-metal instance at https://console.equinix.com/devices/a43f8238-9cfb-492d-8421-eecb0780d2e0

Calvin-L:
The changes and justification all look good to me. 👍

lemmy:
@craft095 Please let me know if you're going to do another review. If not, I will get this merged.

craft095:
@lemmy I will do review today. Thank you!

tlatools/org.lamport.tlatools/src/tlc2/value/impl/ModelValue.java
```diff
@@ -131,13 +131,48 @@ public static void setValues() {
...
+		if (obj instanceof ModelValue) {
+			ModelValue mobj = (ModelValue) obj;
+			if ((mobj.type == this.type) || (mobj.type == 0)) {
+				return this.val.compareTo(((ModelValue) obj).val);
+			} else {
+				Assert.fail("Attempted to compare the differently-typed model values "
+						+ Values.ppr(this.toString()) + " and " + Values.ppr(mobj.toString()), getSource());
+			}
+		}
+		Assert.fail("Attempted to compare the typed model value " + Values.ppr(this.toString())
```
craft095, line 173: This prohibits comaprisons like MV("A_x").compareTo(StringValue("SomeString"))
However, StringValue("SomeString").compareTo(MV("A_x")) [is not prohibted](https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/src/tlc2/value/impl/StringValue.java#L48). That's true not only for StringValues.
Is it intended?

PR #716, opened 2022-02-17
WIP: Purge UniqueString

zwergziege:
Took a comment (https://github.com/tlaplus/tlaplus/pull/685#issuecomment-959782110) to heart and tried to get rid of `UniqueString`. If this is a refactoring that is too risky / too much work to be considered, just close the PR. If not, there are still some issues for which I could use some help / input.

I tried to carve out `UniqueString` step by step. In essence, I replaced all occurrences of `UniqueString` with `String` and all instances of `==` involving `UniqueString` with `.equals` (without null checks though). For the order of state variables, I created a static map. Ideally, this will be removed completely at some point in the future. The checkpoint code of `UniqueString` was removed without replacement since the order of the SVs is inferred while parsing the spec.

Edit: Checkpoint creation definitely needs another look. Distributed TLC probably as well.

Note: While I already found some bugs in the refactoring, I did not take the time to create a new history with patched commits instead of erroneous ones. For this sake, I use square brackets to denote checking out a commit and applying patches. 

**Intermediate bugs / patches**

Description: ModelValue.fingerprint has an infinite loop
Introduced by:  Changed ModelValue to use String internally c47f8d55d48b8b0f19e96112b22e92cc786d00c3
Fix:  fixed infinite recursion in ModelValue 075a8b6ed8b15bcd232c5515d7dd4ded8d4449c0

Description: Context.walkGraph casts a key of this.table to UniqueString. Keys have been changed to be String | ModuleName
Introduced by:  make sany context use String internally and deprecate functions that use US 0ffc03a7e65c1ea7249044b4a70ce6278c7cb7fe
Is resolved in: Removed deprecated functions and purged UniqueString from a few files c3bbb3ed9ea448beadc1798eb0575ef3e7ea80f1

Description: Location.name can be null which leads to problems on inclusion and equality checking
Introduced by: Purged from Location 67dda5ed4b1a3eb328f54157bd162b2f3eeea97a
Fix:  Assure Location.name != null f7bf3696755cf21a7181616648efc8432eaeb4dc

Description: Mixed up calls in OperatorStack.reducePostfix()
Introduced by:  SyntaxTreeNode Filename 58fdb46ea9597f037245123efc493a968cbd81c1
Fix: fixed wrong filename of SyntaxTreeNode 9c06db1317c8d1de533ea47de64c25f4ec2104f9 [1] for short

**Failing Tests**

I used the normal test suite for reference, but my master was also a bit outdated and thus the list of failing tests is probably incomplete. 

The following failures are caused by: Let StringValue use String internally 4569d2718e097e45ffc4a00ccdbdeedd3887941d  [2]
and can be further tracked to not every StringValue.val getting UniqueStringed and the different results on compare (convert to UniqueString in ctor, compareTo and equals (?) for a "fix", [2+fix])
tlc2.tool.liveness.CodePlexBug08EWD840FL1Test
tlc2.tool.liveness.CodePlexBug08EWD840FL2Tests
tlc2.tool.liveness.CodePlexBug08EWD840FL3Test
tlc2.tool.liveness.CodePlexBug08EWD840FL4Test
tlc2.tool.liveness.OneBitMutexNoSymmetryTest

the same holds for those, but I'm pretty positive that they are not indeed errors and can be fixed simply by changing the order of elements in the expected output (i.e. the order in which states are explored / values are printed has changed):
tlc2.tool.EvalExceptionTest
tlc2.tool.PrintTraceRaceTest
tlc2.tool.PrintTraceRaceTest_TTraceTest
tlc2.tool.liveness.CodePlexBug08AgentRingTest
tlc2.debug.EWD840DebuggerTest
tlc2.debug.EWD998TraceDebuggerTest

I'm guessing similar things happen for
pcal.SBBTest
but I'm not 100% sure because the number of generated states varies, which seems like a bug...

tlc2.tool.liveness.CodePlexBug08EWD840FL2FromCheckpointTest probably fails due to broken a broken checkpoint (which could be fixed) and then exhibits the same behavior as the tests above (i.e. works on [2+fix])

tlc2.debug.EWD840DebuggerSimTest 
is first caused by SyntaxTreeNode Filename 58fdb46ea9597f037245123efc493a968cbd81c1 which has been fixed but still shows problems similar to the tests above (i.e. works on [2+fix+1])

tlc2.debug.EWD998ChanDebuggerTest same as above but reordering is also a problem for records. those have been changed before Let StringValue use String internally 4569d2718e097e45ffc4a00ccdbdeedd3887941d

For the following tests, I have not had the time to look for the root causes.

tlc2.tool.TLCExtTraceSimTest

pcal.SBBTest_TTraceTest
tlc2.tool.liveness.CodePlexBug08AgentRingTest_TTraceTest
tlc2.tool.liveness.CodePlexBug08EWD840FL1Test_TTraceTest
tlc2.tool.liveness.CodePlexBug08EWD840FL2Test_TTraceTest
tlc2.tool.liveness.CodePlexBug08EWD840FL3Test_TTraceTest
tlc2.tool.liveness.CodePlexBug08EWD840FL4Test_TTraceTest
tlc2.tool.liveness.OneBitMutexNoSymmetryTest_TTraceTest


lemmy:
Thanks, let's find out if this makes TLC faster: https://github.com/tlaplus/tlaplus/actions/workflows/perf.yml

lemmy:
Performance test failed with:

```
Semantic processing of module Integers
Semantic processing of module IOUtils
Semantic processing of module measure
Starting... (2022-02-18 00:[33](https://github.com/tlaplus/tlaplus/runs/5239664920?check_suite_focus=true#step:7:33):25)
The exception was a java.lang.NoClassDefFoundError
: util/UniqueString
Finished in 00s at (2022-02-18 00:33:25)
```

https://github.com/tlaplus/tlaplus/runs/5239664920?check_suite_focus=true

zwergziege:
I will have to take a closer look at this, but I suspect this to be due to references to UniqueString in the community modules.

lemmy:
I like that the CommunityModules look much cleaner and clearer without UniqueString; this makes implementing module overrides easier.  That alone, however, doesn't justify this major refactoring.  Thus, let's check the performance benefit or penalty of this change.  For the moment, I suggest adding `CommunityModules.jar` and `tla2tools.jar` to this PR and hacking `.github/workflows/perf.yml` in this PR accordingly.  Alternatively, we could add two optional inputs to `perf.yml` that override the location where `perf.yml` gets `tla2tools.jar` and `CommunityModules-deps.jar`.

zwergziege:
Amongst other things, I added the workflow hack and CM jar. However, when I ran `measure.tla` the TLC instances created from within that spec (that's sick btw) all returned errors, but I also got them with the jars generated from `master`, so I don't know if that's an issue... Also, I have never worked with github workflows, so maybe I'm missing something.
I played around with the measurement script a little bit and found that the purged version tends to be slower. Sometimes by quite a heavy margin (~3 or smth). Thus, I'm not too confident that this is actually worth it. Maybe I can check where performance is lost with flame graphs or smth in the future. I assume it has smth to do with context lookups but am not entirely sure. Apparently, fast inequality is not enough.
Concerning the remaining errors in the standard tests: Are the liveness counterexamples guaranteed to be of minimal length? If not maybe there are only reorder issues left.

lemmy:
Btw. feel free to share constructive criticism, but please don't call other people's work "sick".

zwergziege:
Oh.. I'm sorry! I didn't mean to sound derogatory. I just forgot for a sec that the semantics of the word 'sick' are extremely context dependent. I actually love it :D

lemmy:
Results from the performance benchmark:

|Spec        | Workers|increase |
|:-----------|-------:|:--------|
|EWD840      |       4|0%       |
|EWD840      |       8|0%       |
|Grid5k      |       2|-25%     |
|PaxosCommit |       2|-41%     |
|PaxosCommit |       4|-43%     |
|Bookkeeper  |       2|-47%     |
|SwarmKit    |       2|-54%     |
|SwarmKit    |       4|-55%     |
|Bookkeeper  |       4|-58%     |
|MongoRepl   |       2|-60%     |
|SwarmKit    |       8|-79%     |
|Grid5k      |       4|-86%     |
|PaxosCommit |       8|-88%     |
|Bloemen     |       2|-123%    |
|EWD840      |       2|-136%    |
|MongoRepl   |       4|-170%    |
|Bookkeeper  |       8|-259%    |
|Ghostferry  |       2|-262%    |
|Grid5k      |       8|-321%    |
|Bloemen     |       4|-585%    |
|MongoRepl   |       8|-737%    |
|Bloemen     |       8|-1 066%  |
|Ghostferry  |       4|-1 296%  |
|Ghostferry  |       8|-1 949%  |

zwergziege:
I assume the column `increase` gives the increase in performance and thus negative numbers are bad?
Correct me if I'm wrong, but I'm under the impression that `PaxosMadeSimple` also crashes in previous specs.

lemmy:
Yes, negative is bad.

Re `PaxosMadeSimple: 69c931f738f13e8930a6a586dc6f302ec0fb6e26

PR #741, opened 2022-06-24
Add Transit module and add it as a trace output format

pfeodrippe:
[Transit](https://github.com/cognitect/transit-java) is a format that allows us to store (or transmit) arbitrary data, it's extensible and can represent any type of data by leveraging transit tags. It encodes the data in a non-readable human format using message pack.

Like the `json` format, the use case here is that the trace can be easily read by programs written in Java, Clojure or other languages, but different from `json`, it can represent sets natively.

In theory, we could extend it to support custom formats directly mapped to TLA+ types (e.g. a `FcnRcdValue` could be serialized and then deserialized as a `FcnRcdValue`, we would just miss anything that cannot be converted deterministically), but I find it better to make it simple now and create a stronger one if there is a use case for it later.

Let me know if this is something we want here.


lemmy:
@pfeodrippe Thanks for the PR.  I will look at it as soon as possible, which might be a while. :-(

lemmy:
@pfeodrippe I would lean toward moving most of this PR (sans `TLC.java`) into the CommunityModules to minimize bloat of `tla2tools.jar`.  What do you think?

pfeodrippe:
> @pfeodrippe I would lean toward moving most of this PR (sans `TLC.java`) into the CommunityModules to minimize bloat of `tla2tools.jar`. What do you think?

Hi Markus, if it's not to bloat here, I could add it to a separate repo and people could donwload it independently using https://github.com/pfeodrippe/tladeps, wdyt? 

With this I also would be free to use a non-Java language (like Clojure which is a superset of EDN, the data format which inspired transit) and would be native to deal with Jensen/Elle histories. It would also be much more faster and maintainable to do like this, one, because of the expressive of the language and, two, it's a separate project, showing that people can also contribute to TLA+ community without having to add it to the community modules repo.

pfeodrippe:
Whatever solution, we would need to add a hook here to be able to add formats dynamically to the CLI.

lemmy:
Is there a problem with contributing to the CommunityModules?   A hook could be added to TLC.

pfeodrippe:
> Is there a problem with contributing to the CommunityModules? A hook could be added to TLC.

No, I guess not, it's that the code there has to be Java and it will become a bloated project over time. But I can move this to there o/

pfeodrippe:
> > Is there a problem with contributing to the CommunityModules? A hook could be added to TLC.
> 
> No, I guess not, it's that the code there has to be Java and it will become a bloated project over time. But I can move this to there o/

I mean, technically there is not, but you are forced to use tools that the community modules gives to you for local development. It's like there is no other easy way to extend the TLA+ other than there and here

PR #755, opened 2022-08-22
Fix OpRcdValue initialization

Calvin-L:
TLC allows configuration files to define partial implementations for constant operators:

```
CONSTANT
    F(0) = 1
    F(1) = 2
    F(2) = 2
```

Because of the way values are initialized, this functionality does not work in current `master`. This PR fixes the problem by defining a custom `initialize()` method for `OpRcdValue` that does not call `fingerPrint()`.

It also adds a regression test to help us preserve this functionality in the future.


lemmy:
LGTM

PR #758, opened 2022-09-10
Fix some instances of string comparison using ==

Calvin-L:
In Java, == compares pointers.  Two `String` instances can have the same characters but live at different locations in memory.  Therefore, strings should always be compared using `a.equals(b)`, not `a == b`.

There are exceptions to the rule: if two strings are "interned" (using `str.intern()`) then they can be safely compared using ==.  All string literals written in the program text are guaranteed to be interned, but strings from user input are not, even if they are equal to a string literal.

Since there are no calls to `.intern()` in the TLA+ source code that I can find, we ought to fix these dangerous comparisons.  However, it is worth noting that these have not caused any (reported) issues so far, and may actually be safe if the strings being compared always point to program string literals.

There may be instances of this pattern that I have not caught.  I identified the locations to fix using IntelliJ's built-in string comparison inspection.  There are limitations to this analysis! If two strings are cast to `Object` before the comparison, the analyis will not detect a problem.

This issue was originally reported and fixed by @ElliotSwart in pull request #756.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/pcal/PcalSymTab.java
+++ b/tlatools/org.lamport.tlatools/src/pcal/PcalSymTab.java
@@ -571,7 +571,7 @@ public class PcalSymTab {
     }
         
     private void ExtractVarDecl(AST.VarDecl ast, String context) {
-        int vtype = (context == "") ? GLOBAL : PROCESSVAR;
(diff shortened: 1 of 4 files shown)


lemmy:
LGTM

Issue #766, opened 2022-10-04
Trace is truncated in invariant


dariusf:
## Description

Sorry, another bug. 😅

It's not related to -simulate, and it doesn't happen when indexing the result of `ToTrace(CounterExample)` in a POSTCONDITION.

## Expected Behavior

```
[i \in 1 .. Len(Trace) |-> Trace[i]]
```
should be equal to Trace.

## Actual Behavior

The second-last element is missing:

```
Error: Evaluating invariant JsonInv failed.
Attempted to access index 5 of tuple
<<[x |-> 1], [x |-> 2], [x |-> 3], [x |-> 5]>>
which is out of bounds.
```

## Steps to Reproduce

```sh
java -cp tla2tools.jar -XX:+UseParallelGC tlc2.TLC Counter_json.tla
```

<details>
<summary>Inputs</summary>

```tla
---- MODULE Counter ----
EXTENDS Naturals

VARIABLE x
vars == <<x>>

Init == x = 1

Next == x' = x + 1

Spec == 
    /\ Init
    /\ [][Next]_vars

Inv == x < 5

====
```

```tla
------------------------------- MODULE Counter_json -------------------------------
EXTENDS Counter, TLC, TLCExt, Sequences

JsonInv ==
    \/ Inv
    \/
      LET Trace1 == [i \in 1 .. Len(Trace) |-> Trace[i]] IN
      /\ PrintT(Trace)
      /\ PrintT(Len(Trace))
      /\ PrintT(<<Trace[1], Trace[2], Trace[3], Trace[4], Trace[5]>>)
      /\ PrintT(Trace1)

=============================================================================
```

```
SPECIFICATION Spec
INVARIANT JsonInv
```

</details>

<details>
<summary>Output</summary>

```
TLC2 Version 2.18 of Day Month 20?? (rev: ${git.shortRevision})
Running breadth-first search Model-Checking with fp 24 and seed -3919750552308392028 with 1 worker on 8 cores with 3641MB heap and 64MB offheap memory [pid: 7750] (Mac OS X 12.5.1 x86_64, Homebrew 18.0.2 x86_64, MSBDiskFPSet, DiskStateQueue).
Parsing file /private/tmp/aa/Counter_json.tla
Parsing file /private/tmp/aa/Counter.tla
[...]

PR #770, opened 2022-10-20
Fix -dumpTrace output when initial state violates invariant

dariusf:
Thanks for the patch in #766! I did the same thing for -simulate and added a test.


lemmy:
Thanks for the PR.  Could you please also create a test case (copy of [Github766Test.java](https://github.com/tlaplus/tlaplus/pull/770/commits/867789b74f51de09e8d6e2990c1ab027505546e0#diff-aae6f51cec85d3e395af4a9f0333f4c3c658f828d87ed0516a0fec1a1902c0f2)) for model-checking mode?

dariusf:
Added. There are tests for both now.

PR #780, opened 2023-01-25
Add File.separator after BASE_DIR wherever it is used

Calvin-L:
This change slightly simplifies the requirements for running the TLA+ tools tests.  Previously, BASE_DIR (read from the "basedir" system property) required a trailing separator character (e.g. "/" on Linux) to function correctly.

With this change the separator is no longer required, which simplifies running the tests from custom IDEs or other build systems.

In particular, note that Maven sets the "basedir" system property to a normalized path while running tests, meaning that it strips off trailing separators.  This behavior is undocumented and there is no way to override it.


lemmy:
LGTM

Issue #778, opened 2023-01-10
Improve contributor experience


lemmy:
- [x] Tag issues with new ["good first issue"](https://github.com/tlaplus/tlaplus/labels/good%20first%20issue) label
- [ ] Separate TLA+ Tools and Toolbox build (ant & maven)
- [x] Documentation how to set up a lightweight TLC development environment and [full-blown Toolbox dev environment](https://github.com/tlaplus/tlaplus/tree/master/general/ide)
- [x] Polish and extend https://github.com/tlaplus/tlaplus/blob/master/README.md
- [x] Convert (?) [contributor guide](https://github.com/tlaplus/tlaplus/blob/master/CONTRIBUTING.md) into ]Github issue template](https://github.com/tlaplus/tlaplus/tree/master/.github/ISSUE_TEMPLATE)
- [ ] Convert [feature wishlist](https://github.com/tlaplus/tlaplus/blob/master/general/docs/contributions.md) into a more accessible format such as issues, projects, ... ?
- [ ] Document [performance test framework](https://github.com/tlaplus/tlaplus/actions/workflows/perf.yml)
- [ ] ...

/cc @konnov @Calvin-L @ZacAttack @thesammiller @muenchnerkindl @xxyzzn

thesammiller:
I've added a PR (https://github.com/tlaplus/tlaplus/pull/791) for how to build TLC on the command line with ant. Not sure if it's too lightweight for "how to setup a lightweight TLC development environment", but happy to generate a longer document if that would be more helpful with additional context. If so, please just let me know where it should go.

Calvin has also provided instructions for setting up in IntelliJ. Can that be added to the `ide` file or should that exist somewhere else?

lemmy:
Closing in favor of individual issues.

PR #782, opened 2023-01-28
Add documentation and code quality improvements to the wishlist

Calvin-L:
(Finally) following up on #756, here is a bit of writing about documentation and code quality improvements we're looking for.

Did I miss any? Or include any that we actually want to discourage?

I considered including one about improving the build system as well (see e.g. the separation bullet in #778). However, improving/rewriting the tools build system seems like a complex job to me, and I wasn't sure exactly what I could put in the wishlist that would be useful.


lemmy:
LGTM, especially the  "cosmetic problams" :-)

Calvin-L:
That's the funniest mistake I've made in a while! 🤣 Let me fix that and then get this merged...

Issue #790, opened 2023-02-08
<>TRUE => <>[]TRUE does not hold


eras:
## Description

I would expect property `<>TRUE => <>[]TRUE` to hold true, however this is not the case with versions 753a970 (the latest Git commit as of writing) or ca5ad64 (an older version I happened to have a around).

It was true in version 4653876 (an even older version I had).

I have retrieved these versions from the GitHub release page.

As the spec is short, I'll just describe it here:
```
VARIABLES a
Init == a = 0
Next == UNCHANGED a
Spec == Init /\ [][Next]_a
AlwaysTrue == <>TRUE => <>[]TRUE
```
And in .cfg there is `PROPERTY AlwaysTrue`.

## Expected Behavior
I expect the property to hold.

## Actual Behavior
The property does not hold.

## Steps to Reproduce
An easy way to reproduce this is to get the repo https://github.com/eras/tla-issue/ branch issue790 (some irrelevant stuff snipped):

```
% git clone -b issue790 https://github.com/eras/tla-issue/
Cloning into 'tla-issue'...
% cd tla-issue 
% tlc True.tla
TLC2 Version 2.18 of Day Month 20?? (rev: ca5ad64)
...
Implied-temporal checking--satisfiability problem has 1 branches.
Computing initial states...
Finished computing initial states: 1 distinct state generated at 2023-02-08 20:57:16.
Error: Temporal properties were violated.

Error: The following behavior constitutes a counter-example:

State 1: <Initial predicate>
a = 0

2 states generated, 1 distinct states found, 0 states left on queue.
The depth of the complete state graph search is 1.
```

## Steps Taken to Fix
I tried the few versions of TLAToolbox releases I had.

## Possible Fix
The exact commit introducting this regression probably easy to find with `git bisect`..

## Your Environment
 - TLC version: various
 - Operating System: Debian GNU/Linux bookworm/sid

lemmy:
@eras Thanks for reporting.  Can you bisect and check if either 1b854688226c3d67acb775435c53318512ee98f7 or 773e75e481b4c1c958818da14b076da44c0f32e8 changed TLC's behavior?

eras:
Is there some way to compile TLC from command line? Just the tla2tools.jar with TLC, for example.

I tried to compile the whole thing with maven but failed due to

```
[ERROR] Cannot resolve target definition:
[ERROR]   Software being installed: org.eclipse.contribution.weaving.feature.group 2.2.4.202007241909
[...]

PR #791, opened 2023-02-14
Add TLC command-line build instructions

thesammiller:
Provides general instructions for building TLC without the Toolbox, using ant on the command line.

Based on the conversation in https://github.com/tlaplus/tlaplus/issues/790, I'm adding to `CONTRIBUTING.md`. I'm keeping the instructions short to address the need of having these commands somewhere. Would a longer version be helpful? Would there be a better file to put it in?

The `ide` instructions mention OpenJDK 14, but this approach was tested and confirmed working with OpenJDK 11. Please let me know if that should be updated to 14.

Directions thanks to @lemmy and @Calvin-L. 

Thank you!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,3 +38,23 @@ Quality Metrics
 ---------------
 
 We collect [quality metrics](https://sonarcloud.io/organizations/tlaplus/projects). If you want to help out with the project, the reports indicate several low hanging fruits to pick.
+
+Building TLC Without the Toolbox
+---------------------------------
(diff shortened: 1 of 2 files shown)


thesammiller:
Looks like some of this information is in [the internal developer documentation] (https://github.com/tlaplus/tlaplus/blob/master/DEVELOPING.md ).

Not sure how this document compares/relates to `DEVELOPING.md`, but should maybe add a link to `DEVELOPING.md` in `CONTRIBUTING.md`. It didn't occur to me until reading the changes in https://github.com/tlaplus/tlaplus/pull/782

lemmy:
It's probably good if the new and the old `DEVELOPING.md` are consolidated into a single document that is more easily discoverable by future contributors.

thesammiller:
Added a commit to address the comment by @lemmy. IDE instructions for setting up in Intellij are still forthcoming and could be added to this PR or in a new one.

Calvin-L:
Thanks for doing this. :)

PR #793, opened 2023-02-17
Add a docstring to IValue

Calvin-L:
The `Value` classes are very important, but there isn't much documentation about them. I've tried to capture some of the key design points and subtleties on the top-level `IValue` interface.


lemmy:
LGTM, the build breakage is unrelated: https://github.com/tlaplus/tlaplus/issues/794

PR #799, opened 2023-03-16
LazyValue can cause TLC to generate too many or too few states.

lemmy:
Part of Github issue #798
https://github.com/tlaplus/tlaplus/issues/798

[Bug][TLC][Changelog]

PR #800, opened 2023-03-18
LazyValue refactor

### LINKED ISSUE #798: LazyValue can cause TLC to generate too many or too few states
## Description

The model checker generated empty states which did not meet expectations.
I wrote a simple TLA+ specfication, https://github.com/ybbh/tla_state.
When I used the model checker to check it, it generated empty initial states.

## Expected Behavior

Model checker should generate non-empty initial states.

## Actual Behavior

Model checker generated empty initial states.

## Steps to Reproduce
1. git clone git@github.com:ybbh/tla_state.git
2. Toolbox: File -> Open Spec -> Add New Spec -> open state.tla
3. Toolbox: models -> New Model -> filling Temporal formula with **Spec** -> setting constants NODE_ID by **{"n1", "n2"}**
4. Toolbox: Run TLA on the model
5. Toolbox: The model checker result generated empty initial states
result

## Steps Taken to Fix

1. When I changed the constants NODE_ID by **{"n1"}**, it generated correct result.

2. The initial states would be OK if I uncomment this line in [state.tla](https://github.com/ybbh/tla_state/blob/main/state.tla).
  
    ```
      __WhyWeNeedThis == S2
    ```

## Possible Fix

Possible when producing the initial state, the model checker threw some unknown exceptions.

## Your Environment

 - TLC version: 2.15, 2.16
 - version: 1.8.0.202303090234, 1.7.1.202012311918
 - Operating System: Windows 10

Calvin-L:
This is a small(ish) refactor to the way that `LazyValue` works. It fixes #798, although I want to clarify that I'm not certain this is the right approach. I'm more interested in discussing whether this is a good direction to push. If we decide this is worth doing, I believe it is also compatible with the changes in #799.

**Key idea:**

TLC has had several bugs involving misuse of `LazyValue.val`. Instead of proving every use of that field correct, this change shifts the management of the field to the `LazyValue` class itself. Now callers _must_ specify the behavior and evaluation mode that they are evaluating with when reading the cached value. The cached value is only returned if it was computed using the same behavior and evaluation mode. If it wasn't, it must be recomputed.

**Pros:**

A lot of complexity can be removed without losing any potential benefits of `LazyValue`.

**Cons:**

There are some test failures related to interaction between the debugger and `LazyValue`. I believe they are relatively benign. Also, the `LazyValue` class now stores several more fields. It is possible that this will worsen TLC's performance since many instances of this class are created during evaluation.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/EvalControl.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/EvalControl.java
@@ -1,9 +1,12 @@
 // Copyright (c) 2003 Compaq Corporation.  All rights reserved.
 // Portions Copyright (c) 2003 Microsoft Corporation.  All rights reserved.
+// Copyright (c) 2023, Oracle and/or its affiliates.
 // Last modified on Mon 30 Apr 2007 at  9:27:06 PST by lamport 
 //      modified on Fri Jan  4 22:46:57 PST 2002 by yuanyu 
 
 package tlc2.tool;
+
+import tlc2.util.PartialBoolean;
 
 public class EvalControl {
 
@@ -81,4 +84,46 @@ public class EvalControl {
 	public static boolean isConst(int control) {
 		return isSet(control, Const);
 	}
+
+    /**
+     * Determine whether two {@link EvalControl} settings are semantically equivalent.  Formally,
+     * they are semantically equivalent if
+     * <pre>
+     *     \A expr, context, behavior:
+     *         eval(expr, context, behavior, control1) =
+     *         eval(expr, context, behavior, control2)
+     * </pre>
+     *
+     * <p>Many control settings like {@link #Clear} and {@link #Enabled} are not equivalent; they
+     * will often result in different computed values.
+     *
+     * <p>This function always returns {@link PartialBoolean#YES} when
+     * <code>control1 == control2</code>, and it can identify a small set of other cases where
+     * the inputs are semantically equivalent.
+     *
+     * @param control1 the first control value
+     * @param control2 the second control value
+     * @return whether the values are semantically equivalent
+     */
+    public static PartialBoolean semanticallyEquivalent(int control1, int control2) {
+        // *** CAUTION ***
+        // The implementation of this function is quite subtle.  First we'll define `flagsThatCanBeSafelyIgnored`
+        // as a special whitelist of flags we know won't affect the evaluation outcome.  Each whitelisted flag
+        // has to be carefully justified:
+        //   - KeepLazy: this is a performance hint that affects handling of function definitions.  Although it
+        //     can change the exact structure of the IValue returned by eval(...), it does not affect the semantic
+        //     meaning of that value.
+        //   - Init: this is a flag to indicate that we are evaluating part of a specification's initial condition.
+        //     It is used for debugging.
+        //   - Const: similar to Init, this is a flag to indicate that we are evaluating part of a specification's
+        //     constant definitions.  It is used for debugging.
+        int flagsThatCanBeSafelyIgnored = KeepLazy | Init | Const;
+
+        // Compute a mask capturing all possible flags that CAN'T be safely ignored.
+        int mask = ~flagsThatCanBeSafelyIgnored;
+
+        // If the two inputs are identical on all flags not present in `flagsThatCanBeSafelyIgnored`, then they are
+        // certainly equivalent.  Otherwise, conservatively return `MAYBE`.
+        return (control1 & mask) == (control2 & mask) ? PartialBoolean.YES : PartialBoolean.MAYBE;
+    }
 }
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/TLCStateFun.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/TLCStateFun.java
@@ -1,5 +1,6 @@
 // Copyright (c) 2003 Compaq Corporation.  All rights reserved.
 // Portions Copyright (c) 2003 Microsoft Corporation.  All rights reserved.
+// Copyright (c) 2023, Oracle and/or its affiliates.
 // Last modified on Mon 30 Apr 2007 at 15:29:58 PST by lamport
 //      modified on Fri Jul 20 23:54:51 PDT 2001 by yuanyu
 
@@ -26,9 +27,9 @@ import util.WrongInvocationException;
  * can not be used in getInitStates and getNextStates.
  */
 public final class TLCStateFun extends TLCState {
-  private SymbolNode name;
-  private IValue value;
-  private TLCStateFun next;
+  private final SymbolNode name;
+  private final IValue value;
+  private final TLCStateFun next;
 
   public final static TLCState Empty = new TLCStateFun(null, null, null);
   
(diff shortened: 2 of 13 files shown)


lemmy:
The changes look good to me; however, to minimize risk, I suggest we proceed with a hotfix for https://github.com/tlaplus/tlaplus/issues/798, which either disables LV or is limited to the changes in PR https://github.com/tlaplus/tlaplus/pull/799. After addressing the issue with the hotfix, we can then merge this thereby allowing us to gain confidence in the solution.

lemmy:
LGTM

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/Tool.java
```diff
@@ -1818,107 +1816,22 @@ protected final Value evalApplImpl(final OpApplNode expr, Context c, TLCState s0
...
-					// this.enabled* methods pass EvalControl.Enabled (the only exception being the
-					// call on line line 2799 which passes EvalControl.Primed). This is why we can
-			    	// be sure that ENALBED expressions are not affected by the caching bug tracked
-			    	// in Github issue 113 (see below).
-					val = this.eval(lv.expr, lv.con, s0, s1, control, lv.getCostModel());
 				} else {
-					val = lv.getValue();
-					if (val == null) {
-						final Value res = this.eval(lv.expr, lv.con, s0, s1, control, lv.getCostModel());
-						// This check has been suggested by Yuan Yu on 01/15/2018:
```
lemmy, line 1833: I'd would suggest to keep the full comment for posterity.

Calvin-L:
[Initial benchmark results](https://github.com/tlaplus/tlaplus/actions/runs/4641862172) for this change look _very_ negative. (Er, at least as far as I can tell. I'm still not certain how to convert the numbers there into insights.)

It may be that this approach is not worth the performance penalty.

lemmy:
> [Initial benchmark results](https://github.com/tlaplus/tlaplus/actions/runs/4641862172) for this change look _very_ negative. (Er, at least as far as I can tell. I'm still not certain how to convert the numbers there into insights.)
> 
> It may be that this approach is not worth the performance penalty.

stats.R is my very basic attempt at reporting: https://github.com/tlaplus/tlaplus/blob/b81af62f2e03ba358f2e46100564865b297c0352/general/performance/stats.R#L1-L44

Calvin-L:
Yes! I should be able to take a look today.

Calvin-L:
@lemmy This fixes #817, but only if we revert the no-chains-of-LazyValue fix. See [1e5c3cf](https://github.com/tlaplus/tlaplus/pull/800/commits/1e5c3cfa2ffc74eed159a68acd7738c53c9354b6), now included in this PR.

lemmy:
Sounds to me that we want this PR instead of the no-chains-of-LazyValue.

Calvin-L:
I'm inclined to agree---but I'm not excited about the performance impact. 😓

lemmy:
I triggered another performance run to get more reliable data.

Calvin-L:
This is hardly the simple refactor I imagined---but it does seem to fix some problems and I think I can live with the [latest performance results](https://github.com/tlaplus/tlaplus/actions/runs/5284346122):
```
                Spec |   W |  States/s (old) |  States/s (new) |     Change
--------------------------------------------------------------------------------
            SwarmKit |   4 |            7815 |           10647 |       +36%
            SwarmKit |   2 |            4482 |            6098 |       +36%
            SwarmKit |   8 |           12511 |           16763 |       +34%
          Ghostferry |   8 |          369276 |          401027 |        +9%
              EWD840 |   4 |         3544216 |         3555334 |          -
              Grid5k |   2 |         1991196 |         1995004 |          -
          Ghostferry |   4 |          274617 |          273576 |          -
              Grid5k |   4 |         2887778 |         2839275 |        -2%
             Bloemen |   8 |          465941 |          457665 |        -2%
             Bloemen |   4 |          362490 |          354873 |        -2%
             Bloemen |   2 |          208378 |          203733 |        -2%
              EWD840 |   2 |         1953797 |         1891369 |        -3%
         PaxosCommit |   4 |          216087 |          207940 |        -4%
         PaxosCommit |   2 |          120815 |          116038 |        -4%
         PaxosCommit |   8 |          346973 |          332761 |        -4%
          Ghostferry |   2 |          153229 |          146474 |        -4%
          Bookkeeper |   8 |          161369 |          142828 |       -11%
          Bookkeeper |   2 |           69828 |           59213 |       -15%
          Bookkeeper |   4 |          115211 |           97436 |       -15%
           MongoRepl |   2 |          146430 |          120030 |       -18%
           MongoRepl |   4 |          264131 |          215329 |       -18%
           MongoRepl |   8 |          432427 |          348341 |       -19%
     PaxosMadeSimple |   4 |           80872 |           64795 |       -20%
     PaxosMadeSimple |   2 |           44389 |           35171 |       -21%
     PaxosMadeSimple |   8 |          134780 |          104641 |       -22%
```

It would be nice to know what causes such a big swing in e.g. PaxosMadeSimple (which suffers under this change) and SwarmKit (which wins a lot under this change).

lemmy:
Maybe comparing the coverage results of Paxos/SwarmKit before and after the recent changes could provide some insight into the situation.

UnrealEugene:
This refactor of LazyValue allows for easy caching of definitions in LET ... IN expressions (even state level ones). I tried to change `LetInKind` case in `Tool#getNextStatesImpl` method from this:
```java
case LetInKind:
  {
    LetInNode pred1 = (LetInNode)pred;
    return this.getNextStates(action, pred1.getBody(), acts, c, s0, s1, nss, cm);
  }
```
to this:
```java
case LetInKind:
  {
    LetInNode pred1 = (LetInNode)pred;
    OpDefNode[] letDefs = pred1.getLets();
    Context c1 = c;
    for (OpDefNode opDef : letDefs) {
      if (opDef.getArity() == 0) {
        Value rhs = new LazyValue(opDef.getBody(), c1, cm);
        c1 = c1.cons(opDef, rhs);
      }
    }
    return this.getNextStates(action, pred1.getBody(), acts, c1, s0, s1, nss, cm);
  }
```
On a few of my specifications which use a lot of LET IN expressions this resulted in ~x3 faster model checking. Although I'm not sure if I've done it correctly, this definitely needs to be considered as an addition. Also there may be other places where this optimization could be implemented.

lemmy:
@UnrealEugene If you open a PR, we should be able to get performance numbers from our benchmarks.

lemmy:
+1 for merging the LazyValue refactoring despite the performance regressions.

Calvin-L:
Tomorrow I will have some time to clean up this branch and get it merged. ⛵

tlatools/org.lamport.tlatools/src/tlc2/tool/EvalControl.java
```diff
@@ -81,4 +84,46 @@ public static boolean isInit(int control) {
...
+    public static PartialBoolean semanticallyEquivalent(int control1, int control2) {
+        // *** CAUTION ***
+        // The implementation of this function is quite subtle.  First we'll define `flagsThatCanBeSafelyIgnored`
+        // as a special whitelist of flags we know won't affect the evaluation outcome.  Each whitelisted flag
+        // has to be carefully justified:
+        //   - KeepLazy: this is a performance hint that affects handling of function definitions.  Although it
+        //     can change the exact structure of the IValue returned by eval(...), it does not affect the semantic
+        //     meaning of that value.
+        //   - Init: this is a flag to indicate that we are evaluating part of a specification's initial condition.
+        //     It is used for debugging.
```
lemmy, line 117: What do you mean by "It is used for debugging."?  This statement would be false if it refers to the TLA+ debugger.

tlatools/org.lamport.tlatools/src/tlc2/tool/TLCStateFun.java
```diff
@@ -64,10 +65,16 @@ public final boolean containsKey(UniqueString var) {
...
   
   public final TLCState copy() {
-      // The following code added blindly by LL on 28 May 2010
-      // to fix a bug.  I have no idea what's going on here.
-       return new TLCStateFun(this.name, this.value, this.next);
-      // throw new WrongInvocationException("TLCStateFun.copy: This is a TLC bug.");
+    // NOTE 2023/4/10: Since instances of this class are immutable, there is no need
+    // to create an actual copy.  In addition to being more performant, this fixes a
+    // latent bug: LL's code below can copy the `Empty` TLCState, creating "corrupt"
+    // instances of TLCStateFun that have null fields but are not equal to `Empty`.
```
lemmy, line 71: Good catch!

tlatools/org.lamport.tlatools/src/tlc2/tool/EvalControl.java
Calvin-L: You are right, there are a few things wrong with this comment. "is used" is vague, and it's not obvious what I was getting at here.

I think when I wrote that I had just looked at every place in the code where control flow depends on `EvalControl.isInit` and `EvalControl.isConst`. There are (I think) exactly two:
 - `TLCGetSet`, where `isInit` affects how the level of the state is computed
 - `DebugTool`, where `isInit` helps set the correct value for the `mode` field

The second one is the one I intended to point at with this comment.

`Init` and `Const` are "used" lots of places (in the sense that they are set during evaluation). However, as far as I could tell, they are only "used" in a very small number of places (in the sense that their value is inspected).

Anyway, I'm open to suggestions for how this can be better.

PR #804, opened 2023-03-24
Use a unique metaDir for each instance of TLC

### LINKED ISSUE #787: Running multiple instance of TLC on the same spec in parallel can cause filesystem clashes
## Description
When launching multiple instances of TLC concurrently from the same directory, some of the instances will fail with return code 1 since multiple instances all try to create the same states directory based on a timestamp. It is occasionally useful to run multiple instances of TLC in parallel instead of using the `-workers` command to parallelize TLC itself, for example if you are running a large number of small models in a script. I recall this also occurred when running tests on TLC itself.

## Expected Behavior
Multiple instances of TLC should be able to be launched concurrently from the same directory.

## Actual Behavior
Some instances of TLC exited with error code 1 and the following error:
```
Exception in thread "main" util.Assert$TLCRuntimeException: TLC writes its files to a directory whose name is generated from the current time.
This directory should be /home/ahelwer/src/tlaplus/examples/states/23-01-30-11-08-24.889, but that directory already exists.
Trying to run TLC again will probably fix this problem.
        at util.Assert.check(Assert.java:139)
        at util.FileUtil.makeMetaDir(FileUtil.java:257)
        at tlc2.TLC.handleParameters(TLC.java:1078)
        at tlc2.TLC.main(TLC.java:308)
```

## Steps to Reproduce
1. Clone tlaplus/examples repo and download tla2tools.jar to repo root
2. Run this bash script:
```sh
TLC_COMMAND="java -Dtlc2.TLC.stopAfter=5 -cp tla2tools.jar tlc2.TLC specifications/DieHard/DieHard"
$TLC_COMMAND &> output1.txt & pid1=$!
$TLC_COMMAND &> output2.txt & pid2=$!
$TLC_COMMAND &> output3.txt & pid3=$!
$TLC_COMMAND &> output4.txt & pid4=$!

wait $pid1
echo "pid1 exit code $?"
wait $pid2
echo "pid2 exit code $?"
wait $pid3
echo "pid3 exit code $?"
wait $pid4
echo "pid4 exit code $?"
```
Output will vary but one example is:
```
pid1 exit code 1
pid2 exit code 12
pid3 exit code 12
pid4 exit code 12
```
Where error code 1 means it hit the states dir failure, and output1.txt will show the states dir error message. Error code 12 is expected for this model.

## Steps Taken to Fix
This can be worked around with the `-metadir` parameter:
```sh
TLC_COMMAND="java -Dtlc2.TLC.stopAfter=5 -cp tla2tools.jar tlc2.TLC specifications/DieHard/DieHard"
eval "$TLC_COMMAND -metadir ./states1" &> output1.txt & pid1=$!
eval "$TLC_COMMAND -metadir ./states2" &> output2.txt & pid2=$!
eval "$TLC_COMMAND -metadir ./states3" &> output3.txt & pid3=$!
eval "$TLC_COMMAND -metadir ./states4" &> output4.txt & pid4=$!

wait $pid1
echo "pid1 exit code $?"
wait $pid2
echo "pid2 exit code $?"
wait $pid3
echo "pid3 exit code $?"
wait $pid4
echo "pid4 exit code $?"
```

## Possible Fix
Append a random nonce to the state dir name or, if the state dir exists, append an incrementing number to the name until it is created successfully.

## Your Environment
 - TLC version: TLC2 Version 2.18 of Day Month 20?? (rev: ad45483)
 - Operating System: Arch Linux

Calvin-L:
Fixes #787.

Previously, TLC's "metaDir" was determined using the current date with millisecond precision.  Two instances of TLC spawned in the same millisecond would race to create the same directory, and one would fail.

This commit allows both to succeed; one will create a directory with a slightly different name.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/util/FileUtil.java
+++ b/tlatools/org.lamport.tlatools/src/util/FileUtil.java
@@ -1,5 +1,7 @@
 // Copyright (c) 2003 Compaq Corporation.  All rights reserved.
 // Portions Copyright (c) 2003 Microsoft Corporation.  All rights reserved.
+// Copyright (c) 2023, Oracle and/or its affiliates.
+
 package util;
 
 import java.io.BufferedInputStream;
@@ -16,7 +18,10 @@ import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
 import java.math.BigInteger;
+import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.text.SimpleDateFormat;
 import java.util.Date;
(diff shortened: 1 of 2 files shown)


lemmy:
IIRC there is a second problem where TLC extracts some specs such as the standard libraries to $TMP_DIR.

Calvin-L:
I'm not quite sure how far to go. I fixed the ones I think are most likely to cause problems, but here are a few other places TLC might have similar problems:
 - `MonolithicSpecExtractor`, which calls `FileUtil.createTempFile` (multiple TLC instances might try to write to the same file in `java.io.tmpdir`)
 - `Simulator`, which writes some liveness state information directly to `java.io.tmpdir`
 - `DistributedFPSet`, which writes state information to a subdirectory of `java.io.tmpdir` named with the current millisecond
 - `MailSender`, which writes to some (hard-coded?) filenames in `java.io.tmpdir`

Any of these might cause problems with concurrent TLC runs---do we want to try fixing all of them now?

lemmy:
I don't think distributed and cloud TLC have many users, but I would assume `MonolithicSpecExtractor` and `Simulator` are worth it.

PR #814, opened 2023-05-05
Add IHooks interface

pfeodrippe:
The main idea for these hooks is to enable the user to do whatever they want, e.g. setting a state writer, dump, trace dump, log etc.

More hooks can be added over time.

This shouldn't interfere with the normal processing of TLC.

Let me know what you think o/

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/TLC.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/TLC.java
@@ -9,6 +9,7 @@ import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.lang.reflect.Constructor;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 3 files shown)


lemmy:
@pfeodrippe, could you please provide more details about the specific use cases? Additionally, is it feasible to consider using AspectJ for implementing such extensions?

pfeodrippe:
> @pfeodrippe, could you please provide more details about the specific use cases? Additionally, is it feasible to consider using AspectJ for implementing such extensions?

This would add points of execution that the user could call arbitrary code (so this would work with TLA+ VSCode extension). I see that AJ5 has some annotation-based approach (also something called point of execution), do you think that this could do the same thing (checking a Java property, dynamically loading a class out of it etc)?

One use case would be the user writing a configuration file (per folder) that custom code would read and, for example, set options (just like the command line or the VSCode extension config do, just with the flexibility the user needs, let’s say they need to read it from a S3 bucket or from a GH repo or from… etc). 

Others would be to add modules and import dependence’s dynamically or set custom state writers or opening a web browser (or a VSCode window) showing reports in real time or etc.

 I

pfeodrippe:
Let me know if I should close this o/

lemmy:
I don't think we should add hooks here. We can do the same thing, but even better, with AspectJ, and we won't need to touch TLC's code. We've done something similar with AspectJ in distributed TLC before. Check it out here: https://github.com/tlaplus/tlaplus/tree/master/tlatools/org.lamport.tlatools/src-aj"

pfeodrippe:
> I don't think we should add hooks here. We can do the same thing, but even better, with AspectJ, and we won't need to touch TLC's code. We've done something similar with AspectJ in distributed TLC before. Check it out here: https://github.com/tlaplus/tlaplus/tree/master/tlatools/org.lamport.tlatools/src-aj"

Got it, thanks!

Issue #819, opened 2023-06-16
Unexpected temporal formula error when universally quantifying over an empty set in liveness conditions.


klinvill:
## Description
Liveness checking a condition that universally quantifies over an empty set results in an error message that states:
> Temporal formulas containing actions must be of forms <>[]A or []<>A.

The error message doesn't make sense in the context of my liveness condition. Such a condition is a tautology so I would either expect it to work like other tautologies (for which liveness trivially holds) or would expect the error message to mention something about a tautology if that's why an error is thrown. A minimal example specification where this error occurs is given below. You can ignore the rest of the specification and focus only on the `Live` property and its component conditions. Specifically, including the `TestFailing` condition in the liveness property results in the observed error message. If `TestFailing` is commented out, then the liveness checking succeeds as expected given the `TestSucceeding` condition which is another tautology, but one that quantifies over a non-empty set.

The minimal example I have has this spec in `LivenessTest.tla`:
```tla
---- MODULE LivenessTest ----

EXTENDS Integers

CONSTANTS STEP_SIZE
  
VARIABLES i

TypeOK == i \in Int

Step == i' = i + STEP_SIZE

Init == i = 0
Next == Step

Spec == Init /\ [][Next]_i

Constraint == i < 5

--------------------------
(*** Liveness ***)

TestFailing ==
  \A x \in {} :
    []<>TRUE

TestSucceeding ==
  \A x \in {42} :
    []<>TRUE

Live ==
  \* This condition causes liveness checking to fail with the error:
  \*  "Temporal formulas containing actions must be of forms <>[]A or []<>A."
  \* 
  \* If you comment this condition out, you'll see that liveness checking
  \* succeeds.
  /\ TestFailing

  /\ TestSucceeding

==========================
```

And has this configuration file specified in `LivenessTest.cfg`:
```tla
CONSTANTS
  STEP_SIZE = 1

SPECIFICATION Spec

PROPERTY Live

INVARIANT TypeOK

CONSTRAINT Constraint
```

## Expected Behavior
Liveness checking should succeed without an error, or with a tautology related warning message, when using the `TestFailing` condition. I also expect the liveness checking to have the same result when only checking `TestFailing` as when only checking `TestSucceeding`.

## Actual Behavior
Liveness checking fails with the aforementioned error when checking liveness with the `TestFailing` condition.

## Your Environment
 - TLC version: TLC2 Version 2.18 of Day Month 20?? (rev: cc4177a)
[...]

PR #820, opened 2023-06-16
Unexpected temporal formula error when universally quantifying over an

lemmy:
empty set in liveness conditions.

https://github.com/tlaplus/tlaplus/issues/819

[Bug][TLC]


lemmy:
https://github.com/tlaplus/tlaplus/pull/820 causes this regression because the more generic level checking in #820 causes the level of the `[]<>P` in `\A x \in {}: []<>P` to be raised to temporal, which ultimately triggers the bogus error message.

Calvin-L:
The fix looks good. Do we need a similar change in the `BoundedExists` block above to handle formulas like
```
TestFailingE ==
  \E x \in {} :
    []<>TRUE
```
?

Calvin-L (approved):

PR #824, opened 2023-07-27
LazyValue refactor: use LazyValue to cache LetInKind definitions in Tool#getNextStatesImpl

UnrealEugene:
Part of LazyValue refactor (https://github.com/tlaplus/tlaplus/pull/800)
Relevant comment: https://github.com/tlaplus/tlaplus/pull/800#issuecomment-1654202331

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/EvalControl.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/EvalControl.java
@@ -1,5 +1,6 @@
 // Copyright (c) 2003 Compaq Corporation.  All rights reserved.
 // Portions Copyright (c) 2003 Microsoft Corporation.  All rights reserved.
+// Copyright (c) 2023, Oracle and/or its affiliates.
 // Last modified on Mon 30 Apr 2007 at  9:27:06 PST by lamport 
 //      modified on Fri Jan  4 22:46:57 PST 2002 by yuanyu 
 
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/EvalControl.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/EvalControl.java
@@ -5,6 +5,8 @@
 
 package tlc2.tool;
 
+import tlc2.util.PartialBoolean;
+
 public class EvalControl {
 
   public static final int KeepLazy = 1;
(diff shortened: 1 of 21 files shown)


UnrealEugene:
I also would like to implement similar logic in `Tool#getActions` method but since it saves actions into `actions` field which is shared between workers it could cause contention and thread safety issues inside `LazyValue`.

lemmy:
Beware of thread safety unless values are deeply normalized!

Calvin-L:
@lemmy  Is thread safety a risk here? My intuition says this code should be safe; we have (1) `WorkerValue` and (2) `fingerPrint`-implies-`deepNormalize`, which together should protect this kind of change from thread safety issues. But, I could easily be missing something.

I would want to see a few additional tests with this change. There are lots of things you can do with `LET`, and I think we should have a few small tests to cover them. I don't know if the existing tests cover these cases.

Try to force a LazyValue into an initial state:
`Init == LET x = 1 IN var = x` -- equivalent to `Init == var = 1`

Try to force a LazyValue into a successor state:
`Next == LET x = 1 IN var' = x` -- equivalent to `Next == var' = 1`

Try to force a LazyValue into a nested position:
`Next == LET x = 1 IN var' = [field |-> <<10, x>>]` -- equivalent to `Next == var' = [field |-> <<10, 1>>]`

Use a let-bound expression to compute next-states:
`Next == LET x = (var' = 1) IN x` -- equivalent to `Next == var' = 1`

Have a let-bound expression cross an `EvalControl` boundary:
`LET x = var' IN ENABLED (x = 0)` -- equivalent to `ENABLED (var' = 0)`

lemmy:
@Calvin-L My comment was more of a general observation.  I haven't considered the implications of this change at all.

UnrealEugene:
Added simple tests. Also I'm not planning to implement similar caching logic in `Tool#getActions` because actions and their contexts are shared between workers and it would need some additional refactoring to implement (e.g. make `Context` field of `Action` class to be thread local).

UnrealEugene:
Actually now that I'm thinking more about it, changes in this PR are independent of LazyValue refactor and could be merged into master branch instead. But without LazyValue refactor it may be risky to merge.

lemmy:
Below are the [performance results](https://github.com/tlaplus/tlaplus/suites/14653144832/artifacts/831941361) how this PR (commit db78b5c) compares to master (45700fd):

|   |RevTag  |Spec            | Workers| change|
|:--|:-------|:---------------|-------:|------:|
|2  |db78b5c |Bloemen         |       2|  -1.83|
|4  |db78b5c |Bloemen         |       4|  -2.37|
|6  |db78b5c |Bloemen         |       8|  -1.97|
|7  |db78b5c |Bookkeeper      |       2|  -5.43|
|10 |db78b5c |Bookkeeper      |       4|  -4.80|
|12 |db78b5c |Bookkeeper      |       8|  -1.73|
|13 |db78b5c |EWD840          |       2|  -1.63|
|15 |db78b5c |Ghostferry      |       2|   1.08|
|18 |db78b5c |Ghostferry      |       4|   0.22|
|20 |db78b5c |Ghostferry      |       8|  -6.89|
|22 |db78b5c |Grid5k          |       2|   0.22|
|24 |db78b5c |Grid5k          |       4|   2.58|
|25 |db78b5c |MongoRepl       |       2| -17.35|
|28 |db78b5c |MongoRepl       |       4| -19.69|
|29 |db78b5c |MongoRepl       |       8| -19.71|
|31 |db78b5c |PaxosCommit     |       2|  -1.78|
|34 |db78b5c |PaxosCommit     |       4|  -2.05|
|36 |db78b5c |PaxosCommit     |       8|  -2.58|
|37 |db78b5c |PaxosMadeSimple |       2| -13.72|
|40 |db78b5c |PaxosMadeSimple |       4| -15.17|
|42 |db78b5c |PaxosMadeSimple |       8| -16.18|
|44 |db78b5c |SwarmKit        |       2|  40.49|
|46 |db78b5c |SwarmKit        |       4|  40.03|
|48 |db78b5c |SwarmKit        |       8|  43.86|

```R
## Read input csv file.
library(here)
data <- read.csv(header=TRUE, sep = "#", file = here("out_run-stats.csv"))

## Convert epoch to date.
library(anytime)
data$Date <- anytime(data$Timestamp)
  
## Aggregate multiple runs.
data <- aggregate(cbind(Generated,Duration) ~ Spec + RevTag + Workers, data = data, FUN = mean, na.rm = TRUE)

## Calculate Throughput on aggregated data.
data$Throughput <- data$Generated / data$Duration

##################

# Get the baseline throughput values
baseline <- data[data$RevTag == "45700fd", c("Spec", "Workers", "Throughput")]
names(baseline)[3] <- "baseline_throughput"

# Merge baseline_throughput to the original data frame
data <- merge(data, baseline, by = c("Spec", "Workers"))

# Calculate the change in throughput percentage
data$change <- (data$Throughput - data$baseline_throughput) / data$baseline_throughput * 100

library(knitr)
kable(format = "markdown", data[data$RevTag != "45700fd", ][, c(3,1,2,8)], digits=2)
```

lemmy:
@UnrealEugene Will this PR changes the behavior of expressions such as the following?

```tla
Foo ==
    LET rnd == RandomElement(1..2)
    IN A(rnd) \/ B(rnd)
```

lemmy:
Related: https://github.com/tlaplus/tlaplus/blob/0e41129ebaf346dd6b2ee46a9dc977aaf954a2d0/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/TLCExt.tla#L146-L154
https://github.com/tlaplus/tlaplus/blob/0e41129ebaf346dd6b2ee46a9dc977aaf954a2d0/tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java#L325-L347

UnrealEugene:
> @UnrealEugene Will this PR change the behavior of expressions such as the following?
> 
> ```tla
> EXTENDS TLC
> 
> Foo ==
>     LET rnd == RandomElement(1..2)
>     IN A(rnd) <=> A(rnd)
> ```

Yeah, for example in the following spec:
```tla
---- MODULE Test ----
EXTENDS Integers, TLC
VARIABLE x

A(y) == y = 1

Next ==
    /\ LET rnd == RandomElement(1..2)
       IN /\ A(rnd) <=> A(rnd)
          /\ x' = (x + 1) % 20

Spec == x = 0 /\ [][Next]_x
====
```
With changes in this PR the line `A(rnd) <=> A(rnd)` is always true, but without changes it can be sometimes false.

Issue #828, opened 2023-08-10
Is there a way to get access to state information during model checking?


craft095:
I wonder if it is possible by using Java overrides (or by some other means) to get following information "online" while TLC is working:
1. State information
2. Transition information - a pair of states and some action id

I am looking for a way to build state graph projection "on the fly", because dot files quickly become impractical for larger models. Projected graphs can be much smaller and problem-oriented so that they are easier to analyze.

Thank you!

UnrealEugene:
It's kinda possible using Java module overrides with `@EvaluatingValue` annotation since it provides you with current and next state (but not sure if it's valid to work with next state when it's not fingerprinted) although you would need to put new overridden operator at the end of every action and sacrifice some readability of your spec. 

The other way is to fork the repository and write your own `IStateWriter` implementation using `DotStateWriter` class as a reference. This interface contains `writeState(...)` methods which are called when new state/transition is discovered by TLC. 

I tackled this problem myself some time ago the second way but it would be certainly be nice to have a way to export state graph without directly modifying TLC source code. Maybe add Java property for TLC to use custom `IStateWriter` implementation and then provide it in the classpath?

lemmy:
> The other way is to fork the repository and write your own `IStateWriter` implementation using `DotStateWriter` class as a reference. This interface contains `writeState(...)` methods which are called when new state/transition is discovered by TLC.

We can add an (undocumented) system property that causes TLC to load a custom `IStateWriter` instance, but the IStateWriter` interface was never intended to be public and, thus, has [quirks](https://github.com/tlaplus/tlaplus/blob/04c46d87c1d58c87ed779c088461bbe5ec837a4c/tlatools/org.lamport.tlatools/src/tlc2/util/DotStateWriter.java#L220-L223).

@craft095 I'm aware that the dot files can become huge, but have you considered tools like [Gephi](https://gephi.org) to do projections offline?

UnrealEugene:
> We can add an (undocumented) system property that causes TLC to load a custom `IStateWriter` instance, but the `IStateWriter` interface was never intended to be public and, thus, has [quirks](https://github.com/tlaplus/tlaplus/blob/04c46d87c1d58c87ed779c088461bbe5ec837a4c/tlatools/org.lamport.tlatools/src/tlc2/util/DotStateWriter.java#L220-L223).

I can provide a PR for this.

craft095:
Awesome! I'll try this. Thanks, @lemmy , @UnrealEugene !

[...]

PR #829, opened 2023-08-11
Support custom IStateWriter implementation to be used during model checking

UnrealEugene:
Added JVM system property `-Dtlc2.TLC.customStateWriter=...` which when provided causes TLC to use specified class as a `IStateWriter` instance type.

A follow-up to https://github.com/tlaplus/tlaplus/issues/828#issuecomment-1673748702


lemmy:
Closed with https://github.com/tlaplus/tlaplus/commit/0e41129ebaf346dd6b2ee46a9dc977aaf954a2d0

Issue #771, opened 2022-10-20
Release date not shown in nightly release


heidihoward:
Would it be possible to get the release date in the first line of tlc output for the nighly release?

Currently, I see the following when I run tlc from the nighly:
```
TLC2 Version 2.18 of Day Month 20?? (rev: 27eace5)
```

heidihoward:
That's handy and sufficient for my needs, thank you!

ahelwer:
This is done by PR #832, but unfortunately I would recommend both that PR and this issue be closed because adding timestamps to builds is a classic way to stop builds from being reproducible. Not that we have reproducible builds now, but it's a good thing to keep an eye on.

heidihoward:
Personally, I've vote for usability over reproducible builds in this particular setting but I don't feel strongly (git commit is sufficient for me). It might look better if `Day Month 20??` was removed entirely from the cli output for the nightly build. Feel free to close

lemmy:
@ahelwer There is no consensus that we need/want reproducible builds.

ahelwer:
I recall you were concerned about build reproducibility in relation to OS-varying line endings of JavaCC-generated code.

As an idea here, we could replace the `TLC2 Version 2.18 of Day Month 20?? (rev: 27eace5)` (where does 2.18 come from?) text with just:
```
The TLC finite model checker for TLA⁺ (rev: 27eace5)
https://github.com/tlaplus/tlaplus/commit/27eace5
```

ahelwer:
Actually I thought of something even better - commits contain timestamps, so just use the timestamp from the commit instead of the timestamp from the system clock at build time! Probably more semantically useful than the build time even. As long as we normalize the commit timestamp to UTC the build should be reproducible.

ahelwer:
So I've taken a liking to the idea of adding the commit time to the build & have been looking at this, and unfortunately [the git ant task](https://github.com/alx3apps/jgit-buildnumber) we use is unmaintained & hasn't seen an update in a decade, and also does not expose the git commit time. There are a few possible things I could do:

1. Fork the ant task we're using and modify it to emit the latest commit time, then update the `tlatools/org.lamport.tlatools/lib/jgit-buildnumber-ant-task-1.2.10.jar` file in our repo to use a jar built from my fork
[...]

PR #832, opened 2023-09-25
Include build time in version string, rather than placeholder

eddyashton:
The string `Version 2.18 of Day Month 20??` appears in all TLC outputs. Although it is followed by a git revision, it's not immediately clear how old the related build is.

This PR replaces that placeholder `Day Month 20??` with the actual build date, ISO ordered for international clarity, so we now get a more friendly line:

```
TLC2 Version 2.18 of 2023-09-25 (rev: 8290358)
```

I note that `versionOfTLC` is used in many places, not just `printWelcome()`. I assume they'll also appreciate the change, though I don't know how to confirm that.


lemmy:
In a release, the date will be formatted as `TLC2 Version 2.18 of 20 March 2023`, and the absence of a timestamp used to mark tla2tools.jar as a nightly build.  How about formatting the date as seconds since epoch in nightly builds?

By the way, `TLCGet("revision").date` equals TLC's build date.

eddyashton:
> In a release, the date will be formatted as `TLC2 Version 2.18 of 20 March 2023`, and the absence of a timestamp used to mark tla2tools.jar as a nightly build. How about formatting the date as seconds since epoch in nightly builds => easily distinguishable yet solves the problem?
> 
> By the way, `TLCGet("revision").date` equals TLC's build date.

How is that done - is the string manually edited for a real release?

I think adding "2.18 (nightly)" or "2.18 (candidate)" would make this distinction clearer, but I'm also happy to match the existing date format if there's precedent in prior releases (ie - potentially other tools parsing this).

lemmy:
Exactly, the string is manually edited. :-)

lemmy:
It occurred to me that `TLC2 Version 2.18 of Day Month 20??` is also hardcoded and appears in the Toolbox.  There are also version strings and dates for SANY, PlusCal, ...

lemmy:
It occurred to me that `TLC2 Version 2.18 of Day Month 20??` is also hardcoded and appears in the Toolbox.  There are also version strings and dates for SANY, PlusCal, ...

ahelwer:
I don't think this is a good idea for build reproducibility reasons.

lemmy:
> I don't think this is a good idea for build reproducibility reasons. Timestamps are the classic example of something you should never include in the build if you want it to be reproducible.

See https://github.com/tlaplus/tlaplus/issues/771#issuecomment-2171863334

lemmy:
Closing in favor of https://github.com/tlaplus/tlaplus/pull/1328

PR #836, opened 2023-10-23
Remove uses and overrides of `Object.clone()`

Calvin-L:
This commit makes one iota of progress toward the suggestions in #756 by replacing uses of `.clone()` with copy constructors, adding new copy constructors as necessary, and removing `implements Cloneable` from all types that have it.  In my mind, these changes are prerequisites for removing the tools' custom data structures (e.g. `util.Set` and `tlc2.util.Vect`).

The `clone()` method has widely been considered bad practice for many years [1].  Problems include:
 - Lack of type safety
 - A vague contract
 - "Magical" capabilities that can't be replicated in pure Java

Copy constructors are considered a better pattern.

[1]: https://www.artima.com/articles/josh-bloch-on-design

### DIFF
--- a/tlatools/org.lamport.tlatools/src/util/Set.java
+++ b/tlatools/org.lamport.tlatools/src/util/Set.java
@@ -9,24 +9,27 @@
 package util;
 
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.Enumeration;
 import java.util.NoSuchElementException;
 
 class SetEntry {
-  /*
-   * Set collision list.
-   */
-  int hash;
-  Object key;
-  SetEntry next;
+    /*
+     * Set collision list.
+     */
+    int hash;
+    Object key;
+    SetEntry next;
 
-  protected Object clone() {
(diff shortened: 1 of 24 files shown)


tlatools/org.lamport.tlatools/src/util/Set.java
```diff
@@ -375,18 +376,21 @@ public void intersection (Set x) {
...
-   * the smaller (set size) of x, y because that's what intersection
-   * iterates on.
+   * Create a new set which is the intersection of x and y.  We make
+   * a shallow copy of the smaller (set size) of x, y because that's
+   * what intersection iterates on.
+   *
+   * TODO 2023/10/23: the comment above is wrong; the branches are identical.
+   *     We are missing out on whatever performance benefits could be gained.
    */
   public static Set intersection(Set x, Set y) {
```
lemmy, line 386: All the static variants of intersection, minus, complement, ... appear to be dead code anyway.

lemmy:
LGTM

PR #842, opened 2023-11-09
Fix `checkResolveStandardModule` test

Calvin-L:
There were two problems here:

 1. A copy-paste bug: I generalized this logic from an older test but did not update the arguments to `resolve`.
 2. The final check does not pass when the tests run in a JAR file (as they do for the `test-dist` Ant target).


Calvin-L:
This is a test-only change and the tests are passing, so I am going to merge this right away.

Issue #855, opened 2024-01-03
tla2tex with `-out` option fails if the output file does not exist already


nano-o:
If `out.tex` does not exists already, running `java -cp tla2tools.jar tla2tex.TeX -out out.tex in.tex` fails with:
```
TLATeX unrecoverable error:

 -- Error while renaming files.


TLATeX unrecoverable error:

 -- I/O error: TLATeX unrecoverable error: -- Error while renaming files..

Exception in thread "main" tla2tex.TLA2TexException: TLATeX unrecoverable error: -- I/O error: TLATeX unrecoverable 
error: -- Error while renaming files..
        at tla2tex.Debug.ReportError(Debug.java:26)
        at tla2tex.TeX.main(TeX.java:411)
```

This seems to be because line 406 of `TeX.java` tries to rename the output file, which fails if the output file does not exist already. A workaround is to create an empty output file first.

lemmy:
@nano-o I believe this is fixed.

PR #857, opened 2024-01-03
tla2tex with `-out` option fails if the output file does not exist already.

lemmy:
tla2tex with `-out` option fails if the output file does not exist already.

Fixes Github issue #855
https://github.com/tlaplus/tlaplus/issues/855

[Bug][TeX]


lemmy:
@nano-o Please review.

Issue #872, opened 2024-02-05
Normalize line endings as LF or CRLF and encode choice in .gitattributes file


ahelwer:
Currently files in the repo are a mix of LF or CRLF. Since contributions can be written on any of the major OS platforms, a .gitattributes file should be added enforcing that files are either CRLF (Windows-style) or LF (Linux-style). This is especially important because code generation is used in SANY, and depending on the platform it will generate files with either CRLF or LF endings. This can lead to spurious diffs if those generated files replace existing files & are then committed.

https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/

lemmy:
Based on the [execution statistics](https://metabase.tlapl.us/public/dashboard/cf7e1a79-19b6-4be1-88bf-0a3fd5aa0dec?jvm_vendor=&os_name=&ide_launched=&fp_set_type=&category=tlc), unix-based platforms are way more popular.

kape1395:
I like this idea. And I'm for the Linux-style EOLs.

lemmy:
Can we incrementally change all existing CRLF line endings to LF using a `.gitattributes` file, so that only the portions of a file modified by a commit have their line endings altered? Converting all CRLFs to LFs in a single operation would make comparing file versions to identify regressions very difficult.

ahelwer:
I don't know, I think it would be better to get it out of the way in one large commit just called "Normalize line endings" so people know to ignore that one. If it's mixed in with another commit changing the file it can make it harder to see the actual change.

ahelwer:
Documenting this to add to the confusion, some files have both LF and CRLF line endings within the same file. For example, this section:

https://github.com/tlaplus/tlaplus/blob/de0c1811f291fcb68e81873af1fba7eb3fa9814a/tlatools/org.lamport.tlatools/src/tla2sany/parser/TLAplusParserTokenManager.java#L4267:L4278

has LF line endings while (most of) the rest of the file is CRLF. You can see this by opening the file with `vim -b filename`; lines ending with CRLF will have a `^M` character displayed.

This is unfortunate because I'm trying to see what modifications have been made to the javacc-generated parser code but these line endings are confounding the diff attempt.

Issue #246, opened 2019-03-05
Sync input files of the Configuration parser with the source code


Henne:
TLA2SANY uses two parsers which were generated with JavaCC.
The input files are tla+.jj for the grammar of TLA+ and config.jj for the grammar of the built-in operators which I call Configuration parser.
The generated files from tla+.jj reside in tla2sany/parser/ and the files from config.jj in tla2sany/configuration/.

These generated files were modified manually, but the changes were not put back into the jj input files.
So the input and generated files are out of sync, for both TLA+ and the Configuration parser.
That makes it:
- hard to modify these grammars and 
- hard to keep up-to-date with newer JavaCC Versions, which may fix issues in the generated code.

The generated files from config.jj, which are directly used by tla2sany and therefore by tlc2, have been generated with JavaCC 0.7pre6, which is currently not available anymore.
See: [JavaCC Downloads](https://javacc.org/downloads/)
Warning: These files of the old releases aren't classes. They are html files and of no use.

I would like to provide a patchset I have written, which 
- updates the generated files into the form they were generated by JavaCC 4.0 without breaking the build and
- updates the config.jj file, so that the output of it can be just-copied into tla2sany/configuration and works without manual modifications.

One benefit would be that the TLA+ parser, which has already been generated with JavaCC 4.0,
together with the config.jj can be updated with lesser effort to newer versions of JavaCC.
The last release of JavaCC was version 7.0.5.
Version 4.0 is also very old, but luckily available in some current Linux distributions.

Another benefit can be, that the generated code could be shared between both parsers,
since 4 of the 7 generated files are identical, which would remove duplications and code-smell.

There is one design decision, which I want to discuss that with you:

JavaCC supports generating static parsers, which do not need to be instantiated, and non-static parsers, which need to be instantiated.

At the moment the Configuration parser is static and the TLA+ parser is non-static.

The Configuration parser has one noticeable method called Configuration.load(),
with is called in tla2sany/drivers/SANY.java without instantiating a Configuration object.
But this load() method, defined in config.jj and Configuration.java,
instantiates a configuration object which then does the parsing.

So one way would be to make the Configuration parser non-static, since
- it has already been used that way all the last 10 years,
- there is no need to think about a static and a non-static parser in tlatools and
- the ReInit() method added manually by DRJ can be removed and the one and only method
  used by tla2sany is load() which does all the job.

For keeping it static I'm lacking arguments.

lemmy:
[...]

PR #873, opened 2024-02-08
Make tla+.jj javacc-generated files all end in CRLF

ahelwer:
Almost all the tla+.jj javacc-generated files are entirely CRLF except for the changed sections of these files. I just ran two files (out of seven generated) through the `unix2dos` command line utility, so the only things changed are the line endings from `\n` to `\r\n` (wish github displayed this better). This small change is a prelude to identifying & extracting the hand-modified parts of these files so we can go back to the tla+.jj file as the source of truth (ref #246).

In future work/policy involving #872 we might want to actually enforce these files as going to LF instead of CRLF, but this is a smaller change until we decide on that so it seems best to do this for now.


lemmy (approved):
LGTM

Issue #851, opened 2023-12-01
IDE setup instructions need updating


Calvin-L:
The current [IDE setup instructions](https://github.com/tlaplus/tlaplus/blob/193da949dc8eb783a2683f01f3a54da93fd85a39/general/ide/README.md) are not working for me. They may be out of date. It is also possible that my machine has some bizarre (mis)configuration that is preventing it from succeeding.

My setup:
 - MacOS 12.6.8, M1 (aarch64)
 - Eclipse Installer (Oomph) 1.31.0 Build 329, downloaded from https://www.eclipse.org/downloads/packages/installer

Notes about the current instructions:

The Oomph link https://wiki.eclipse.org/Eclipse_Installer is a wiki page with a big warning about the wiki being deprecated. I believe the new link is https://www.eclipse.org/downloads/packages/installer.

Eclipse platform 2020-12 is no longer available. The earliest version available is 2021-12:

![Screen Shot 2023-12-01 at 2 28 41 PM](https://github.com/tlaplus/tlaplus/assets/1393639/99c31e30-cf36-4c29-9057-f496b84af2e7)

Although, the installer does not seem to respect that selection. It seems to always install Eclipse 2023-09 (4.29). I selected 2021-12, but in the installer logs (`<installdir>/Eclipse.app/Contents/Eclipse/configuration/org.eclipse.oomph.setup/setup.log`):
```
[2023-12-01 13:32:12] Product org.eclipse.products.org.eclipse.platform.ide.2021-12
[...]
[2023-12-01 13:32:12] Performing P2 Director (Eclipse Platform (2021-12) + JRE 11.0.21)
[...]
[2023-12-01 13:33:21] Installing org.eclipse.platform [4.29.0.v20230903-1000]
```

And sure enough, the newest version is what gets installed:
![Screen Shot 2023-12-01 at 2 33 01 PM](https://github.com/tlaplus/tlaplus/assets/1393639/b3bedc4a-1a21-4f96-8d2c-050e5f3b931a)

That version of Eclipse seems to be incompatible with the Toolbox project. After install, during Eclipse's project setup, I get this message:
![Screen Shot 2023-12-01 at 2 14 32 PM](https://github.com/tlaplus/tlaplus/assets/1393639/db3ced8f-a73f-40a6-bc1a-e408fdecf06c)

And forevermore the installed IDE seems completely unusable:
[...]

PR #876, opened 2024-02-10
Run JavaCC at build to generate TLA+ parser files

ahelwer:
This PR does three things:

- Vendors JavaCC 4.0 to `lib/javacc-4.0.jar` for easy access during grammar development
- Adds a target to the tlatools ant build to generate source files from `tla+.jj`, normalize their line endings, then overwrite the current source files with the newly-generated ones
- Adds a check to the PR CI to ensure `tla+.jj` and the code files are in sync (borrowing method from tree-sitter-tlaplus, which also has generated parser artifacts in source control)

The generated source files are kept in the tree instead of being deleted.


lemmy:
Is it correct that this breaks our [IDE setup story](https://github.com/tlaplus/tlaplus/tree/master/general/ide)?

lemmy:
One more thing to check is if this causes (additional?) differences between the `tla2tools.jar` when built on Windows, macOS, and Linux, due to the different line endings in the generated files.

ahelwer:
I found an ant build target that normalizes line endings, so that should take care of the build reproducibility issue. I will look into the IDE setup.

ahelwer:
Well I've spent a few hours trying to test that the IDE setup still works and I can't get it to work at baseline. Tried on macOS and ran into #851. Tried on Windows and it gives me a bunch of errors about not being able to find jdk.jfr. So I don't really know what to do here.

lemmy:
I am fine with removing the instructions for setting up the IDE and allowing people to figure out how to configure their development environment on their own.  Even better, let's get rid of the Toolbox.

ahelwer:
Moving this back to draft because in the near future it seems useful to actually see what changes to the output code are made when modifying the grammar file, in a PR.

ahelwer:
@lemmy this PR has made a comeback. I believe it is much more useful and less invasive this time. JavaCC is automatically run to regenerate the grammar files before tla2tools.jar is built, the generated files aren't deleted, and a check is added to the CI to ensure they stay in sync.

ahelwer:
@lemmy @Calvin-L these are some quite simple changes it would be nice to get in!

tlatools/org.lamport.tlatools/customBuild.xml
```diff
@@ -101,7 +99,24 @@
 		<antcall target="dist" inheritall="true" inheritrefs="true" />
 		<antcall target="test-dist-long" inheritall="true" inheritrefs="true" />
 	</target>
-	
+
+	<!-- Runs JavaCC to generate the TLA+ parser files. -->
+	<target name="generate">
+		<tempfile property="javacc-tmpdir" destdir="${java.io.tmpdir}" prefix="javacc_" />
+		<mkdir dir="${javacc-tmpdir}" />
+		<java classpath="lib/javacc-4.0.jar" classname="javacc" fork="true" failonerror="true">
```
lemmy, line 107: Have you considered using https://ant.apache.org/manual/Tasks/javacc.html

ahelwer: I think that's only for later javacc versions

ahelwer: but I will test it!

ahelwer: Oh nevermind it wouldn't really work with how we have set up the vendored libs directory.

ahelwer:
Yeah those warnings have been there forever. Maybe I'll get around to figuring out what they mean and fixing them eventually.

PR #878, opened 2024-02-13
Ensure LF line endings for JavaCC-generated .cfg parser files

ahelwer:
Converting both these files to `LF` using the `dos2unix` command. In these ones `CRLF` line endings were a significant minority, at 214 / 2722 lines. So this makes the smallest diff, as a prelude to figuring out the differences between freshly-generated JavaCC code and the current files.

Ref line endings issue: #872 
In furtherance of the goal of #246
Similar changes for files generated from `tla+.jj`: #873


ahelwer:
@lemmy do you think the hanging test is caused by recent TE changes or just some transient CI machine issue?

lemmy:
Let's find out - triggered a rerun.

lemmy:
Merging because the files are JavaCC generated.  In general, we don't reformat source files to not artificially inflate source code diffs.

PR #879, opened 2024-02-15
Fix org.lamport.tlatools build errors in Eclipse

ahelwer:
When opening `tlatools/org.lamport.tlatools` in eclipse, users are presented with around 100 build errors. There are four causes:
- Unable to resolve modules from JUnit; fixed by adding `lib/junit-4.12.jar` and `lib/hamcrest-core-1.3.jar` to `.classpath` file
- Unable to resolve modules from EasyMock; fixed by adding `lib/easymock-3.3.1.jar` to `.classpath` file
- Unable to resolve modules from AspectJ; fixed by adding `lib/aspectjtools-1.9.2.jar` to `.classpath` file
- "`@Ignore` annotation cannot be applied to a class" error in `test/tlc2/tool/suite/Test61.java`; this is actually because the file `lib/jpf.jar` contains an old `org.junit.Ignore` class (probably accidentally) which eclipse was finding before the one in `lib/junit-4.12.jar`; moving `lib/junit-4.12.jar` above `lib/jpf.jar` in the `.classpath` file fixed this

### DIFF
--- a/tlatools/org.lamport.tlatools/.classpath
+++ b/tlatools/org.lamport.tlatools/.classpath
@@ -26,11 +26,15 @@
 	<classpathentry kind="lib" path="lib/jline/jline-builtins-3.25.0.jar"/>
 	<classpathentry kind="lib" path="lib/jline/jline-reader-3.25.0.jar"/>
 	<classpathentry kind="lib" path="lib/jline/jline-terminal-3.25.0.jar"/>
-	<classpathentry kind="lib" path="lib/jpf.jar"/>
 	<classpathentry kind="lib" path="lib/jmh/jmh-core-1.21.jar"/>
 	<classpathentry kind="lib" path="lib/jmh/commons-math3-3.2.jar"/>
 	<classpathentry kind="lib" path="lib/gson/gson-2.8.6.jar"/>
 	<classpathentry kind="lib" path="lib/lsp/org.eclipse.lsp4j.debug_0.21.1.v20230829-0012.jar"/>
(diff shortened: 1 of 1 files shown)


tlatools/org.lamport.tlatools/.classpath
```diff
@@ -32,5 +32,8 @@
 	<classpathentry kind="lib" path="lib/gson/gson-2.8.6.jar"/>
 	<classpathentry kind="lib" path="lib/lsp/org.eclipse.lsp4j.debug_0.21.1.v20230829-0012.jar"/>
 	<classpathentry kind="lib" path="lib/lsp/org.eclipse.lsp4j.jsonrpc_0.21.1.v20230829-0012.jar"/>
+	<classpathentry kind="lib" path="lib/junit-4.12.jar"/>
+	<classpathentry kind="lib" path="lib/easymock-3.3.1.jar"/>
```
lemmy, line 36: All three jars are part of Eclipse's target platform.  Does this cause conflicts, for example, when launching the Toolbox from Eclipse?

ahelwer: I don't know, I can't get the toolbox to launch.

lemmy:
I'd suggest to focus on setting up a tlatools/ (not the Toolbox) dev environment for VSCode ([CommunityModules](https://github.com/tlaplus/CommunityModules/tree/master/.vscode) already have one).  If such an environment is available and attracts contributions, it would be a convincing argument in favor of ditching the Eclipse-based dev environment (and the Toolbox).

Issue #885, opened 2024-03-06
SANY incorrectly allows subexpression prefixes for labels with parameters


ahelwer:
Here is the input which SANY should fail on but instead accepts:
```tla
---- MODULE Test ----
op == \A x \in {} : A!B!lbl(x) :: x
====
```
The `A!B!` should be disallowed at the parser level. The semantic level doesn't catch this because it assumes nobody would ever add a subexpression prefix there. Here is the parser trace:
```
****** SANY2 Version 2.2 created 08 July 2020

Parsing file /home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/Test.tla
Beginning Module definition
Beginning Begin module
Ending Begin module
Beginning Extends
Ending Extends
Beginning Module body
Beginning Definition
Beginning Identifier LHS
Ending Identifier LHS
Beginning Expression
Beginning Quantified form
Beginning Quant Bound
Beginning Expression
Beginning ExtendableExpr
Beginning Some { } form
Ending Some { } form
Ending ExtendableExpr
Ending Expression
Ending Quant Bound
Beginning Expression
Beginning ExtendableExpr
Beginning Primitive expression
Beginning Bang Extension
Ending Bang Extension
Beginning Bang Extension
Beginning Optional Arguments
Beginning Expression
Beginning ExtendableExpr
Beginning Primitive expression
Ending Primitive expression
Ending ExtendableExpr
Ending Expression
Ending Optional Arguments
Ending Bang Extension
Ending Primitive expression
Beginning Expression
Beginning ExtendableExpr
Beginning Primitive expression
Ending Primitive expression
Ending ExtendableExpr
Ending Expression
Ending ExtendableExpr
Ending Expression
Ending Quantified form
Ending Expression
Ending Definition
Ending Module body
Ending Module definition
Semantic processing of module Test
```
SANY correctly rejects this behavior when the label does not take parameters; this fails:
```tla
---- MODULE Test ----
op(x) == A!B!lbl :: x
====
```
Bug discovered during the course of working on #882.

Issue #884, opened 2024-02-23
SANY fails to parse negative and !! operators in nonfix form


ahelwer:
This spec will fail to parse:
```tla
---- MODULE Test ----
EXTENDS Naturals
op == -(2, 1)
====
```
Running SANY with `-DTLA-StackTrace=on` produces this output:
```
****** SANY2 Version 2.2 created 08 July 2020

Parsing file /home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/Test.tla
Beginning Module definition
Beginning Begin module
Ending Begin module
Beginning Extends
Ending Extends
Beginning Module body
Beginning Definition
Beginning Identifier LHS
Ending Identifier LHS
Beginning Expression
Beginning Infix Op
Ending Infix Op
Beginning ExtendableExpr
Beginning Expression
Beginning ExtendableExpr
Beginning Primitive expression
Ending Primitive expression
Ending ExtendableExpr
Ending Expression
***Parse Error***
Encountered "," at line 3, column 10 and token "2" 

Residual stack trace follows:
ExtendableExpr starting at line 3, column 8.
Expression starting at line 3, column 7.
Definition starting at line 3, column 1.
Module body starting at line 3, column 1.
Module definition starting at line 1, column 1.


Fatal errors while parsing TLA+ spec in file Test.tla

tla2sany.semantic.AbortException
*** Abort messages: 1

In module Test

Could not parse module Test from file Test.tla
```

Caught during work on #882, since the tree-sitter grammar correctly parses this and the case is in its corpus (tree-sitter also flagged the ambiguity during static analysis, because it's a LR(1) parser while JavaCC uses recursive descent).

This is a fairly classic parser lookahead problem; SANY commits to interpreting `-(` as a negative operator applied to some forthcoming expression in parentheses, when in fact it is the minus operator in nonfix form.

This issue is somewhat similar to #637, where SANY cannot parse `/\` or `\/` in nonfix form, but we decided to resolve that one by defining SANY's current behavior to be correct; see https://github.com/tlaplus/rfcs/issues/4

In contrast I believe this is an actual bug with SANY that should be fixed.

ahelwer:
This will also fail to parse, so SANY can't handle the `!!` operator in nonfix form because it conflicts with the `!` subexpression separator:
```tla
---- MODULE Test ----
---- MODULE Inner ----
x !! y == TRUE
====
M == INSTANCE Inner
op == M!!!(1, 2)
====
```

PR #886, opened 2024-03-07
Adapt tree-sitter corpus tests for use with SANY

### LINKED ISSUE #882: Work proposal: add corpus of tests for SANY at syntax parsing level
I would like to run this proposal past the maintainers before spending much time on it.

Tree-sitter grammars have a very nice method of writing parser tests, using a collection of .txt files with a very basic DSL; for example:
<details>
  <summary>Sample from tree-sitter-tlaplus test corpus</summary>

```
=============|||
Multiline Conjlist
=============|||

---- MODULE Test ----
op ==
  /\ 
   1
  /\
   2
====

-------------|||

(source_file (module (header_line) (identifier) (header_line)
  (operator_definition (identifier) (def_eq)
    (conj_list
      (conj_item (bullet_conj) (nat_number))
      (conj_item (bullet_conj) (nat_number))
    )
  )
(double_line)))

=============|||
Nested Conjlist
=============|||

---- MODULE Test ----
op ==
 /\ /\ 1
    /\ 2
 /\ 3
====

-------------|||

(source_file (module (header_line) (identifier) (header_line)
  (operator_definition (identifier) (def_eq)
    (conj_list
      (conj_item
        (bullet_conj)
        (conj_list
          (conj_item (bullet_conj) (nat_number))
          (conj_item (bullet_conj) (nat_number))
        )
      )
      (conj_item (bullet_conj) (nat_number))
    )
  )
(double_line)))
```
</details>

Each test involves a title, the source code input, and the expected parse tree output in an S-expression-like DSL.

I would like to take the fairly large corpus of tests already existing in the tree-sitter-tlaplus repo and write a simple translator so they can be run against SANY at the syntax parsing level. This would involve no changes to any SANY code itself. I believe there are the following benefits:

- SANY doesn't currently have any unit tests similar to this, so it seems an unqualified good thing to add them
- sharing test inputs across the various TLA+ parsers (perhaps TLAPM parser could also be added in future) increases incentive to expand the corpus and helps ensure conformance between the parsers
- Using a lightweight DSL such as this instead of encoding the expected parse tree as Java objects greatly reduces the difficulty of adding additional tests

You can see the current tree-sitter-tlaplus test corpus [here](https://github.com/tlaplus-community/tree-sitter-tlaplus/tree/main/test/corpus). There are some tests including unicode symbols which will be omitted.

If this work is performed, will it likely be merged? Follow-up decision if so: it would be nice to upgrade to JUnit 5 (currently we use version 4) so these corpus tests can be emitted at runtime as [dynamic tests](https://junit.org/junit5/docs/5.0.2/api/org/junit/jupiter/api/DynamicTest.html) instead of running all of them inside a single unit test, although this is not strictly necessary.

ahelwer:
Adapts all tree-sitter-tlaplus corpus tests to work with SANY and test it at the syntax processing level. Also greatly expands the number of tests, benefiting both projects.

Turns out translating parse trees from one format to another is a lot of work! Who knew! But I did it, so hurrah for that. Take a look at the monstrous `SanyTranslator.translate` function.

The new unit test class `tla2sany.ParseCorpusTest` round-trips 182 parse inputs and executes in 0.3 seconds.

Closes #882


tlatools/org.lamport.tlatools/.classpath
```diff
@@ -26,11 +26,15 @@
...
 	<classpathentry kind="lib" path="lib/jline/jline-terminal-3.25.0.jar"/>
-	<classpathentry kind="lib" path="lib/jpf.jar"/>
 	<classpathentry kind="lib" path="lib/jmh/jmh-core-1.21.jar"/>
 	<classpathentry kind="lib" path="lib/jmh/commons-math3-3.2.jar"/>
 	<classpathentry kind="lib" path="lib/gson/gson-2.8.6.jar"/>
 	<classpathentry kind="lib" path="lib/lsp/org.eclipse.lsp4j.debug_0.21.1.v20230829-0012.jar"/>
 	<classpathentry kind="lib" path="lib/lsp/org.eclipse.lsp4j.jsonrpc_0.21.1.v20230829-0012.jar"/>
+	<classpathentry kind="lib" path="lib/junit-4.12.jar"/>
+	<classpathentry kind="lib" path="lib/jpf.jar"/>
+	<classpathentry kind="lib" path="lib/easymock-3.3.1.jar"/>
```
lemmy, line 36: Adding these libraries to .classpath will cause the to be part of the tlatools jar that is part of the Toolbox.

ahelwer: How then should the project be set up so unit tests can be run from eclipse without having to load in all the toolbox stuff?

ahelwer: Reverted these changes in this branch.

ahelwer:
@lemmy this is now ready to review. All these corpus test .txt files are duplicated from the tree-sitter-tlaplus repo, so if you want we could make a third repo called like "unified-syntax-test-corpus" or something to hold them then import it as a submodule. On the other hand it's just a bunch of text files and in-tree duplication can be considered best practice for archival purposes.

tlatools/org.lamport.tlatools/test/tla2sany/ParseCorpusTest.java
```diff
@@ -0,0 +1,93 @@
...
+	 */
+	@Test
+	public void testEntireCorpus() throws ParseException {
+		int testCount = 0;
+		for (CorpusTestFile corpusTestFile : ParseCorpusTest.corpus) {
+			System.out.println(corpusTestFile.path);
+			for (CorpusTest corpusTest : corpusTestFile.tests) {
+				// keeping this here because it is useful to debug why a specific test fails
+				if (!corpusTest.name.equals("Subexpression Tree Navigation")) {
+					//continue;
```
lemmy, line 64: ?

tlatools/org.lamport.tlatools/test/tla2sany/SanyTranslator.java
```diff
@@ -0,0 +1,1633 @@
...
+
+/**
+ * Handles the translation of raw SANY output to the test file AST DSL.
+ */
+public class SanyTranslator {	
+
+	/**
+	 * Sometimes SANY provides very little structure in its parse output,
+	 * basically just a flat list of tokens. This is most common when dealing
+	 * with constructs that SANY doesn't care much about, like proof-related
```
lemmy, line 27: How much is missing in SANY to add more structure for proof-related syntax?  There is an unfinished attempt to replace TLAPS' parser with SANY and this sounds like a potential show-stopper for this effort.

tlatools/org.lamport.tlatools/test/tla2sany/ParseCorpusTest.java
ahelwer: For future development and adding of tests. The main way I translated SANY's parse tree was running the debugger and seeing what SANY emits at a given point in the giant switch statement. Some of the rules are run through hundreds of times if you run the entire corpus, so it's useful to be able to pick out a single test to run so you get the debugger state in a specific rule for that test. That's what this code stub does, if you uncomment continue. It can be removed but it might be useful to the next person (probably me) who adds tests.

tlatools/org.lamport.tlatools/test/tla2sany/SanyTranslator.java
ahelwer: Not really that much. I would say use-or-hide is the main culprit, which you can see because I had to break that parsing logic out into its own set of functions instead of writing it in the main switch statement. It's annoying but surmountable. The parse output of SANY could also be improved.

lemmy:
> @lemmy this is now ready to review. All these corpus test .txt files are duplicated from the tree-sitter-tlaplus repo, so if you want we could make a third repo called like "unified-syntax-test-corpus" or something to hold them then import it as a submodule. On the other hand it's just a bunch of text files and in-tree duplication can be considered best practice for archival purposes.
> 

Redundancy doesn't hurt.  Redundancy doesn't hurt.  However, I'd suggest to add a big fat warning here and in the tree-sitter-tlaplus repo to always modify the files in both places.

tlatools/org.lamport.tlatools/test/tla2sany/SanyTranslator.java
lemmy: Makes sense, I suggest adding this note to the code.

tlatools/org.lamport.tlatools/test/tla2sany/ParseCorpusTest.java
ahelwer: Added this explanation as a comment in the code.

tlatools/org.lamport.tlatools/test/tla2sany/SanyTranslator.java
ahelwer: Added this note as a comment in the code here.

Calvin-L:
> (@Calvin-L are you going to review it as well?)

I skimmed over it, but I'm not planning to do a deep review. I'm happy to see this change, and the two things that matter most to me are
1. This is test-only change, so it's essentially zero-risk.
2. It has already proved its worth by discovering lots of corner cases like #884, #885, and #888. :)

Issue #835, opened 2023-10-23
Java 21 causes TLC's BufferedRandomAccessFileTest and OffHeapDiskFPSetTest to fail


lemmy:
Some methods such as `tlc2.util.BufferedRandomAccessFile#readLong` that used to throw `IOException` have started to throw `ArrayIndexOutOfBoundsExceptions`.

Calvin-L:
I'm looking at putting together a fix for this.

I'm very impressed with the quality and clarity of `BufferedRandomAccessFile`, but it seems to contain a critical incorrect assumption:
https://github.com/tlaplus/tlaplus/blob/7fcaceb9c17abd2c2b65903ad7a76b6cf17e5cca/tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java#L55

But, in reality, ["The offset may be set beyond the end of the file"](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/RandomAccessFile.html#seek(long)).

In JDK<21, [`readLong()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/RandomAccessFile.html#readLong()) and friends were implemented using repeated calls to `read()`. In JDK>=21 they are implemented using a bulk array read. That change in behavior seems to have revealed the incorrect assumption above. When seeking past the end of the file, some of the checks and math in `BufferedRandomAccessFile` don't work as intended. It ends up breaking its own internal invariants (in particular, it enters a state where `this.curr > this.hi`).

I think fixing this issue will require relaxing the invariant that `lo <= curr <= hi`:
https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java#L72

Calvin-L:
@lemmy I discovered something else interesting while testing my changes. The tests include:
```java
	@Test
	public void testReadSeek() throws IOException {
		final File tmpFile = File.createTempFile("BufferedRandomAccessFileTest_testReadSeek", ".bin");
		tmpFile.deleteOnExit();
		final java.io.RandomAccessFile raf = new BufferedRandomAccessFile(tmpFile, "rw");
		
		raf.setLength(BufferedRandomAccessFile.BuffSz + 1L);
		raf.seek(1);
		
		for (int i = 0; i < BufferedRandomAccessFile.BuffSz / 8; i++) {
			assertEquals(0L, raf.readLong());
		}
		raf.close();
[...]

PR #890, opened 2024-03-11
Flush writes to disk before asserting facts about the disk files

Calvin-L:
Discovered while working on #835.

I am not in love with this fix, and I am open to other ideas. But, it seemed to be the most straightforward way to fix the problem.

----

This commit fixes a latent issue in the test suite.

`CommonTestCase.assertNodeAndPtrSizes` makes assertions about the on-disk state of some files that are internal to `AbstractDiskGraph`. These checks are useful, but today they only pass because of two brittle implementation details:

 - `BufferedRandomAccessFile` flushes writes to disk when seeking backward in the file.
 - Backward seeks are very common.

The checks will break if we optimize `BufferedRandomAccessFile` to flush less often or if we change how `AbstractDiskGraph` reads/writes its internal files.

To make the tests less brittle, this commit introduces a new method that instructs the liveness checker to flush writes to disk immediately.  The new method is only called by the tests, and it is called before making assertions about on-disk state.


lemmy (approved):

lemmy:
I'm also fine with deleting `CommonTestCase.assertNodeAndPtrSizes` instead.

Calvin-L:
I'll keep that in mind if that method ever gives us trouble again. :)

Issue #893, opened 2024-03-19
SANY handles operator precedence incorrectly with the negative prefix operator


ahelwer:
Regardless of operator precedence values, there is no parse ambiguity when an infix operator has a prefix operator on its right hand side:
```tla
---- MODULE Test ----
ASSUME A & ~B
====
```
For all prefix operators except for negative, SANY follows this rule. The precedence range of the negative prefix operator is 12-12. Every infix operator of precedence 13 or greater (about a quarter of them) will generate a parse conflict where there should not be one; for example:
```tla
---- MODULE Test ----
ASSUME A & -B
====
```
There is no parse ambiguity here and yet SANY outputs a parse error:
```
****** SANY2 Version 2.2 created 08 July 2020

Parsing file /home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/Test.tla
Beginning Module definition
Beginning Begin module
Ending Begin module
Beginning Extends
Ending Extends
Beginning Module body
Beginning Definition
Beginning Identifier LHS
Ending Identifier LHS
Beginning Expression
Beginning ExtendableExpr
Beginning Primitive expression
Ending Primitive expression
Beginning Infix Op
Ending Infix Op
Beginning Infix Op
Ending Infix Op
***Parse Error***

  Precedence conflict between ops & in block line 2, col 9 to line 2, col 9 of module Test and -..

Residual stack trace follows:
ExtendableExpr starting at line 2, column 7.
Expression starting at line 2, column 7.
Definition starting at line 2, column 1.
Module body starting at line 2, column 1.
Module definition starting at line 1, column 1.


Fatal errors while parsing TLA+ spec in file Test.tla

tla2sany.semantic.AbortException
*** Abort messages: 1

In module Test

Could not parse module Test from file Test.tla
```
I assume this is because SANY special-cases the negative prefix operator due to its alternative `-.` symbol-only presentation (for example, it shows up as "minus" in prefix op expressions in the parse tree) so the logic handling these types of expressions are different than with other prefix operators.

ahelwer:
The changes in #1129 and @lemmy's suggestion to use JUnit's Assume functionality reveal that this bug doesn't manifest in all 121 test cases where negative is the rightmost operator; instead, it only manifests in these twenty expressions:
1. `A & - B`
2. `A && - B`
3. `A \odot - B`
4. `A (.) - B`
5. `A \oslash - B`
6. `A (/) - B`
7. `A \otimes - B`
[...]

Issue #892, opened 2024-03-19
Incorrect operator precedences in either documentation or SANY


ahelwer:
SANY draws operator precedence and associativity values from a string DSL that it parses living [here](https://github.com/tlaplus/tlaplus/blob/069738ed892c2adcb056db536470d0b0050f3e17/tlatools/org.lamport.tlatools/src/tla2sany/configuration/ConfigConstants.java#L74). The ["Summary of TLA" PDF](https://lamport.azurewebsites.net/tla/summary-standalone.pdf) documents these precedence & associativity values in a human-readable form on page 7. There are several prefix operators where these numbers do not match:

| Operator | SANY | Docs |
|-------------|----------|--------|
| SUBSET | 10-13 | 8-8 |
| UNION | 10-13 | 8-8 |
| DOMAIN | 10-13 | 9-9 |

Either SANY is wrong or the documentation is wrong. Probably the documentation, because I assume we don't want to change these values in SANY for backward compatibility reasons.

ahelwer:
Closing this because it's an issue with the documentation; I'll email Leslie about it.

Issue #891, opened 2024-03-18
Proposed refactoring goal: remove all static global variables


ahelwer:
The tools use static global variables in quite a few places. I've personally encountered them both in the very front end of TLC when reading & setting command line parameter values for the model check run, and way down in the core of SANY which requires some static initialization.

I hope I don't have to write too many words on why static global variables are bad to form a consensus on why we should want to remove them, but just as a concrete example it recently took me three days of painstaking debugging to figure out I had to call [these two functions](https://github.com/tlaplus/tlaplus/blob/069738ed892c2adcb056db536470d0b0050f3e17/tlatools/org.lamport.tlatools/test/tla2sany/ParseCorpusTest.java#L44-L45) in order to initialize the SANY parser so I could call it directly and have parsing succeed. That was a lot of energy that could have been better directed toward other things, and I anticipate having to expend similar quantities of time & energy to do testing or development on other parts of the codebase. It is simply much, much easier to write tests for things when you can just call a function and don't have to perform some weird incantation first, and don't have to worry about resetting some global state in between tests. For example, it is not possible to run TLC twice within the same JUnit test. At one point in 2020 I hacked together some Java classloader code to unload then load the tlatools jar on request to reset the global state, but it barely worked.

We can broadly split this problem into two categories with some overlap:

1. Static global variables that are used as invisible parameters affecting the behavior of functions far down the stack. These should be extracted & encapsulated into context objects that are propagated down to whatever function consumes them, and the lifecycle of the context object should be managed by the top-level calling function.
2. Static initialization of datastructures requiring you to call several other "incantation" functions before a target function will work as expected. The code in these incantation functions should be moved into `static` blocks that are automatically executed once when the tlatools jar is loaded, or moved into a constructor which must be called before the target function can be accessed.

I understand there is not a great appetite for refactoring in and of itself here, but this particular type of refactoring I believe should be supported and prioritized because of its multiplicative effect on the ease of all other codebase development. What does everybody think?

lemmy:
Addendum: Leslie Lamport and Simon Zambrowski attempted a similar refactoring a decade ago and gave up  because it turned into a time sink.

ahelwer:
Certainly any such refactoring would be incremental and probably take several years to complete across many PRs.

Calvin-L:
I am very much in favor.

[...]

PR #894, opened 2024-03-19
Test operator precedence & associativity parsing

ahelwer:
These tests exhaustively check all pairwise interactions of infix, prefix, and postfix operator symbols. 10,000 parse inputs are checked in approximately one second. These tests found the following two bugs:
- #892
- #893

They serve as a solid basis for near-future work changing the SANY initialization process to not use a JavaCC-generated "Configuration" grammar to parse a string to load all the operator details into itself, which will also facilitate some work on #891.

There is no risk to approving these changes as it is only unit test code.


lemmy (approved):

ahelwer:
@lemmy it would also be useful to have this in the main branch for further PRs.

PR #259, opened 2019-03-20
Javacc4 config input file

Henne:
This patchset applies the code from Configuration.java to the input file config.jj which is used to generate a fresh parser.
**Only config.jj is changed** in this patchset, so **nothing can break**.

A generated parser by JavaCC V4.0 from this config.jj has still differences to the currently used code,
but they are either trivial or unmodifiable by editing config.jj.

This is the diffstat of a patch from the current Configuration.java to a fresh one from JavaCC 4.0:

Configuration.java |   95 +++++++++++++++++++++--------------------------------
 1 file changed, 38 insertions(+), 57 deletions(-)

Best regards.

P.S.: I fully understand and agree with your arguments and I'm not discouraged by your rejection.
Also I understand that my initial pull request [#248](https://github.com/tlaplus/tlaplus/pull/248) was too ambitious and could have been made more comprehensible.

Sadly I could not reproduce the failed JUnit-Tests you mentioned in [#254](https://github.com/tlaplus/tlaplus/pull/254#issuecomment-473422610), since I ran them also on my machine where they did not fail. I'm using:
openjdk version "1.8.0_191"
OpenJDK Runtime Environment (build 1.8.0_191-8u191-b12-2ubuntu0.18.04.1-b12)
OpenJDK 64-Bit Server VM (build 25.191-b12, mixed mode)

PR #895, opened 2024-03-20
Statically initialize operators table

ahelwer:
### What these changes do
`Operators.DefinitionTable` is one of three separate tables that must be initialized before SANY will parse anything (the others are `Context.initialContext` and `BuiltInLevel.LevelData`). Prior to these changes, `Operators.DefinitionTable` was initialized by calling `Configuration.Load` which used the JavaCC-generated parser in `tla2sany/configuration` to parse a string in `tla2sany/configuration/ConfigConstants.java` containing precedence & associativity info for all the TLA+ operator symbols, then loading those symbols into the table. This changes that operation into a simple loop in a `static` block over a static array of `Operator` objects.

### Benefits of these changes

1. Partially (but not completely) removes the necessity of calling `Configuration.load` before using SANY, in line with the goal of #891 
2. Enables defining unicode operator symbols, which the ancient JavaCC-generated parser code did not support reading from a string
3. Moves toward being able to delete the Configuration parser entirely, which would close issue #246 and PR #259 in addition to getting rid of a lot of fairly pointless code

### Reasons to believe these changes will not break anything

1. The `Operators.DefinitionTable` in particular is heavily covered by the operator precedence tests from #894 and corpus tests of #886
2. The changes are very contained and easy to understand, just loading some objects into a table.
3. The Operator objects were not transcribed by hand and thus prone to error; I auto-generated the code by putting the following snippet in the `Operators.addOperator` method then running the original JavaCC parser method:

```java
String assoc = null; 
switch (op.Associativity) { 
  case Operators.assocNone: assoc = "Operators.assocNone"; break; 
  case Operators.assocLeft: assoc = "Operators.assocLeft"; break; 
  case Operators.assocRight: assoc = "Operators.assocRight"; break; 
  default: assert(false); 
} 
String fix = null; 
switch (op.Fix) { 
  case Operators.nofix: fix = "Operators.nofix"; break; 
  case Operators.prefix: fix = "Operators.prefix"; break; 
  case Operators.infix: fix = "Operators.infix"; break; 
  case Operators.postfix: fix = "Operators.postfix"; break; 
  case Operators.nfix: fix = "Operators.nfix"; break; 
  default: assert(false); 
} 
String rep = String.format( 
  "    new Operator(%-16s %-4s %-4s %s, %s),", 
  "\"" + name.toString().replace("\\", "\\\\") + "\",", 
  String.valueOf(op.Low) + ",", 
  String.valueOf(op.High) + ",", 
  assoc, 
  fix 
  ); 
System.out.println(rep);
```

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/configuration/ConfigConstants.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/configuration/ConfigConstants.java
@@ -10,19 +10,10 @@
 
 package tla2sany.configuration;
 
-interface ConfigConstants {
-  int nada = 0;
-  int infixOP = 1;
-  int prefixOP = 2;
-  int postfixOP = 3;
-  int preINfixOP = 4;
+import tla2sany.parser.Operator;
+import tla2sany.parser.Operators;
 
-  int leftAssoc = 10;
-  int rightAssoc = 11;
-  int noAssoc = 12;
-
-  int EOF = 0;
-  int SINGLE_LINE = 4;
+public interface ConfigConstants {
   int CONSTANT = 7;
   int OPERATOR = 8;
   int INFIX = 9;
@@ -35,14 +26,9 @@ interface ConfigConstants {
   int RIGHTASSOC = 16;
   int NOASSOC = 17;
   int BUILTIN = 18;
-  int OPCHAR = 19;
-  int LETTER = 20;
   int OPID = 21;
   int NUMBER = 22;
   int RESTRICTED = 23;
-
-  int DEFAULT = 0;
-  int IN_COMMENT = 1;
 
   String[] tokenImage = {
     "<EOF>",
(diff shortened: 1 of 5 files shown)


lemmy:
@ahelwer @Calvin-L Please don't wait for my review. I won't have the bandwidth to review anything in the next couple of days.

tlatools/org.lamport.tlatools/src/tla2sany/parser/Operators.java
```diff
@@ -1,121 +1,149 @@
-// Copyright (c) 2003 Compaq Corporation.  All rights reserved.
```
Calvin-L, line 1: I don't know all the rules, but I think these copyright notices should be retained unless the file is _completely_ deleted or replaced. (You've rewritten a lot of this file, but several names/comments/methods are unchanged, only reindented.)

tlatools/org.lamport.tlatools/src/tla2sany/configuration/Configuration.java
```diff
@@ -187,16 +187,18 @@ static final public void OpBody(String s) throws ParseException {
    } else {
      op = new Operator( UniqueString.uniqueStringOf(t.image), low, high, assoc, kind );
    }
-   Operators.addOperator( UniqueString.uniqueStringOf(s), op );
+   // This is initialized statically now.
+   //Operators.addOperator( UniqueString.uniqueStringOf(s), op );
```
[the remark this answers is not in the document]
ahelwer: That is the hope, although I'm less familiar with the other static tables and they are more involved with the semantic and level-checking part of the system. So I need to look at them to figure out whether I should write more tests before changing anything there.

tlatools/org.lamport.tlatools/src/tla2sany/parser/Operators.java
ahelwer: Maybe a @lemmy question, to my understanding Compaq is defunct and everything to do with copyright is assigned to the Linux Foundation?

lemmy: IANAL, but I would just leave it there.

tlatools/org.lamport.tlatools/src/tla2sany/configuration/ConfigConstants.java
```diff
@@ -73,141 +59,10 @@ interface ConfigConstants {
...
-    "operator \\prec 50 50 None infix\n" +
-    "operator \\succ 50 50 None infix\n" +
-    "operator \\preceq 50 50 None infix\n" +
-    "operator \\succeq 50 50 None infix\n" +
-    "operator \\sim 50 50 None infix\n" +
-    "operator \\simeq 50 50 None infix\n" +
-    "operator \\ll 50 50 None infix\n" +
-    "operator \\gg 50 50 None infix\n" +
-    "operator \\asymp 50 50 None infix\n" +
-    "operator \\subset 50 50 None infix\n" + // subseteq is builtin
```
lemmy, line 199: Let's keep this comment.

tlatools/org.lamport.tlatools/src/tla2sany/configuration/ConfigConstants.java
```diff
@@ -73,141 +59,10 @@ interface ConfigConstants {
...
-    "operator ... 90 90 None infix\n" +
-    "operator .. 90 90 None infix\n" +
-    "operator | 100 110 Left infix\n" +
-    "operator || 100 110 Left infix\n" +
-    "operator && 130 130 Left infix\n" +
-    "operator & 130 130 Left infix\n" +
-    "operator $$ 90 130 Left infix\n" +
-    "operator $ 90 130 Left infix\n" +  
-    "operator ?? 90 130 Left infix\n" +
-    //"operator ? 90 130 Left infix\n" +  // Removed requested by Leslie (16 Feb. 01)
```
lemmy, line 148: Lets keep this comment.

tlatools/org.lamport.tlatools/src/tla2sany/configuration/Configuration.java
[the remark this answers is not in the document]
lemmy: I'm not aware of users that make use of config parsing.

tlatools/org.lamport.tlatools/src/tla2sany/configuration/ConfigConstants.java
ahelwer: Deal, but if these two are your only requested changes I would like to make them in the unicode changes just to avoid the merge conflict headache

lemmy:
LGTM except that I'm worried that typos could have been made in the precedence table.

ahelwer:
I did generate the table automatically out of fear of typos, and yes the precedence tests would catch typos; that's how I found this issue: https://github.com/tlaplus/tlaplus/issues/892

lemmy (approved):

ahelwer:
@Calvin-L do you sign off?

ahelwer:
@lemmy when you say tag the commits with [changelog] do you mean amend them to have that in the title? You can also add that to the title when you squash/merge.

Calvin-L (approved):

lemmy:
> @lemmy when you say tag the commits with [changelog] do you mean amend them to have that in the title? You can also add that to the title when you squash/merge.

I would be good if you could amend the commit that removes config parsing and add a few sentences along with the tag what the feature was and that is has been dropped.

ahelwer:
@lemmy done!

PR #896, opened 2024-03-22
Add unicode support to SANY

ahelwer:
Some of the files in this changeset are autogenerated from `tla+.jj`, so ignore them:
- `TLAplusParser.java`
- `TLAplusParserConstants.java`
- `TLAplusParserTokenManager.java`

Mostly these changes just involved adding unicode lexeme options to the definitions in `tla+.jj`, and adding unicode operator synonyms to `ConfigConstants.java`. The only parsing logic changes were in `BracketStack.java`, where I just changed it from storing the token end column to the token start column. My main fear was that JavaCC would use byte offsets instead of codepoint offsets to calculate columns, but happily they use codepoints! So everything really was almost too easy.

There were also some simple changes to the level-checking AST generation code in `semantic/Generator.java`. It branched based on the ASCII value of some constructs; I switched it to use the parser node kind. These errors were caught by the tlaplus/examples integration tests described below.

For testing, I added a whole bunch of unicode cases to the syntax corpus parsing test. I also added a job to the PR CI so it runs the tlaplus/examples integration tests after converting all the specs to unicode with TLAUC. This runs in parallel to the other jobs so does not increase CI run time.


tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/Naturals.tla
```diff
@@ -12,6 +12,7 @@
 (* is defined in the static block at the beginning of the Java class.      *)
 (***************************************************************************)
 Nat       == { }
+ℕ         == Nat
```
lemmy, line 15: Naturals is the most fundamental module.  I suggest to hold off on this and gain more confidence with unicode in TLA+ files before the standard modules migrate.

lemmy:
Changed the Unicode Github label from "Unicode support for the TLA+ Toolbox" to "Unicode support for TLA". :-)

tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/Naturals.tla
lemmy: Does TLAPS handle unicode symbols in modules?

ahelwer: I'll check but probably not. Perhaps it would be better to hide/embed this definition in the Java code version of the module somehow.

lemmy: How can user use these defs if they are hidden in Java code?  Also, wouldn't that cause problems with e.g. your tree-sitter parser, because it won't see the symbol?

ahelwer: The tree-sitter parser only resolves symbols within a single file, it doesn't do imports (I would have to integrate with [stack graphs](https://github.com/github/stack-graphs) for that) so there wouldn't be a problem there. How I imagine it working for unicode in general is the user has a plugin for their editor similar to [this one](https://github.com/tlaplus-community/tlaplus-nvim-plugin) I made for neovim, and as they type it auto-replaces symbols with their unicode equivalents. So if they type `Nat` it will automatically replace it with `ℕ`, and then the java overload of the `Naturals` module will automatically resolve that to `Nat`.

ahelwer: After poking around a bit I don't think there's a way to easily add an extra definition to the standard modules without it being in the .tla file. Unless you have any ideas?

lemmy: SANY/TLC support overriding any module, include the standard modules (somebody has to check how that works with the Java module overrides).  For the time being, adopters of the Unicode feature could perhaps provide a user-defined version of Natuals, ...?

ahelwer: That would be acceptable, they can define `ℕ == Nat` at the top of their .tla file. When the change is eventually integrated into the modules themselves it would be a breaking change for them but the fix is simple enough that that's probably all right.

ahelwer:
@lemmy @Calvin-L  I believe this is ready for review. I ran it against all the tlaplus/examples specs (after adding some shims for the standard module number sets) and they parse & model-check with the same number of states when translated to their unicode form! I also added these checks to the PR workflow but this might make it run too long so they could be split out into a workflow that runs only periodically.

lemmy:
Impressive!  Where is this WRT e.g. error locations in TLC stack traces, which won't be tested through the examples?

ahelwer:
@lemmy I haven't tested that but can take a look. There are also a few things I suspect do not work:

- Translation to LaTeX
- PlusCal translation

Those can be included in this batch of changes or I can add them in another.

ahelwer:
@lemmy I checked the `specifications/CheckpointCoordination/MCCheckpointCoordinationFailure.cfg` model with everything translated to Unicode and all the line/column numbers in the trace were as expected.

I think it would be good to get this PR in then I work on PlusCal separately for the remaining time. Hopefully I can finish that this week and maybe have time to add an optional unicode translate-as-you-type feature to the VS code extension.

lemmy:
That's a great, but what I was actually referring to are the evaluation errors leading to TLC outputting a stack trace, not a counterexample. Our most extensive testing of evaluation errors consists of TLC's unit tests suite with its specs/models found in the test-models directory.

.github/workflows/pr.yml
```diff
@@ -1,95 +1,195 @@
-name: PullRequest builder TLA+
+name: TLA+ PR Validation
```
lemmy, line 1: Can the CI changes be moved to a dedicated PR?

ahelwer: They're dependent on this one because of the inclusion of the unicode stuff. I guess I could split them out to their own PR and then rebase this PR on top of that but I guess I question the purpose of all that effort.

lemmy: divide et impera

.github/workflows/pr.yml
```diff
@@ -1,95 +1,195 @@
...
+        cp tlatools/org.lamport.tlatools/dist/tla2tools.jar communitymodules/tlc/
+        ant -f communitymodules/build.xml -Dskip.download=true   
 
-      ##
-      ## Trigger run of examples as integration tests.
-      ##
-    - uses: actions/checkout@v2
+  toolbox-build-and-test:
+    name: Eclipse Toolbox Build & Test
+    runs-on: ubuntu-latest
```
lemmy, line 38: Building (and testing) the Toolbox used to be possible on Win, Mac, and Linux prior to this PR.

.github/workflows/pr.yml
ahelwer: Opened PR here: https://github.com/tlaplus/tlaplus/pull/900

.github/workflows/pr.yml
ahelwer: Continuing this conversation here: https://github.com/tlaplus/tlaplus/pull/900#issuecomment-2028845070

ahelwer:
@lemmy @Calvin-L is there any additional testing you want to see here? I think the coverage is very good.

lemmy (approved):

tlatools/org.lamport.tlatools/src/tla2sany/configuration/ConfigConstants.java
```diff
@@ -260,24 +262,72 @@ public interface ConfigConstants {
...
-    new String[] {"\\union", "\\cup"},
-    new String[] {"\\o", "\\circ"},
-    new String[] {"\\oplus", "(+)"},
-    new String[] {"\\ominus", "(-)"},
-    new String[] {"\\odot", "(.)"},
-    new String[] {"\\oslash", "(/)"},
-    new String[] {"\\otimes", "(\\X)"},
+    new String[] {"\\lnot", "~", "\\neg", "¬"},
+    new String[] {"[]", "□"},
+    new String[] {"<>", "◇"},
```
Calvin-L, line 267: This PR looks great, but I will insist on one small change:

Because we now have non-ASCII characters in the Java code, we should explicitly specify the source encoding in all of the `<javac>` Ant tasks. [There is an `encoding` option for this.](https://ant.apache.org/manual/Tasks/javac.html) I am guessing it should be UTF-8 (although I haven't actually checked what encoding you used for these characters in the Java code).

A couple of stray notes ---

 - While UTF-8 seems to be winning the unicode encoding war, it is still far from a universal default. ISO-8859-1 and UTF-16 are also popular. (Fun fact! ISO-8859-1 is the official default encoding for HTTP, not UTF-8, despite what lots of code seems to believe---leading to many corrupted web pages that are written by server software using UTF-8 and reinterpreted by standards-compliant browsers as ISO-8859-1. This seems to be improving with time, but I find it is still a frequent problem.)
 - To my surprise, the Java Language Specification is [strangely silent](https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html) on the subject of Java's default encoding. Since it only mentions ASCII, I am guessing it only _promises_ ASCII compatibility. Fortunately javac has an encoding argument to let us be explicit, and it does support many common encodings.
 - To my continued surprise, the major build systems (including Ant) are also silent on the subject of default Java source encoding. Again, I think we can only assume ASCII compatibility unless we set it to something else.

Calvin-L: Also, when we document unicode support for TLA+ end users, we should be explicit about what character encoding to use for TLA+ code. Previously it was ASCII, and now it has been expanded to UTF-8. Unlike Java and HTTP, there are no legal alternative encodings for TLA+ (which I think is good).

lemmy:
Have you considered ⊥ and ⊤ as synonymous to `FALSE` and `TRUE`?  The latter are especially verbose when printing values in e.g. counterexamples.

ahelwer:
Sure, we could do that. It would take a bit of legwork to get that in the tree-sitter grammar and things are a bit gummed up right now for TLAUC (see https://github.com/tlaplus-community/tlauc/pull/10#issuecomment-2021478768) but it would not be too difficult to add. `λ` as a substitute for `LAMBDA` is another one I was looking at adding.

I guess one thing to consider is whether to auto-translate `TRUE` and `FALSE` to `⊥` and `⊤` with TLAUC or the various as-you-type editor plugins. Currently those don't take any configurations for translating only subsets of the symbols, which is nice and simple. But some people might find `⊥` and `⊤` confusing and so not desire it.

lemmy:
`:>` and `@@` are two more examples that new users typically find confusing.

PR #900, opened 2024-03-31
Parallelize PR CI workflow

ahelwer:
These steps used to execute sequentially and now execute in parallel. This uses three CI VMs instead of one, but now the PR CI completes in 15 minutes instead of 30 minutes. The jobs are as follows:
1. tlatools unit tests (this takes the longest, ~15 minutes)
2. eclipse toolbox maven build/test (~5 minutes)
3. tlaplus/examples integration tests (~10 minutes)

I also upgraded some of the github action versions so they stop printing out node version warnings.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,95 +1,114 @@
-name: PullRequest builder TLA+
+name: TLA+ PR Validation
 on: [pull_request]
 
 jobs:
-  build:
+  tlatools-build-and-test:
+    name: TLA+ Tools Build & Test
     runs-on: ubuntu-latest
-    env:
-      TLAPS_VERSION: 202210041448
-      MVN_COMMAND: xvfb-run mvn -Dtest.skip=true -Dmaven.test.failure.ignore=true
     steps:
-
-    - uses: actions/checkout@v2
+    - name: Clone tlaplus/tlaplus
+      uses: actions/checkout@v4
       with:
         # Number of commits to fetch. 0 indicates all history.
         # jgit task nested in customBuild.xml fails without history.
-        fetch-depth: '0'
-        
+        fetch-depth: 0
     - name: Set up JDK 11
-      uses: actions/setup-java@v1
+      uses: actions/setup-java@v4
       with:
+        distribution: adopt
         java-version: 11.0.3
-
-      ##
-      ## Run TLC tests.
-      ##
-    - name: Run TLC tests
+    - name: Build & Test Tools
       run: ant -f tlatools/org.lamport.tlatools/customBuild.xml -Dtest.halt=true compile compile-test test dist
-
-      ##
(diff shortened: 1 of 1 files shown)


ahelwer:
@lemmy continuing from your comment here: https://github.com/tlaplus/tlaplus/pull/896#discussion_r1545754310

Do you want to add a matrix so the CI runs on Windows too? Could also do macOS but that tends to be slower. However if it's just for the eclipse toolbox build it should still complete within 15 minutes, which is how long the tlatools unit tests take. Only concern is efficiency of VM usage I guess.

lemmy:
> Do you want to add a matrix so the CI runs on Windows too? Could also do macOS but that tends to be slower. However if it's just for the eclipse toolbox build it should still complete within 15 minutes, which is how long the tlatools unit tests take. Only concern is efficiency of VM usage I guess.

The current CI has to run the Toolbox build on macOS to properly [sign](https://github.com/tlaplus/tlaplus/blob/23f765078cd098a2bc2bc3d19889ba0c64583ca4/.github/workflows/main.yml#L255-L292) the Toolbox on macOS.  

"(Linux)" is wrong: https://github.com/tlaplus/tlaplus/blob/master/.github/workflows/main.yml#L112

ahelwer:
Oh okay so what does that mean for the PR workflow?

lemmy:
The maven build is platform independent: it always creates a Windows, macOS, and Linux build artifact regardless on what platform it runs on.  However, the macOS build artifact goes through a post-processed step (appstore sign), before it is uploaded to our download locations.  Thus, for as long as there is a Toolbox, the Toolbox build has to run on macOS.  A second reason to run the build on Windows, macOS, and Linux is to run UI tests on all platforms.  Unfortunately, this part has been defunct on Windows for a while.

ahelwer:
Well I'm assuming the post-process signing step isn't relevant for the PR CI because the build artifact isn't uploaded anywhere, but do you want the maven build run on windows/linux/macOS in the PR CI or what?

lemmy:
I suggest keeping the maven part of PR.yml until we have a firm plan when to sunset the Toolbox.

ahelwer:
The maven part is kept, it is just run in parellel now.

lemmy:
But can it run on all three platforms?

ahelwer:
Sure, I added a matrix in the maven job to do that.

ahelwer:
@lemmy okay well the Windows build doesn't seem to actually work, what do you want to do about that

lemmy:
Keeping the Toolbox on a lifeline seems to be too much work.   Just go ahead without Windows and lets discuss the Toolbox's EOL at the next Community Meeting.

ahelwer:
Okay done. Workflow succeeds.

lemmy (approved):

PR #903, opened 2024-04-09
Fix unit test error reporting and support `-Dthreads=N` parameter

ahelwer:
The threads parameter limits parallelization so users can reserve part of their workstation for other tasks as the tests are running, although if that task is running TLC then it might interfere with the tests.

### DIFF
--- a/tlatools/org.lamport.tlatools/test/tlc2/tool/liveness/ModelCheckerTestCase.java
+++ b/tlatools/org.lamport.tlatools/test/tlc2/tool/liveness/ModelCheckerTestCase.java
@@ -171,8 +171,16 @@ public abstract class ModelCheckerTestCase extends CommonTestCase {
 				args.add("1");
 			}
 			
+			// Some tests require a large number of threads to successfully
+			// test their desired property. They request this by overriding
+			// the getNumberOfThreads() method. This can lead to over-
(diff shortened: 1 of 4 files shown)


tlatools/org.lamport.tlatools/test/tlc2/tool/liveness/ModelCheckerTestCase.java
```diff
@@ -171,8 +171,16 @@ public void setUp() {
...
+			// test their desired property. They request this by overriding
+			// the getNumberOfThreads() method. This can lead to over-
+			// subscribing system cores when running tests in parallel, but
+			// is usually okay because the tests run for only a second or so.
+			// If a test runs for a longer period of time using many threads
+			// then it should be run sequentially instead of in parallel.
 			args.add("-workers");
-			args.add(Integer.toString(getNumberOfThreads()));
+			int requestedThreads = this.getNumberOfThreads();
+			args.add(Integer.toString(requestedThreads));
```
Calvin-L, line 183: Nit: I don't see a need for a new variable here (but it clearly isn't harmful).

(Also, thanks for adding the comment above. 👍 )

Calvin-L (approved):

lemmy (approved):

PR #907, opened 2024-04-10
Rework `BufferedRandomAccessFile`

### LINKED ISSUE #835: Java 21 causes TLC's BufferedRandomAccessFileTest and OffHeapDiskFPSetTest to fail

Calvin-L:
This commit makes major alterations to `BufferedRandomAccessFile` to bring it in line with the general `RandomAccessFile` contract:

 - Invariant V1 has been strengthened to include a contract for the `length` field.
 - Invariants V2-V5 have been altered to account for cases where `curr > length`, which the `RandomAccessFile` contract allows.
 - Additionally, invariant V5 has been adjusted to clarify that `dirty=true` does not imply that there are unflushed bytes.  The implication is only one-way: if there are unflushed bytes, then `dirty=true`.
 - The `hi` and `maxHi` fields have been removed.  These largely duplicate information that can be quickly recomputed, meaning they add additional complexity to the class invariants.  Without those fields, invariant V6 can be removed.
 - The `close()` method now closes the underlying file descriptor even if `flush()` throws an exception.

Due to the complexity of this change, I have added a TLA+ specification for `BufferedRandomAccessFile` to enable model checking of its design.

To ensure that the code is correct, this commit also adds a substantial amount of test infrastructure:

 - Several new regression tests have been added.  These are a little cryptic because they were randomly generated, but they do reflect true behaviors of the `RandomAccessFile` contract.
 - A fuzz test has been added.  This test ensures that many sequences of random operations behave correctly.  It also minimizes failures so they can be easily converted to regression tests.

This commit also makes some whitespace and formatting adjustments.

Fixes #835.  I have verified that the tests now pass under Java 21.


lemmy:
Impressive!

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,384 @@
...
+    ELSE ArbitrarySymbol
+
+BufferedIndexes == lo .. (Min(lo + BuffSz, length) - 1)
+
+Inv1 ==
+    /\ diskPos = file_pointer
+    /\ length = ArrayLen(LogicalFileContent)
+
+\* Inv2 is a bit special.  Most methods restore it just before they return.  It
+\* is generally restored by calling `seeek(curr)`.  But, that behavior is
```
ahelwer, line 70: spelling

lemmy: Next to `seek`, `seeek` is a method in BRAF.

Calvin-L: Nope, `seeek` is actually the correct name! Maybe I should add [sic] here.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,384 @@
+\* Copyright (c) 2024, Oracle and/or its affiliates.
```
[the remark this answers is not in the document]
lemmy: IANAL: The copyright remains with Oracle, but the DCO makes sure that the TLA+ foundation may use the code in any way it likes within the bound of the open source license (MIT in our case).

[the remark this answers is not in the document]
ahelwer: I think it would be worth emailing your contact with the Linux Foundation with this question.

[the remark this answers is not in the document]
Calvin-L: My company requires these notices on any code I write on company hardware and time. (Even for internal code!)

----

If you want more info than that...

I am not a lawyer, so I can't claim to fully understand the requirement, but here's what I know (or think I know anyway):
 - These notices do not change the project _license_ (MIT) in any way. Everyone is still free "to deal in the Software without restriction" and so forth. As you know, Oracle even signed a CLA attesting that it understands that fact.
 - These notices do not claim _exclusive_ copyright, and are compatible with overlapping copyrights from other organizations. Again, see the project license.
 - These notices primarily serve as glorified authorship information, and they might be best read by non-lawyers as meaning "this file contains code that was originally owned by Oracle and that Oracle is therefore legally allowed to release to the world under the terms of this project's license".

On a personal note, I find the notices aesthetically ugly. But, I do plan to follow the letter of the law so my company will let me keep contributing to TLA+ with company resources.

If the notices are unacceptable in new code, I can have a chat with my company's legal team and see if they'll accept something different. I know in the past they've been open to a [simple blanket copyright in the license file](https://github.com/wilcoxjay/mypyvy/blob/3a055dfad3d13fb35b6125ab6f43fa6ea4493b5f/LICENSE), but that could take a while to approve (and it might not ever be approved for a larger project like TLA+).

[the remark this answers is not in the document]
lemmy: I've been adding a similar Microsoft copyright notice in my contributions, and Microsoft also signed the CLA.  That's how it is done.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,384 @@
...
+    diskPos           |-> diskPos,
+    file_content      |-> file_content,
+    file_pointer      |-> file_pointer,
+
+    \* abstract vars
+    abstract_contents |-> LogicalFileContent]
+
+===============================================================================
+
+--------------------------- MODULE RandomAccessFile ---------------------------
```
IamfromSpace, line 257: I think it's still worth a brief documentation string here, as it's the place I jumped to first to ensure I understood what was going on.  For many folks, even interacting with files in this manner (rather than wholesale reads or writes) is still quite a low level operation.  A suggestion might be:

> Here we model abstractly how one interacts with files assuming a single thread that reads and writes, given a pointer into the file.  The pointer can be moved to an arbitrary position, or it can read or write data linearly from its current position, which simultaneously advances the pointer.  We can modify the length of the file and if the pointer would fall outside of the bounds after truncation, it is moved to the end of the file.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,384 @@
...
+
+===============================================================================
+
+-------------------------------- MODULE Common --------------------------------
+
+EXTENDS Naturals, Sequences
+
+CONSTANTS
+    Symbols,
+    ArbitrarySymbol,
```
IamfromSpace, line 313: Seems like the purpose and value of `ArbitrarySymbol` is worth calling out (maybe here or in `TruncateOrExtendFile` or there might be some better place).  If I understand it correctly, it seems that its to reign in state space explosion by mixing in a bit of symbolic checking?  It appears as though on could consider all possible extensions of a file using `Symbols`, but the downside of this is that we're going to potentially explore many states.  Whether I understood correctly or not, it's probably worth explaining.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,384 @@
...
+EmptyArray == [elems |-> <<>>]
+ArrayLen(a) == Len(a.elems)
+ArrayToSeq(a) == a.elems
+SeqToArray(seq) == [elems |-> seq]
+ArrayGet(a, i) == a.elems[i+1]
+ArraySet(a, i, x) == [a EXCEPT !.elems[i+1] = x]
+ArraySlice(a, startInclusive, endExclusive) == [elems |-> SubSeq(a.elems, startInclusive + 1, endExclusive)]
+ArrayConcat(a1, a2) == [elems |-> a1.elems \o a2.elems]
+
+WriteToFile(file, offset, data_to_write) ==
```
IamfromSpace, line 334: It is interesting to me that most of these functions are about Arrays.  Even `WriteToFile` and `ExtendFile` are relatively easy make generic array operations, maybe `OverlapAt` and `TruncateOrExtendWith`, where both take a placeholder as an argument.  It may simply be worth making this an `Array` module, and then `Common` stays more focused on the spec itself `WriteToFile(file, offset, data_to_write) == Array.Overlap(file, data_to_write, offset, ArbitarySymbol)` (I have no idea why I felt compelled to reverse arguments).

I suppose, otherwise I was going to suggest that we add brief documentation on what the semantics of `WriteToFile` and `ExtendFile` were, but in many ways something like `ExtendFile == Array.TurncateOrExtendWith(file, new_length, ArbitrarySymbol)` does this for you.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,384 @@
...
+    LET length == Max(file_len, offset + data_len) IN
+    MkArray(
+        length,
+        [i \in 0..(length-1) |->
+            CASE
+                i < offset -> IF i < file_len THEN ArrayGet(file, i) ELSE ArbitrarySymbol
+                []
+                i >= offset /\ i < offset + data_len -> ArrayGet(data_to_write, i - offset)
+                []
+                i >= offset + data_len -> IF i < file_len THEN ArrayGet(file, i) ELSE ArbitrarySymbol])
```
IamfromSpace, line 346: i think it is invariant that the final `ELSE` never matches (unless I've missed something).  The only situation in which is greater than or equal to `offset + data_len` is when the original file extends beyond the total overlap of the new data.

I think it makes sense to avoid the `IF`, as otherwise a cursory glance makes it appear that we can end up `ArbitrarySymbol`s at the end of the file, which isn't the case.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
IamfromSpace: I am probably overthinking these utility functions, but it struck me that maybe the most intuitive implementation involves slice & concat.

```
TruncateOrExtend(a, T, len) == IF len > ArrayLen(a) THEN ArrayConcat(a, Array(T, len - ArrayLen(a))) ELSE ArraySlice(a, 0, len)

OverlayAt(a1, a2, T, i) == ArrayConcat(TruncateOrExtend(a2, T, i), a2)
```

lemmy: Should any of those utility operators be contributed as an Array module to our CommunityModules?

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
Calvin-L: Yes, great catch! The whole `IF` can be simplified to its then-branch `ArrayGet(file, i)`.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +190,371 @@ public void invalidateBufferedData() throws IOException {
...
+        // In that case, the file must be extended to contain the write, and the unset part of the buffer becomes part
+        // of the official file contents.  Fortunately the RandomAccessFile contract gives us an out: when the file is
+        // extended in this way, the unwritten portion can be filled with arbitrary data.  Therefore, we do not need to
+        // do anything to the unset parts of buff.
+        //
+        // It is tempting to offer a slightly stronger contract by zero-filling the unset part of "buff":
+        //
+        //     Arrays.fill(this.buff, cnt, BuffSz, (byte)0);
+        //
+        // But do not be tempted!  That call incurs a performance penalty WITHOUT OFFERING A STRONGER CONTRACT!  There
```
IamfromSpace, line 304: This is great to call out this optimization (or trap, depending how you're looking at it) and the reasoning behind it.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -5,149 +5,175 @@
...
+     * @see #reset()
+     */
+    private long mark;
+
+    /** Protects {@link #availBuffs} and {@link #numAvailBuffs}. */
+    private static final Object mu = new Object();
     private static byte[][] availBuffs = new byte[100][];
     private static int numAvailBuffs = 0;
 
     /* To describe the above fields, we introduce the following
```
IamfromSpace, line 72: Before the spec existed, this is clearly the right place for this comment to exist.  However, given that there is now a formal spec for this module, it seems like it might be appealing to put the bulk of this description into the spec, and then refer the reader to the spec here in this file.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -5,149 +5,175 @@
...
 /**
  * A <code>BufferedRandomAccessFile</code> is like a
  * <code>RandomAccessFile</code>, but it uses a private buffer
- * so that most operations do not require a disk access.<P>
- * 
- * Note: The operations on this class are unmonitored.
- * The code was written by Allan Heydon
+ * so that most operations do not require a disk access.
+ *
+ * <p>This class is not thread safe.
```
IamfromSpace, line 16: Since we're specifying and re-examining the properties of this class, it's maybe interesting that it also has slightly weaker guarantees across executions than RAF.  With both RAF and BRAF, should the program crash in the middle of a write to disk, then the value on disk may be corrupted.

However, unique to BRAF, we also lose causality between the state of the file and future side effects.  With RAF, if you do `write()` then `f()`, and you can confirm that `f()` took place, then you also know that the `write()` took place.  With BRAF, it's possible that the `write()` call did not flush before a crash that occurred after `f()`.  So confirming that `f()` took place does not imply that the `write()` call did.

I wouldn't expect this has implications for tlc, but it is interesting to call out as another place where the semantics are weaker.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,384 @@
+\* Copyright (c) 2024, Oracle and/or its affiliates.
+
+----------------------- MODULE BufferedRandomAccessFile -----------------------
```
IamfromSpace, line 3: Just wanted to say that I have absolutely found it useful to have a spec of the class for review, and especially as a refinement.

lemmy: This spec could also be referenced at tlaplus/examples

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -5,149 +5,175 @@
...
-      this.init();
+    /**
+     * Open a new <code>BufferedRandomAccessFile</code> on the
+     * file named <code>name</code> in mode <code>mode</code>,
+     * which should be "r" for reading only, or "rw" for reading
+     * and writing.
+     */
+    public BufferedRandomAccessFile(File file, String mode) throws IOException {
+        super(file, mode);
+        synchronized (mu) {
```
IamfromSpace, line 143: The static buffer reuse is interesting.  Just to confirm my understanding: While it would be simpler if each new instance just created its own buffer each time, this appears to be a performance optimization to allow reuse of buffers in new files after old files are closed.  Then, even though this class isn't thread-safe if the same file is modified by two BRAFs, these `synchronized` sections are key to it being thread safe when _distinct_ files all are open with _different_ BRAFs.  Let me know if I've misinterpreted anything!

Did you consider adding this as part of the specification, given a naive danger of reading previous and different file's data?  I'm not sure it's necessary or worth it, as it seems correct, but figured it was worth discussion.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +190,371 @@ public void invalidateBufferedData() throws IOException {
...
+     * Extends this.seek(long) by return if a page has been read (used for statistics).
+     *
+     * @return true iff data was read from disk
+     */
+    // NOTE: a call to seeek(this.curr) suffices to restore invariant V2 after this.curr changes.
+    //TODO come up with better name for seeek %)
+    public boolean seeek(long pos) throws IOException {
         // Assert.check(!this.closed);
-        if (pos >= this.hi || pos < this.lo) {
+        boolean didPageRead = false;
```
IamfromSpace, line 324: minor, but with the restructure, there's no need for this intermediate variable.  You can set `this.curr` immediately, and then `return true` at the end of the if condition, and `return false` at the end of the method.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +190,371 @@ public void invalidateBufferedData() throws IOException {
...
+    public void setLength(long newLength) throws IOException {
+        // Per docs:
+        //  > If the present length of the file as returned by the length method is greater than the newLength argument
+        //  > then the file will be truncated. In this case, if the file offset as returned by the getFilePointer
+        //  > method is greater than newLength then after this method returns the offset will be equal to newLength.
+        //
+        // In practice, RandomAccessFile actually moves the pointer if getFilePointer > newLength, regardless of
+        // whether the file was truncated or extended.
+        super.setLength(newLength);
+        if (this.diskPos > newLength) {
```
IamfromSpace, line 358: It may be safer to simply synchronize these unconditionally, given that the documentation already doesn't fully define all behaviors.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +190,371 @@ public void invalidateBufferedData() throws IOException {
...
+            this.diskPos = super.getFilePointer();
+        }
+
+        if (this.curr > newLength) {
```
IamfromSpace, line 362: I first thought that maybe this should just always seek to `diskPos`, but I see this is right, because of the subtle cases around end of file.

I think it's worth explaining this case as for why not:
```
If the buffer landed in a way where the end of the file appears in the middle of the buffer,
such as a short file of just 4 bytes.  When we filled the buffer, the diskPos and curr would
both be in the middle of the buffer, at index 4.  Now consider we seek to the beginning of
the file, but then chop one byte off the file.  In this case, `diskPos` will have moved, but
`curr` should not.

         curr and diskPos
         v
[A B C D * * * * ...]

seek(0)

 curr    diskPos
 v       v
[A B C D * * * * ...]

setLength(3)

 curr  diskPos
 v     v
[A B C * * * * * ...]

File: tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java [hunk 15]
```diff
@@ -164,357 +190,371 @@ public void invalidateBufferedData() throws IOException {
...
-                    // appending to EOF -- bump "hi"
-                    // Assert.check(this.hitEOF);
-                    this.hi = this.maxHi;
-                }
-            }
+
+    /**
+     * Write at most "len" bytes from "b" starting at position "off", and return the number of bytes written.
+     */
+    private int writeAtMost(byte[] b, int off, int len) throws IOException {
```
IamfromSpace, line 515: It seems like it may be worth greater optimization here, given that, in theory, that is the motivation for this class.  This implementation makes sense, in that it is quite safe.  If I understand it, we are simply pushing data into the buffer, up to what it can handle, and if we filled the buffer, we use seek to flush the current buffer and move it upwards a single frame, please correct me if I've misunderstood anything!

However, if we write a substantial amount of data, we're a bit wasteful on two fronts.  For one, we copy data into the buffer, which will never be read, when we could directly write the data we have.  For the second, we are going to fill the buffer with data when the frame moves, and do a read from disk for data that we will never read and immediately overwrite.

It seems like we should detect these cases when we're doing a long enough write and bypass the longer route, leaving the buffer in a temporarily invalid state that we know we'll clean up at the very last frame move.  Otherwise, it seems like this class may not be living up to its mission of optimizing file interactions.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,384 @@
...
+
+BufferedIndexes == lo .. (Min(lo + BuffSz, length) - 1)
+
+Inv1 ==
+    /\ diskPos = file_pointer
+    /\ length = ArrayLen(LogicalFileContent)
+
+\* Inv2 is a bit special.  Most methods restore it just before they return.  It
+\* is generally restored by calling `seeek(curr)`.  But, that behavior is
+\* difficult to model in straight TLA+ because each method may modify variables
```
IamfromSpace, line 71: I liked how Finn and co. did a mix of TLA+ and PlusCal in the [Inconsistency in Cosmos DB paper](https://arxiv.org/pdf/2210.13661.pdf) to play to the two's strengths and get clear multi-step behaviors.  It might be worth it to look at something like that for when variables need multiple changes, and then the invariants would only be expected to hold when the program counter is in safe state.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
Calvin-L: To be precise:

 - Each BRAF has exclusive access to its internal buffer (`this.buff`).
 - Buffers can be re-used across instances. `availBuffs` is a global "pool" of previously-allocated buffers.
 - The lock `mu` allows BRAFs in different threads to safely interact with the global pool. They do this once on construction (to obtain a buffer) and once on `close()` (to return a buffer to the pool).

The purpose of the shared buffer pool is not clear to me. I suspect it is intended to reduce GC pressure: we can allocate buffers without using `new` and the garbage collector never needs to reclaim them.

However, I _strongly_ suspect that the optimization is irrelevant:
 - I don't believe TLC allocates/closes many instances of `BufferedRandomAccessFile`
 - Even if it does, the JVM's modern GC is _very_ good and may outperform this manual approach
 - And even if I'm wrong about those points, the `open()` syscall probably dominates `BufferedRandomAccessFile` allocation performance anyway

We could look at removing the shared pool in a separate PR.

> Did you consider adding this as part of the specification, given a naive danger of reading previous and different file's data?

I did not. I think the buffer pool problem is orthogonal, and there's a lot of evidence that the current approach works. (I didn't change it.) If I am wrong and it really is great for performance, then we could consider factoring it out into a separate class and specifying its behavior separately.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
Calvin-L: Great suggestion!

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
IamfromSpace: That all tracks!  I think it makes the most sense to leave it as is for now to separate concerns and keep the focus of this PR.  And it makes sense to leave out of the spec given that, like you said, it wasn't changed and its future is dubious.

For discussions sake, as to the possible future of this approach, I agree that its benefit seems suspect.  Also, if there was need to reuse available buffers, the cost of `arraycopy` of all buffers on expansion seems odd.  If we suddenly opened and then closed 1000 files, this is a lot of copying.  I guess it's only the pointers (right?), given it's an array of arrays, but it's still strange to me.

Even if there was a benefit over JVM GC, it seems like atomic interactions with a `Stack` would provide a much more natural interface and should perform better on release.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
Calvin-L: I am actually going to leave it here, since some of the other comments in this file refer to specific invariants by name, and I think it's helpful to have all that in one file.

But, I will definitely add a note here pointing to the TLA+ formalization.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
IamfromSpace: Good point!

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
Calvin-L: I thought about this as well---but let's defer that to future optimizations. I will add a note about this idea.

FWIW, TLC's workload on this class does not involve large bulk writes. It reads and writes almost entirely single-byte and 4-byte chunks. But, I believe it _does_ alternate between read phases and write phases, so there may be some unnecessary reads during the write phase. Even so, in my microbenchmarks the buffer more than pays for the extra reads.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
Calvin-L: I actually started with PlusCal, but even though it makes modeling easier it adds a lot of specification overhead. Specifically, verifying that the output of `read()` is correct might require an extra `read_output` variable. It gets a little ugly if `read()` is not a single action.

But, definitely something to keep in mind if this class becomes any more complex.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
IamfromSpace: On the topic of future optimizations (not for this PR), it did strike me that making this a ring buffer over an arbitrary position might help avoid cases where the buffer position flaps back and forth between two frames, causing a series of reads and flushes.  It also might better work with an L2 cache by keeping buffer positions more stable in memory for longer.

Though, again, I’d be curious your estimation of how such changes might actually impact the TLC specific workload.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
Calvin-L: I added some simple documentation to these operators.

> but it struck me that maybe the most intuitive implementation involves slice & concat

I agree that the current definitions are a little ugly. I wrote them after one or two failed attempts to come up with simpler definitions.

The ones you've given are certainly cleaner than my early attempts, and I will adopt your definition for `TruncateOrExtend`. But I might keep the definition I wrote for `WriteToFile`. The equivalent definition in terms of other operators is actually fairly cryptic (your definition for `OverlayAt` needs to include the portion of the contents after the written data if the contents were not written at the end of the array):

```tla
WriteToFile(file, offset, data_to_write) ==
    ArrayConcat(ArrayConcat(
        TruncateOrExtendFile(file, offset),
        data_to_write),
        ArraySlice(file, offset + ArrayLen(data_to_write), ArrayLen(file)))
```

> Should any of those utility operators be contributed as an Array module to our CommunityModules?

Yes! I'll look into doing that soon.

Calvin-L:
Thanks a ton @IamfromSpace for the feedback!

I think I've addressed all the major comments so far.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
IamfromSpace: > your definition for OverlayAt needs to include the portion of the contents after the written data if the contents were not written at the end of the array

Whoops!  Easy to make things simpler if you forget the critical parts :laughing:

IamfromSpace:
These updates look great!  For the sake of thoroughness I still have a few comments coming through on tests, but nothing of real concern.  Just wrapped up the unit tests, but it's taking me a bit longer to digest the fuzzing file :)

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileTest.java
```diff
@@ -80,8 +77,8 @@ public void testWriteSeekNoLength() throws IOException {
 			}
 		} catch (IOException expected) {
 			return;
-		} catch (Exception e) {
-			fail(e.getMessage());
+//		} catch (Exception e) {
```
IamfromSpace, line 80: Was this never necessary?  I haven't been able to figure out why this was ever here in the first place :laughing:

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileTest.java
```diff
@@ -104,6 +101,7 @@ public void testRead() throws IOException {
 	}
 	
 	@Test
+	@Ignore("BufferedRandomAccessFile does not (and never did) guarantee zero-fill when extended")
```
IamfromSpace, line 104: nit: I think it makes more sense to just delete this.

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileTest.java
```diff
@@ -170,4 +168,155 @@ public void testInvalidateBufferedData() throws IOException {
...
+
+			byte[] buffer = new byte[100];
+			raf.write(buffer);
+			assertEquals(10000 + buffer.length, raf.length());
+			assertEquals(10000 + buffer.length, raf.getFilePointer());
+		}
+	}
+
+	@Test
+	public void regressionTest01() throws IOException {
```
IamfromSpace, line 216: nit: I personally like giving these clearer names.  It's easier to read them if you have that context of what they intend to validate (or what you broke if they fail!).

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileTest.java
```diff
@@ -170,4 +168,155 @@ public void testInvalidateBufferedData() throws IOException {
...
..., -43, 14, -69, -98, 34, 119, 46, -30, -71, 45, 69, 31, -81, 60, 48, 37, 59, 15, 69, -91, -100, 125, 35, -43, 97, 106, -71, 24, -90, 26, -4, -117, 79, -89, 6, 46, -90, 52, 3, -2, 71, 121, 13, 14, 14, -82, 71, 82, -84, -31, -37, -6, 108, -84, -16, 23, -29, -125, -70, -78, -83, -23, -111, 10, -51, -51, -76, -36, 46, -70, -30, -57, 112, 34, -54, -108, -16, -5, -114, -93, -3, 109, 101, 11, -119, -102, -14, 121, 22, -112, -47, 106, 96, -6, 8, -47, -7, 42, -66, -90, -60, -50, -43, 78, -36, -10, 29, -51, -112, 26, 4, -97, -97, -84, -40, -76, 65, -82, -91, -35, 42, -56, 44, -50, 112, -12, 36, -109, -107, -120, -79, 49, 99, -123, -25, -51, -68, -79, 94, 38, 57, 81, 75, 0, -25, -70, -5, 3, 116, -74, -55, -115, -127, -34, -21, -26, 100, 122, -122, 112, -97, 71, 48, -21, 37, -128, -85, -88, -2, 86, 22, -24, -115, -16, 41, 73, -104, -47, -22, -70, -56, 125, -32, -127, -111, -63, -94, 102, -27, 121, 56, 13, 58, -32, 112, 119, 105, -24, -75, 119, -53, -64, -123, -124, -108, -42, 9, -30, 36, 30, -109, 39, -18, 42, -79, -78, 62, 73, -37, -96, -41, 85, 39, 58, -54, 19, -69, 111, 60, -89, 30, 37, -78, -112, 29, -69, 98, 108, 120, -54, 59, -19, 60, 74, -58, 20, -56, 126, 23, -125, -110, 85, -52, 42 };
```
IamfromSpace, line 295: nit: this seems a bit overkill.  Can't this just be a `raf.write(0)` and then assert that length is 8312?

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileTest.java
```diff
@@ -170,4 +168,155 @@ public void testInvalidateBufferedData() throws IOException {
...
... 42, -81, 88, -99, -84, -71, 100, -69, 45, 127, -108, -119, 44, 62, 30, 42, 42, -63, 111, -68, 14, -123, -8, -90, -1, 71, 47, 122, 77, 69, 33, -17, 97, 110, 19, 75, -2, -63, 71, 69, 23, 28, 113, -14, -17, 60, -20, 85, 34, 100, 85, 28, -50, 64, 44, -29, -52, -54, -38, -83, 46, 89, -114, -89, -14, -83, -53, 93, 37, -7, 60, -51, -1, -58, 29, 16, 97, 37, 13, 9, 126, -99, 90, 34, -34, 4, 27, 59, 40, -46, 123, -128, -94, 83, 68, 101, -48, -67, -96, -48, 47, -27, -40, -101, -85, -60, 13, -81, -115, -128, 116, 9, -67, 69, 82, -128, 13, 118, -82, -103, 112, 7, 62, -19, 70, -17, 19, -57, -94, -67, -76, -76, -49, 51, 110, 113, -78, -110, 120, 50, -64, -127, 32, -43, 79, -52, 66, -56, 17, -73, 86, 122, 22, -122, -37, 19, 61, 113, 98, 121, 66, 95, -117, -127, -37, -118, 37, 47, -2, -83, -48, 46, 16, -117, 3, 27, 23, -94, -34, -38, 1, -63, -2, 79, 31, -118, 41, 44, -86, -11, -42, 43, 63, 121, -114, -24, -52, -24, 14, 16, 80, -87, -36, 72, -45, -63, -23, 0, -16, 115, -81, -45, -39, 75, 3, -40, -91, 110, 28, 4, 55, -52, -76, 109, 92, 105, -94, 28, 65, -107, -50, 65, -68, -109, 108, 77, 35, -77, -52, -77, 36, 53, 126, 72, -6, 70, -18, 64, 86, -4, -36, 1, -111, -28, 116, 19, 15, 71, -12, -15, 124 };
```
IamfromSpace, line 311: nit: It seems like we can create this case more easily.

Doesn't the following trigger the original `AssertionError`?

```
raf.seek(9247);
raf.setLength(71);
raf.seek(1675);
raf.write(87);
```

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileFuzzTest.java
```diff
@@ -0,0 +1,553 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class BufferedRandomAccessFileFuzzTest {
```
IamfromSpace, line 21: This is great, I'm a big fan of fuzz testing.  There's a lot of code here, but I think I'm maybe I've become spoiled in the Haskell and QuickCheck world :)

My main recommendation is that we call out the overall strategy here to present a mental framework for anyone who needs to work in here in the future.  The main thing that I was slow to put together was that (and how) invalid lists of operations were pruned.  But in retrospect, the approach is quite obvious, so I think a short blurb would go a long way.

Something like: "Fuzzing here generates many random legal lists of file operations and then validates that executing those operations has the same outcome, whether applied to a `RandomAccessFile` or a `BufferedRandomAccessFile`.  When randomly generating operation lists, we simulate these lists on an in memory representation of the file (`AbstractFileState`) which has helpers to indicate if an operation that would result in undefined behavior.  Such illegal operations are then pruned before checking."

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileFuzzTest.java
```diff
@@ -0,0 +1,553 @@
...
+        for (int i = 0; i < nOps; ++i) {
+            ops.add(gen.nextRandomOperation());
+        }
+        if (!legal(ops)) {
+            throw new IllegalStateException("generated illegal ops: " + ops);
+        }
+        return ops;
+    }
+
+    private void generateRandomOpsAndCheck(Random rng, int nOps) throws IOException {
```
IamfromSpace, line 108: This function appears to be redundant.  Even if it had some value in debugging (taking a guess at its purpose), I still think it makes more sense to remove it given that it has otherwise duplicated code.

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileFuzzTest.java
```diff
@@ -0,0 +1,553 @@
...
+        public String toString() {
+            return "RunResult{" +
+                    "opsExecuted=" + successes +
+                    ", expected=" + expected +
+                    ", actual=" + actual +
+                    '}';
+        }
+    }
+
+    private RunResult checkOps(java.util.List<FileOperation<?>> ops) throws IOException {
```
IamfromSpace, line 221: I might recommend calling this `checkLegalOps`.  I think `check` and `legal` make as much sense as any names, but I did have a moment of confusion between the purpose of the two.  This name might make clear that the function's domain is only legal operations, and we should therefore always expect that we're going to have to check their legality first.

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileFuzzTest.java
```diff
@@ -0,0 +1,553 @@
...
+
+		final File tmpFile1 = File.createTempFile("tmp", ".bin");
```
IamfromSpace, line 223: nit: spacing appears to be accidental

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileFuzzTest.java
```diff
@@ -0,0 +1,553 @@
...
+                    return new RunResult(ops, i, a, b);
+                }
+            }
+
+        }
+
+        return new RunResult(ops, ops.size(), null, null);
+    }
+
+    private boolean legal(java.util.List<FileOperation<?>> ops) {
```
IamfromSpace, line 250: minor, but it struck me that your construction of `FileOperation` here is pretty clever, to where it seems like like could instead be the `simulateIfLegal` method on `ListOfOperations` class that implements `FileOperation`.  Seems like it has a valid `execute` and `toString`.

IamfromSpace (approved):
Alrighty, that's 100% everything from me :)  I think there's some test comments that are worth a look over, but there's certainly no deal-breakers in there, so I've got the approval on there now.  Not sure if the repo is configured to reset on changes or not, but if so, I'll keep it active for any commits that come in.

Nice work on this!

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileTest.java
Calvin-L: There's a lot to say about these `regressionTest*` methods that should be written in a comment.

TL;DR: These were generated by a computer, so they don't look particularly sensible, nor can I really assign them sensible names. I don't plan to change them (see below), but I am open to adding additional focused handwritten tests inspired by them.

----

The `regressionTest*` methods are all randomly-generated failures discovered by the fuzzer at some point while I was working on this PR. I copied them unchanged from the fuzzer output (although in a few cases I added a comment about what should happen for a particular test case).

The fact that these cases failed during development indicates they are testing _something_ that is tricky to get right in the implementation, and so I think it's worth having them here to make sure they are tested regularly.

I don't particularly want to change them because:
 - I am not confident I can minimize them reliably; the failures happened on intermediate versions of `BufferedRandomAccessFile` that are now lost. If I make a change and the test still passes, is it _really_ testing the same thing it was before?

And I don't want to remove them either:
 - Some of these came from an earlier version of the fuzzer (i.e. the current fuzzer doesn't cover them exactly as written here).
 - Small changes to the fuzzer generally result in large changes to explored traces (i.e. future versions of the fuzzer may not explore these specific cases).

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileFuzzTest.java
Calvin-L: Good catch; this is leftover from some intermediate version and should be removed.

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileTest.java
Calvin-L: It was not. 😆 It does exactly one thing, which is to hide the stack trace from the thrown exception `e`. That one thing is bad, so I've removed these catch blocks.

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileFuzzTest.java
Calvin-L: Now that you mention it, I think "legal" should be "well-defined". I used the two terms interchangeably, but I think "well-defined" is a better name for it.

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileFuzzTest.java
Calvin-L: Indeed! But there is a fair amount of code that cares about the list structure, so I don't think there's much benefit in implementing `SequencedOperations implements FileOperation`, even if it would be very elegant. :)

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileFuzzTest.java
IamfromSpace: Nice, yeah, I think this is a much clearer change!

tlatools/org.lamport.tlatools/test/tlc2/util/BufferedRandomAccessFileTest.java
IamfromSpace: Sounds good, like a i said, no deal breakers for me at this point :)  I think the test updates you made are already above and beyond!

IamfromSpace (approved):

lemmy:
I won't have time to review this PR anytime soon and I don't want to be a bottleneck. Given @IamfromSpace's review, I feel confident about this PR. @Calvin-L, feel free to merge or consider inviting additional reviewers at the TLA+ Community Meeting in two weeks. Reviewing the spec offers a great opportunity for those unfamiliar with or not fond of Java to contribute to the TLA+ tools.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -5,149 +5,190 @@
...
+    private long diskPos;
+
+    /**
+     * Value of the mark.
+     * @see #mark()
+     * @see #reset()
+     */
+    private long mark;
+
+    /** Protects {@link #availBuffs} and {@link #numAvailBuffs}. */
```
lemmy, line 76: `mu` also seems to protect `buff`.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
+        // do anything to the unset parts of buff.
+        //
+        // It is tempting to offer a slightly stronger contract by zero-filling the unset part of "buff":
+        //
+        //     Arrays.fill(this.buff, cnt, BuffSz, (byte)0);
+        //
+        // But do not be tempted!  That call incurs a performance penalty WITHOUT OFFERING A STRONGER CONTRACT!  There
+        // are still cases where we rely on "super" to extend the file---meaning we inherit the weakest possible
+        // contract from RandomAccessFile.  In particular, this happens when extending the file using setLength() or
+        // when the client makes a giant forward seek followed by a write.
```
lemmy, line 322: This comment could imply a possible completeness bug in `DiskFPSet` s.t. the arbitrary data matches a fingerprint that is probed.  This particular bug would vanish if the content of the buffer was zeroed.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
     }
 
-    /* overrides RandomAccessFile.seek(long) */
+    @Override
     public void seek(long pos) throws IOException {
+        this.seeek(pos); // ignore boolean result
```
lemmy, line 327: Can we remove `seeek` given that `seek` delegates to `seeek`?

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,446 @@
...
+\* multiple times.  So instead, this spec treats Inv2 as a precondition for the
+\* methods and verifies that it is indeed restored by `Seek(pos)`, which should
```
lemmy, line 76: `Inv2CanAlwaysBeRestored` suggests that this should mention `FlushBuffer` in addition to `Seek`.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,446 @@
...
+SetLengthCorrect    == [][\A offset \in Offset: SetLength(offset) => RAF!SetLength(offset)]_vars
+SeekEstablishesInv2 == [][\A offset \in Offset: Seek(offset) => Inv2']_vars
+Write1Correct       == [][\A symbol \in Symbols: Write1(symbol) => RAF!Write(SeqToArray(<<symbol>>))]_vars
+Read1Correct        == [][\A symbol \in Symbols: Read1(symbol) => RAF!Read(SeqToArray(<<symbol>>))]_vars
+WriteAtMostCorrect  == [][\A len \in 1..MaxOffset: \A data \in Array(Symbols, len): WriteAtMost(data) => \E written \in 1..len: RAF!Write(ArraySlice(data, 0, written))]_vars
+ReadCorrect         == [][\A len \in 1..MaxOffset: \A data \in Array(Symbols, len): Read(data) => RAF!Read(data)]_vars
+
+\* Inv2 is a precondition for many actions; it should always be possible to
+\* restore Inv2.
+Inv2CanAlwaysBeRestored ==
```
lemmy, line 239: I translated the prose comment into the following property:

```tla
Inv2CanAlwaysBeRestoredPart2 ==
    [][IF dirty THEN FlushBuffer => Inv2' ELSE Seek(curr) => Inv2']_vars
```

, which TLC says is violated by the following trace:

```tla
Finished computing initial states: 1 distinct state generated at 2024-06-03 15:53:42.
Action property Inv2CanAlwaysBeRestored is violated.
The behavior up to this point is:
1: <Initial predicate>
/\ BuffSz = 2
/\ MaxOffset = 4
/\ dirty = FALSE
/\ length = 0
/\ curr = 0
/\ lo = 0
/\ buff = [elems |-> <<ArbitrarySymbol, ArbitrarySymbol>>]
/\ diskPos = 0
/\ file_content = [elems |-> <<>>]
/\ file_pointer = 0
/\ abstract_contents = [elems |-> <<>>]
2: <WriteAtMost([elems |-> <<A, A>>]) line 193, col 5 to line 202, col 60 of module BufferedRandomAccessFile>
/\ BuffSz = 2
/\ MaxOffset = 4
/\ dirty = TRUE
/\ length = 2
/\ curr = 2
/\ lo = 0
/\ buff = [elems |-> <<A, A>>]
/\ diskPos = 0
/\ file_content = [elems |-> <<>>]
/\ file_pointer = 0
/\ abstract_contents = [elems |-> <<A, A>>]
3: <FlushBuffer line 109, col 5 to line 119, col 43 of module BufferedRandomAccessFile>
/\ BuffSz = 2
/\ MaxOffset = 4
/\ dirty = FALSE
/\ length = 2
/\ curr = 2
/\ lo = 0
/\ buff = [elems |-> <<A, A>>]
/\ diskPos = 2
/\ file_content = [elems |-> <<A, A>>]
/\ file_pointer = 2
/\ abstract_contents = [elems |-> <<A, A>>]
```

Revisiting https://github.com/tlaplus/tlaplus/pull/907/files#r1625145343, this property, perhaps, too strong.  This hypothesis is supported because the following holds:

```tla
Inv2CanAlwaysBeRestoredPart2 ==
    [][Seek(curr) => Inv2']_vars
```

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,446 @@
...
+ArraySlice(a, startInclusive, endExclusive) == [elems |-> SubSeq(a.elems, startInclusive + 1, endExclusive)]
+ArrayConcat(a1, a2) == [elems |-> a1.elems \o a2.elems]
+
+\* General contract of the file `write()` call: extend the file with
+\* ArbitrarySymbols if necessary, then overlay some `data_to_write` at the
+\* given offset.
+WriteToFile(file, offset, data_to_write) ==
+    LET file_len == ArrayLen(file) IN
+    LET data_len == ArrayLen(data_to_write) IN
+    LET length == Max(file_len, offset + data_len) IN
```
lemmy, line 399: Nitpick and you probably know this, but I prefer a single LET/IN:

```tla
LET file_len == ArrayLen(file)
    data_len == ArrayLen(data_to_write)
    length == Max(file_len, offset + data_len) IN
```

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,446 @@
...
+    /\ length' = Max(length, curr')
+    /\ UNCHANGED <<lo, diskPos, file_pointer, file_content>>
+
+Read(data) ==
+    LET numReadableWithoutSeeking == Min(lo + BuffSz, length) - curr IN
+    /\ Inv2
+    /\ numReadableWithoutSeeking >= 0
+    /\ LET numToRead == Min(ArrayLen(data), numReadableWithoutSeeking) IN
+       LET buffOff == curr - lo IN
+        /\ data = ArraySlice(buff, buffOff, buffOff + numToRead)
```
lemmy, line 184: Neither `Read` nor `Read1` are enabled if the buffer contains `ArbitrarySymbol`, i.e., the following invariant holds:

```tla
DebugReadsWithArbitrarySymbol ==
    \E i \in 1..Len(buff.elems): buff.elems[i] = ArbitrarySymbol =>
        /\ \A data \in UNION{ Array(Symbols, len): len \in 1..MaxOffset }:
                ~ENABLED Read(data)
        /\ \A symbol \in Symbols:
                ~ENABLED Read1(symbol) 
```

Why this restriction, given that `BufferedRandomAccessFile.java` does not impose such a restriction?

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
```diff
@@ -0,0 +1,446 @@
...
+
+DiskF(i) == \* denoted disk(f)[i] in .java file
+    IF i >= 0 /\ i < ArrayLen(file_content)
+    THEN ArrayGet(file_content, i)
+    ELSE ArbitrarySymbol
+
+BufferedIndexes == lo .. (Min(lo + BuffSz, length) - 1)
+
+Inv1 ==
+    /\ diskPos = file_pointer
```
lemmy, line 69: This differs from `V1` in BufferedRandomAccessFile.java.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
-        // SZ Feb 24, 2009: never read locally
-        // this.closed = true;
-        synchronized (mu) {
-            // grow "availBuffs" array if necessary
-            if (numAvailBuffs >= availBuffs.length) {
-                byte[][] newBuffs = new byte[numAvailBuffs + 10][];
-                System.arraycopy(availBuffs, 0, newBuffs, 0, numAvailBuffs);
-                availBuffs = newBuffs;
+        try {
+            // Assert.check(!this.closed);
```
FedericoPonzi, line 214: currently, the implementation allows read after close. We close the underlying file, give back the buffer to the pool but the buffer is not nullified, and subsequent reads that don't access the file will succeed.

Sample test adopted from the testRead (see last three rows)
```
	@Test
	public void testRead() throws IOException {
		final File tmpFile = File.createTempFile("BufferedRandomAccessFileTest_testRead", ".bin");
		tmpFile.deleteOnExit();
		final java.io.RandomAccessFile raf = new BufferedRandomAccessFile(tmpFile, "rw");
		for (long i = 0L; i < BufferedRandomAccessFile.BuffSz / 8; i++) {
			raf.writeLong(i);
		}

		raf.seek(0);
		for (long i = 0l; i < BufferedRandomAccessFile.BuffSz / 8; i++) {
			assertEquals(i, raf.readLong());
		}
                raf.seek(0); //seek back to zero
		raf.close(); // closed
                assertEquals(0, raf.readLong()); // still works
	}
```
This could be a problem in case the buffer is reassigned to a different BRAF. To reduce the state, it would be sufficient to call "isClosed" on the underlinying RAF. Because there is no isClosed method, getFilePointer could be used for its side effect of failing if the underlying object is closed.

FedericoPonzi (commented):
Very good job! I haven't got to the end yet, I've found a read after close issue and add some nits.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
Calvin-L: Not quite; `availBuffs` and `numAvailBuffs` implement a shared [free list](https://en.wikipedia.org/wiki/Free_list) for allocating buffers. Once this class has allocated a buffer by removing one from `availBuffs`, it does not need to lock anything to access its privately allocated `buff`.

More details in this comment --> https://github.com/tlaplus/tlaplus/pull/907#discussion_r1573045418

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
Calvin-L: This is a fantastic point. I will look into strengthening this property on Monday. Perhaps it can be fully specified using `\cdot` to compose `FillBuffer` and `Seek(curr)`.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
Calvin-L: You are right; I believe that restriction can and should be relaxed.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
+
+    /**
+     * Extends this.seek(long) by return if a page has been read (used for statistics).
+     *
+     * @return true iff data was read from disk
+     */
+    // NOTE: a call to seeek(this.curr) suffices to restore invariant V2 after this.curr changes.
+    //TODO come up with better name for seeek %)
+    public boolean seeek(long pos) throws IOException {
         // Assert.check(!this.closed);
```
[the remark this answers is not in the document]
Calvin-L: Generally yes, but---

I plan to keep this (and some other dead code you called out) for two reasons:
 1. `closed` is mentioned in the class invariants; it is (what I would call) a "ghost variable" that is part of the specification but not part of the implementation. So, its presence has a little bit of value.
 2. Keeping this field reduces the already-quite-large size of this PR diff. We can consider removing it (and other dead code) in a separate maintenance-only PR.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
+        this.seeek(pos); // ignore boolean result
+    }
+
+    /**
+     * Extends this.seek(long) by return if a page has been read (used for statistics).
+     *
+     * @return true iff data was read from disk
+     */
+    // NOTE: a call to seeek(this.curr) suffices to restore invariant V2 after this.curr changes.
+    //TODO come up with better name for seeek %)
```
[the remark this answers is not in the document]
Calvin-L: This TODO is still relevant; a future maintenance-only PR could rename `seeek` to something more meaningful. :)

This TODO was moved or re-indented, which is why it appears in the diff. It might be easier to review this while hiding whitespace-only changes:
Screenshot 2024-06-07 at 1 02 05 PM

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
     public long getFilePointer() {
         // Assert.check(!this.closed);
         return this.curr;
     }
-    
-    /* overrides RandomAccessFile.length() */
+
+    @Override
     public long length() throws IOException {
         // Assert.check(!this.closed);
```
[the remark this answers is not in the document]
Calvin-L: See note above about `closed`.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -5,149 +5,190 @@
...
+       with substantial modifications. */
+
+    /**
+     * True if unflushed bytes exist.
+     * @see #flushBuffer()
+     */
     private boolean dirty;  // true iff unflushed bytes exist
+
     // SZ Feb 24, 2009: never read
     // private boolean closed; // true iff the file is closed
```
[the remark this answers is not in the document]
Calvin-L: See note above about `closed`.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -5,149 +5,190 @@
...
      */
-    public BufferedRandomAccessFile(String name, String mode)
-    throws IOException 
-    {
+    @Deprecated
+    public BufferedRandomAccessFile(String name, String mode) throws IOException {
         // Simon Z. replaced the original:
         //    super(name, mode);
         //    this.init();
         // with this.
```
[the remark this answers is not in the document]
Calvin-L: See [note above about `closed`](https://github.com/tlaplus/tlaplus/pull/907#discussion_r1631637066).

Er, well, this one doesn't refer to `closed` specifically, but I am retaining this dead code to keep the size of this PR down.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -5,149 +5,190 @@
...
-      this.hitEOF = false;
-      this.diskPos = 0L;
+
+    /**
+     * Initialize the private fields of the file so as to make it valid.
+     */
+    private void init() throws IOException {
+        this.dirty = false;
+        // SZ Feb 24, 2009: never read locally
+        // this.closed = false;
```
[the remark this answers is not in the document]
Calvin-L: See [note above about `closed`](https://github.com/tlaplus/tlaplus/pull/907#discussion_r1631637066).

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
lemmy: Let me clarify: Can we replace `seek` with `seeek`? :-)

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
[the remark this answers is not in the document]
lemmy: @FedericoPonzi In addition to what @Calvin-L said, we generally try to keep commits as small as possible by, for example, limiting them to a single concern. This makes future bisecting simpler should it become necessary. For instance, if this PR introduces a regression, diagnosing the issue will be much easier if the changes are minimal.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
-    
-    /* Flush any dirty bytes in the buffer to disk. */
+
+    /**
+     * Flush any dirty bytes in the buffer to disk.  Sets {@link #dirty} to <code>false</code>.
+     *
+     * @return true iff any disk writes were performed
+     */
     private boolean flushBuffer() throws IOException {
         if (this.dirty) {
```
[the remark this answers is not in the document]
Calvin-L: Well, [the original code was structured this way too](https://github.com/tlaplus/tlaplus/blob/2bfa142d595e54487b74a793cfb5578f091f2b82/tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java#L202), so this change wouldn't be a "revert" per se. In this case I would rather preserve the original structure as much as possible to keep this PR diff smaller and more focused.

In general I agree with you; an early return would be a clearer and more canonical way to structure the method, and that's how I would write it if I were writing this method from scratch. But in this case the existing method follows the "single return principle" (or perhaps the "returns should all be close to each other principle"), which I think is a perfectly acceptable alternative to early returns when there aren't too many nested ifs.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -5,149 +5,190 @@
...
-	// current buffer. It then creates a new buffer, which takes proportionally
-	// longer with larger buffer sizes.
-	static final int LogBuffSz = 13; // 8K buffer
+    // Increase 8k buffer to match modern day hardware? No!!! The implementation
+    // discards the whole buffer each time it seeks to a position outside the
+    // current buffer. It then creates a new buffer, which takes proportionally
+    // longer with larger buffer sizes.
+    static final int LogBuffSz = 13; // 8K buffer
     public static final int BuffSz = (1 << LogBuffSz);
     static final int BuffMask = ~(BuffSz - 1);
```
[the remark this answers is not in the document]
Calvin-L: Tricky, but you are right! I will make this change on Monday and also add a comment about what this field is computing. It is not at all obvious why `-BuffSz` is a meaningful thing to compute (or for that matter why `~(BuffSz - 1)` is meaningful either).

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
-            // grow "availBuffs" array if necessary
-            if (numAvailBuffs >= availBuffs.length) {
-                byte[][] newBuffs = new byte[numAvailBuffs + 10][];
-                System.arraycopy(availBuffs, 0, newBuffs, 0, numAvailBuffs);
-                availBuffs = newBuffs;
+        try {
+            // Assert.check(!this.closed);
+            this.flush();
+            // SZ Feb 24, 2009: never read locally
+            // this.closed = true;
```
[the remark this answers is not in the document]
Calvin-L: See [note above about `closed`](https://github.com/tlaplus/tlaplus/pull/907#discussion_r1631637066).

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
+            synchronized (mu) {
+                // grow "availBuffs" array if necessary
```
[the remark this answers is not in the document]
Calvin-L: Yes, or removed entirely (which would be [my personal preference](https://github.com/tlaplus/tlaplus/pull/907#discussion_r1573045418)). I retained this code to keep this PR more focused; [this code already existed](https://github.com/tlaplus/tlaplus/blob/2bfa142d595e54487b74a793cfb5578f091f2b82/tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java#L181).

A future PR might remove or alter this for improved performance or maintainability.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
Calvin-L: I have mixed feelings about this.

> the implementation allows read after close

Yes and no. The implementation does not throw an exception if you do a read after close, so in that sense it "allows" read after close. But, the implementation also doesn't specify behavior if you read after close, so in that sense it "disallows" read after close. In other words, _clients_ are responsible for ensuring that they do not interact with closed files.

In this case, I would peronsally rather push responsibility to the callers---but it's true that the official [RandomAccessFile contract](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/RandomAccessFile.html) _does_ require `IOException` to be thrown for _some_ read and write methods if the file is closed (but weirdly, not all of them). In your specific example there is no specified behavior, and therefore the current behavior ("return an arbitrary long without throwing an exception") is legal. But if you replace the final `readLong()` with `read(byte[])`, it is indeed a demonstration of a bug, since that method requires an exception if the file is closed.

Also, because you called this out I noticed that this `close()` method is not idempotent, and [it must be since this class implements `Closeable`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/AutoCloseable.html). A follow-up call will return `buff` to the `availBuffs` pool a second time, which could be catastrophic.

In any case, I think there is a strong case for reinstating the commented-out `closed` field, which would make it very easy to fix these problems. I will look into doing this on Monday.

Calvin-L: Aside: it is also irksome to me that the official contract for `RandomAccessFile` is to throw `IOException` if the file is closed, when [`IllegalStateException`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/IllegalStateException.html) already exists and was designed for _exactly this purpose_:

> Signals that a method has been invoked at an illegal or inappropriate time.

But, to my surprise, `RandomAccessFile` is _older_ than `IllegalStateException`, so I suppose it makes sense.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
+        System.arraycopy(this.buff, buffOff, b, off, numToRead);
+        this.curr += numToRead;
 
-    public BigInteger readBigInteger(int size) throws IOException {
-      byte[] b = new byte[size];
-      // Assert.check(this.read(b) == size);
-      return new BigInteger(b);
+        // If we read past the buffer, restore our internal invariants
+        if (this.curr >= this.lo + BuffSz) {
+            seeek(this.curr);
```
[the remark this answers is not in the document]
Calvin-L: I am inclined to factor this out into a helper method with a name like `restoreInv2`, since an identical check and `seeek` call appears other places as well. I will make that change soon. (When I do, I will probably keep the call to `seeek` and not `seek`, since it avoids an extra call stack entry.)

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
```diff
@@ -164,357 +205,402 @@ public void invalidateBufferedData() throws IOException {
...
-            int len = (int)(this.curr - this.lo);
-            super.write(this.buff, 0, len); 
-            this.diskPos = this.curr;
+            int len = (int)Math.min(this.length - this.lo, BuffSz);
+            if (len > 0) {
+                assert super.getFilePointer() == diskPos;
+                if (this.diskPos != this.lo) {
+                    super.seek(this.lo);
+                }
+                super.write(this.buff, 0, len);
```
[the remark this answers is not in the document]
Calvin-L: It is weird, but [`write` takes an int as its third argument](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/RandomAccessFile.html#write(byte%5B%5D,int,int)). That is because Java arrays can only contain an `int`'s worth of entries, and the `len` argument specifies how many entries to read from the `buff` array.

In this case `len` is guaranteed to fit in an `int`, since it can't be larger than `BuffSz` and `BuffSz` is an `int`.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
Calvin-L: @FedericoPonzi @lemmy On second thought, I do not plan to address this comment in this PR. There is no _new_ bug here, so I think it makes sense to address this in a separate PR.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
Calvin-L: @lemmy There is an associativity problem in your definition; I think it should be
```tla
DebugReadsWithArbitrarySymbol ==
    (\E i \in 1..Len(buff.elems): buff.elems[i] = ArbitrarySymbol) => \* NOTE parens around \E expression
        /\ ...
```

Using that modified definition shows that `Read` actually is enabled when the read does not cover `ArbitrarySymbol`, even if there are `ArbitrarySymbol`s in the buffer.

Even so, I do think it makes sense to allow reads of `ArbitrarySymbol` in the next-relation.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java
Calvin-L: I did a small audit of `DiskFPSet` and its subclasses when I was working on this and determined that they don't actually read arbitrary data. After `setLength()` they make an effort to fill the entire allocated file with data. But, it's quite possible I missed something.

tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla
lemmy: +1, I also forgot to mention that I would advise against using `\cdot` (gut feeling).

Calvin-L:
Thank you for the feedback, everyone! I believe I've addressed all the comments so far. There are currently four follow-up items to look into after merging:
 - Add an `Arrays` module to the Community Modules
 - Reference this spec from tlaplus/examples once it is merged
 - Verify that `DiskFPSet` and friends are not reading undefined data from the file after extending it with `setLength`
 - Make `close()` idempotent and make reads and writes throw `IOException` for closed files

lemmy (approved):

PR #909, opened 2024-04-13
Make pcal.trans unicode-safe

ahelwer:
This PR does not add Unicode support to PlusCal. It simply ensures that PlusCal code reads & writes files encoding with UTF-8 (done by default across Java's `nio` API) and uses codepoints (represented as `int`s in Java) instead of `char`s when tokenizing the input.

Prior to these changes running `pcal.trans` on a file with Unicode symbols on Windows would result in garbled output.

Added a step to the PR CI which runs `pcal.trans` against all PlusCal files in the tlaplus/examples repo.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -93,9 +93,10 @@ jobs:
           - unicode: true
           - unicode: false
     env:
-      EXAMPLES_DIR: examples
-      SCRIPT_DIR: examples/.github/scripts
-      DEPS_DIR: examples/deps
+      EXAMPLES_DIR: "examples"
+      SCRIPT_DIR: "examples/.github/scripts"
+      DEPS_DIR: "examples/deps"
+      DIST_DIR: "tlatools/org.lamport.tlatools/dist"
     steps:
     - name: Clone tlaplus/tlaplus
       uses: actions/checkout@v4
@@ -146,11 +147,11 @@ jobs:
             "specifications/SpecifyingSystems/RealTime/MCRealTimeHourClock.tla"
           )
         fi
-        python "$SCRIPT_DIR/parse_modules.py"                               \
-          --tools_jar_path tlatools/org.lamport.tlatools/dist/tla2tools.jar \
-          --tlapm_lib_path "$DEPS_DIR/tlapm/library"                        \
-          --community_modules_jar_path "$DEPS_DIR/community/modules.jar"    \
-          --manifest_path "$EXAMPLES_DIR/manifest.json"                     \
+        python "$SCRIPT_DIR/parse_modules.py"                             \
+          --tools_jar_path "$DIST_DIR/tla2tools.jar"                      \
+          --tlapm_lib_path "$DEPS_DIR/tlapm/library"                      \
+          --community_modules_jar_path "$DEPS_DIR/community/modules.jar"  \
+          --manifest_path "$EXAMPLES_DIR/manifest.json"                   \
           --skip "${SKIP[@]}"
     - name: Model-check small tlaplus/examples models
       run: |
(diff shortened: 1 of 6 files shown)


tlatools/org.lamport.tlatools/src/pcal/trans.java
```diff
@@ -1295,8 +1294,7 @@ private static void WriteStringVectorToFile(final List<String> inputVec, String
     * each element of inputVec written on a new line.                      *
     ***********************************************************************/
     {
-    	// TODO use Apache Commons for this
-        try (final BufferedWriter fileW = new BufferedWriter(new FileWriter(fileName))) {
+        try (final BufferedWriter fileW = Files.newBufferedWriter(Path.of(fileName))) {
```
lemmy, line 1297: Please move the changes related to java.nio here and below to a separate commit of this PR, i.e., a commit for the refactoring to java.nio.  If we ever have to bisect because java.nio changes in subtle ways, it will be easier with more fine-grained commits.

tlatools/org.lamport.tlatools/src/pcal/Tokenize.java
```diff
@@ -1486,7 +1483,7 @@ else if (nextChar == '\t')
                     { addNextChar();
                       // state = PROLOG_SPACES;
                     }
-                  else if (Misc.IsLetter(nextChar))  
+                  else if (Character.isLetter(nextChar) || '_' == nextChar)
```
lemmy, line 1486: Why is Character.isLetter better than Misc.IsLetter?  As the latter is used in many places, perhaps it could be refactored to internal delegate to Character.isLetter..

tlatools/org.lamport.tlatools/src/pcal/Tokenize.java
```diff
@@ -1071,9 +1070,7 @@ else if (Misc.IsLetter(nextChar))
                   break;
 
                 case BSBUILT_IN :
-                  if (Misc.IsLetter(nextChar) && (nextChar != '_'))
-                        // added test for /= '_' on 21 Feb 2004
```
lemmy, line 1075: Let's not remove comments, unless they are factually incorrect.

tlatools/org.lamport.tlatools/src/pcal/Tokenize.java
```diff
@@ -927,15 +926,15 @@ public static TLAExpr InnerTokenize(PcalCharReader charReader,
               * optimizations that are local to a particular case clause.  *
               *************************************************************/
               { case START :
-                  if (Misc.IsSpace(nextChar))
+                  if (' ' == nextChar || '\f' == nextChar || '\r' == nextChar)
```
lemmy, line 929: Same comment as below: Misc is also used by tla2tex, and, thus, should only be replaced iff it is replaced everywhere.

tlatools/org.lamport.tlatools/src/pcal/Tokenize.java
```diff
@@ -835,7 +834,7 @@ private static void startNewLine()
       * whenever a \n character is removed from the input stream.          *
       *********************************************************************/
       { vspec.addElement(linev)    ;
-        linev = new Vector(30, 30) ;
+        linev = new Vector<TLAToken>(30, 30) ;
```
lemmy, line 837: Please move the addition of type annotations to a dedicated commit.

.github/workflows/pr.yml
```diff
@@ -93,9 +93,10 @@ jobs:
           - unicode: true
           - unicode: false
     env:
-      EXAMPLES_DIR: examples
-      SCRIPT_DIR: examples/.github/scripts
-      DEPS_DIR: examples/deps
+      EXAMPLES_DIR: "examples"
+      SCRIPT_DIR: "examples/.github/scripts"
+      DEPS_DIR: "examples/deps"
+      DIST_DIR: "tlatools/org.lamport.tlatools/dist"
```
lemmy, line 99: The `DIST_DIR` refactoring should also go into a dedicated commit.

tlatools/org.lamport.tlatools/src/pcal/Tokenize.java
ahelwer: I think I'll just take the approach of overloading Misc.IsLetter function for ints.

ahelwer:
@lemmy split up the commits with `git checkout -p`!

lemmy (approved):

tlatools/org.lamport.tlatools/src/pcal/trans.java
```diff
@@ -1327,27 +1325,18 @@ private static List<String> fileToStringVector(String fileName) throws FileToStr
...
-        } catch (FileNotFoundException e) {
-            /**************************************************************
-            * Input file could not be found.                              *
-            **************************************************************/
+        Path inputFilePath = Path.of(fileName);
+        if (!Files.exists(inputFilePath)) {
             throw new FileToStringVectorException("Input file " + fileName + " not found.");
+        }
+        try {
+            return Files.readAllLines(inputFilePath);
```
Calvin-L, line 1333: Small nit:

Although most programmers consider it uglier, it is generally more correct _and_ more efficient to catch `NoSuchFileException` instead of checking `Files.exists` first. (Ask forgiveness, not permission.) The removed code followed that pattern correctly, although since you switched to NIO you should replace the old and underspecified `FileNotFoundException` with the new and well-defined `NoSuchFileException`.

A bit more detail:
 - When done in two separate steps, the check for existence and the open call do not happen atomically. The filesystem is a large widely-used global shared mutable concurrent variable, and if it is changing while this program runs, then the call to `Files.readAllLines` can end up throwing `NoSuchFileException` despite the existence check. Fortunately on all mainstream filesystems, the check for existence and opening the file are done atomically at open time. (I don't expect this race to be common, but it's easy to handle so we might as well.)
 - When a file is opened for reading, the operating system has to determine whether it exists or not, so checking for existence and then opening essentially doubles the number of filesystem tree descents. Although we don't think about them much, those tree descents can be the dominant cost for reading small files.

tlatools/org.lamport.tlatools/src/tla2tex/Misc.java
```diff
@@ -364,7 +363,20 @@ public static boolean IsLetter(char c)
...
 
-                 
+    /**
+     * IsLetter function implemented for Unicode codepoints.
+     * 
+     * @param c The codepoint to check.
+     * @return Whether the given codepoint is a letter.
+     */
+    public static boolean IsLetter(int c) {
+        if (c >= 128) {
```
Calvin-L, line 373: Do codepoints use all 32 bits of the `int`? (I actually don't know the answer myself.)

If so, then very large ones will overflow, so we might need to revise this check to `c < 0 || c >= 128` to handle that case. (Same for the other `Is*` methods below.)

tlatools/org.lamport.tlatools/src/pcal/trans.java
```diff
@@ -1702,7 +1691,7 @@ else if (option.equals("-version") || (inFile && option.equals("version")))
                     {
                         throw new NumberFormatException();
                     }
-                    int a = new Integer(args[nextArg]).intValue();
+                    int a = Integer.valueOf(args[nextArg]);
```
Calvin-L, line 1694: Nit: this call is wasteful because it allocates an `Integer` and then immediately unboxes it to `int`. It would be faster (and probably clearer) to use `Integer.parseInt` instead.

Calvin-L: Thanks for removing the deprecated call to `new Integer` though 👍

tlatools/org.lamport.tlatools/src/tla2tex/Misc.java
ahelwer: Good catch!

tlatools/org.lamport.tlatools/src/pcal/trans.java
ahelwer: Good point. Changed this and also removed a place where file existence was checked during command line arg parsing.

Issue #929, opened 2024-05-29
Unable to fully select error message


Hi-Angel:
## Description

While reporting https://github.com/tlaplus/tlaplus/issues/928 I found that for some unknown reason tlabox splits translation error messages to multiple widgets, and you can't copy complete message all at once e.g. for purposes of creating a bug or messaging to someone.

Here's a funny screenshot: you can see inside a single error message there are 3 unrelated selections:

![Screenshot_20240529_210337](https://github.com/tlaplus/tlaplus/assets/7935057/b7c9fe10-ea82-4397-bf24-0e2014fff9b2)

## Expected Behavior

You succeed

## Actual Behavior

You only manage to select a single line.

## Steps to Reproduce
1. In a new Spec paste the following code:
    ```haskell
    ------------------------------ MODULE scratch ------------------------------

    (*--algorithm noop
    error
    end algorithm; *)

    =============================================================================
    ```
2. Translate PlusCal algorithm by pressing <kbd>Ctrl</kbd>+<kbd>t</kbd>
3. Once a window with error pops up on the right, try selecting the complete error.

## Your Environment
 - TLC version: 2.15
 - Toolbox version: 1.8.0
 - Operating System: Archlinux

lemmy:
Unfortunatey, we don't have the resource to work on this.  Please provide a  ready-to-merge Pull Request.

Hi-Angel:
@lemmy yeah, it was pointed by `help wanted` tag. But why you closed the issue?

lemmy:
@Hi-Angel This has been the workflow so far to manage the number of open issues (there are still too many).

PR #930, opened 2024-05-29
ISSUE_TEMPLATE: moving "Steps to reproduce" to go before Expected/Actual

Hi-Angel:
Having created some reports to TLA+ it always feels weird to me to see how in the issue template the "Expected behavior" goes *before* "Steps to reproduce".

Take [this report](https://github.com/tlaplus/tlaplus/issues/929) for example: when I read from top to bottom, I see paragraph "Expected behavior" starting with `You succeed` — and I think to myself "Mhm? Succeed in what?". This is confusing. That's because paragraphs "Expected/Actual" describe whatever happens from the "Steps" *(whereas "Description" paragraph usually provides some general information: usecase, why, when; basically creates context for "minimal steps" further below)*. But since "Steps" are located after the "expected result", this breaks reader's flow and creates confusion.

So, how about moving "Steps to reproduce" to go before Expected/Actual?

### DIFF
--- a/.github/ISSUE_TEMPLATE/---bug-report.md
+++ b/.github/ISSUE_TEMPLATE/---bug-report.md
@@ -7,7 +7,7 @@ assignees: ''
 
 ---
 
-<!--- 
+<!---
 Hi!
 Have you tried searching for your issue on the following forums?
 If you have any questions, please ask them there.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/---bug-report.md
```diff
@@ -7,7 +7,7 @@ assignees: ''
 
 ---
 
-<!--- 
+<!---
```
lemmy, line 10: Please limit PRs to only relevant changes in the future.

Hi-Angel: PRs or commits? I can split it to a separate commit, no problem

lemmy: A PR is a collection of commits.

Hi-Angel: Yeah, I understand, I was just making sure you actually meant the PRs. It's just that I find weird creating a whole separate PR just to remove two trailing whitespaces in a single file, but if you feel more comfortable with that, then sure, it's your call.

lemmy: I'm not suggesting creating a separate PR to remove trailing whitespaces. Just leave them in.

Issue #227, opened 2019-01-15
OffHeapDiskFPSet suffers from long initialization time


lemmy:
During TLC startup the default fingerprint set implementation [OffHeapDiskFPSet](https://github.com/tlaplus/tlaplus/blob/master/tlatools/src/tlc2/tool/fp/OffHeapDiskFPSet.java) zeros the allocated memory as part of its [initialization](https://github.com/tlaplus/tlaplus/blob/17c1f1acec9f43423d778dd16c58df158f25ae5a/tlatools/src/tlc2/tool/fp/OffHeapDiskFPSet.java#L171) ([relevant part of the TLA+ spec](https://github.com/tlaplus/tlaplus/blob/2d2d18e308fed4404fa2e5f718e2d194010a66f1/tlatools/src/tlc2/tool/fp/OpenAddressing.tla#L173)).  In cases where several gigabytes of memory are dedicated to TLC, zeroing takes noticeable time.  This is especially annoying when the model is small and by itself only takes a few seconds to check.

Note: Up until [12/2018 (targeting upcoming 1.5.8 release)](https://github.com/tlaplus/tlaplus/commit/f12b10e796f775c10b764b3c0afdde0bcc328620) an issue with a similar effect existed for the collision probability.

Workarounds
------------------

1. Dedicate less memory to TLC for small models.  The TLAToolbox by default dedicates 25% of a machine's memory to TLC.  For modern machines this can result in several GBs of memory being dedicated to the TLC process.  File > Preferences > TLA+ Preferences > TLC Model Checker allows to tweak the percentage of memory for new models.

![screenshot from 2019-01-15 09-35-14](https://user-images.githubusercontent.com/88777/51198384-f4ac0f80-18a8-11e9-8e14-0823f11fcb38.png)

2.  Use a different fingerprint set implementation such as MSBDiskFPSet by setting -Dtlc2.tool.fp.FPSet.impl=tlc2.tool.fp.MSBDiskFPSet (note that for larger models OffHeapDiskFPSet provides the best runtime performance and scalability)

dricketts:
This seems to also affect the trace explorer, so I assume that the trace explorer uses the memory configuration setting from the Model Overview tab. Would it make sense to always use a small amount of memory for the trace explorer, regardless of the configuration for normal model checking?

lemmy:
@dricketts For the Trace Explorer this has already been fixed in the [nightly build](https://nightly.tlapl.us).

[...]

PR #931, opened 2024-06-04
Treat warnings as errors in main compile target

ahelwer:
These changes introduce a separate pre-compilation step to compile files that generate (currently) unfixable warnings. This enables us to enable `-Werror` in the main compilation target.


lemmy:
-1: This adds a lot of complexity to silence genuine warnings.

ahelwer:
The warnings aren't actionable and make it hard to find actual compile errors when they occur, in addition to keeping us from adding `-Werror`.

lemmy:
Rewriting LongArray with Unsafe's replacement API will make the warnings go away.  Silencing the warnings doesn't get us there.

ahelwer:
Certainly, this can be a stopgap until that transition is complete. Is that rewrite something you want to do imminently?

lemmy:
No, and removing the warnings makes it less likely for anybody else to step up and do it.

ahelwer:
Unless this task is so high priority that we must remind people of it every time they compile the tools, I don't see why it can't be better tracked in an issue.

lemmy:
It can also be tracked on a postit note under a doormat, but let's leave it where most people will notice it.

tlatools/org.lamport.tlatools/customBuild.xml
```diff
@@ -143,17 +143,53 @@
...
+= the build fails after upgrading Java, it might be because this =
+= API has been changed.                                          =
+==================================================================
 		</echo>
 		<mkdir dir="${class.dir}" />
+		<!--
+			Separate compilation step for files which must have their warnings
+			uniformly ignored. Java 11 removed the -XDignore.symbol.file compiler
+			parameter that worked in Java 8, so this is the new method. For some
+			reason Java *really, really* does not want you to silence the warnings
```
lemmy, line 161: That's because they *really, really* want you to get rid of your use of Unsafe.

ahelwer:
Well it looks like our very own @pron is involved in the sun-setting of this API: https://openjdk.org/jeps/471

Seems we would have to update to Java 22 at the very least. Possibly Java 23, which is slated for release in September of this year.

lemmy:
Once #835 has been resolved, we will have until ~2030, which is when most provider will stop public updates for Java 21 (https://en.wikipedia.org/wiki/Java_version_history).

lemmy:
Reading https://openjdk.org/jeps/471 more carefully, we have until JDK 26.  Until then, Unsafe will be available but use of it will cause increasingly scary-looking warnings at runtime.

Related: https://github.com/tlaplus/tlaplus/issues/227

tlatools/org.lamport.tlatools/customBuild.xml
```diff
@@ -163,6 +199,11 @@
 			release="${java.release}"
 			encoding="UTF-8"
 		>
+			<!-- Exclude the files compiled separately up above to silence warnings. -->
+			<exclude name="tlc2/tool/fp/LongArray.java" />
+			<exclude name="tla2sany/parser/SimpleCharStream.java" />
```
lemmy, line 204: `SimpleCharStream` appears to be the source of [newly introduced warnings](https://github.com/tlaplus/tlaplus/commit/0f27f0bdb4864ec92654abf49362e565e332b71e).  Why can they be ignored?

ahelwer: It's a JavaCC-generated file so the only way to fix the warnings is to upgrade the version of JavaCC to one that doesn't generate code with warnings

ahelwer:
I asked a question about how to use the new unsafe APIs to write a long array [on StackOverflow](https://stackoverflow.com/q/78587689/2852699) yesterday and got a good answer. Seems like a drop-in replacement as soon as we upgrade to Java 22.

I would like to do some more reading and maybe benchmarking to see whether we can just replace it with a safe array-of-arrays.

ahelwer:
Since that's in quite a while, if I make a pending PR to update LongArray.java to use the modern unsafe APIs will that be sufficient to merge this PR?

lemmy:
I'm still not a fan of this due to the added complexity to the build. @Calvin-L, please weigh in.

tlatools/org.lamport.tlatools/customBuild.xml
```diff
@@ -146,14 +146,50 @@
...
 		<mkdir dir="${class.dir}" />
+		<!--
+			Separate compilation step for files which must have their warnings
+			uniformly ignored. Java 11 removed the -XDignore.symbol.file compiler
+			parameter that worked in Java 8, so this is the new method. For some
+			reason Java *really, really* does not want you to silence the warnings
+			around sun.misc.Unsafe; every method that works when suppressing every
+			other warning does not work here (SuppressWarning annotations,
+			-Xlint:none compiler arg, even -nowarn). So, we are reduced to
+			compiling the files we want to ignore with an actual command line call,
```
Calvin-L, line 165: I am not clear on why this has to be a command line call and not a `<javac>` task.

Calvin-L (commented):
Since this work is already done, and since I assume @ahelwer will help us maintain it in the future, I actually vote to merge this.

For what it's worth, splitting the `javac` step into two tasks is uncommon but not unheard of. It sometimes appears in projects that have generated sources; the generated sources may have to be compiled with different options.

tlatools/org.lamport.tlatools/customBuild.xml
ahelwer: There's no way to silence stderr output from a javac task.

Calvin-L: Ah. My fear is that `javac` comes from different places depending on the task used:
 - `<exec>` uses `javac` from `$PATH`
 - `<javac>` uses (as I understand it) the `javac` from the same JDK that Ant is being run with, unless you specify `build.compiler`, in which case that one is used.

That creates an awkward failure mode where setting `build.compiler` on the command line does not work as intended. Some of the sources will be compiled with a different compiler.

I know the warnings are verbose and mostly unnecessary, but perhaps this setup could be simpler if we just accept that they will be printed?

ahelwer: Ah, fair enough. Okay I just switched it to a separate java precompile step. It still emits the sun.misc.Unsafe warnings but at least we can add `-Werror` to the main compilation step. This precompile step also includes a JavaCC-generated file to match your example.

tlatools/org.lamport.tlatools/customBuild.xml
ahelwer: Per [JEP 471](https://openjdk.org/jeps/471):

>Distinct from deprecation warnings, javac has issued warnings about the use of sun.misc.Unsafe since 2006:

> warning: Unsafe is internal proprietary API and may be removed in a future release

>These warnings will continue to be issued, and **cannot be suppressed**.

ahelwer:
This work has been updated to just use a preliminary call to the <javac> target instead of an <exec> invocation, which greatly simplifies it.

lemmy (approved):

Calvin-L:
@ahelwer The split `<javac>` setup does not quite work as intended:
 - `<javac>` compiles not just the `<include>` classes, but also everything that they reference (recursively). That means the first `<javac>` actually compiles a substantial subset of the TLA+ tools source code. (You can observe this by commenting out the second `<javac>` and the final `<copy>`, running `ant clean compile`, and then examining the contents of the `class/` directory. It will contain _many_ compiled classes.)
 - `<javac>` does not recompile unchanged classes, so the second `<javac>` will ignore anything that was already compiled. In other words, many warnings may be leaking through this infrastructure, because most files are only compiled in the first phase.

I was alerted to this because on my local machine, running `ant -f customBuild.xml clean compile` contains the line
```
    [javac] Compiling 60 source files to [...]/class
```
for the second `<javac>` task. But, the tools have ~10x that number of source files, meaning 90% of them are being compiled by the first `<javac>` without `-Werror`.

ahelwer:
Hmmm, drats. Do you know of any other workarounds here? I wonder whether it's even possible to compile Java files non-recursively or Java integrates the compile/link steps.

Calvin-L:
As far as I can tell, it is not possible to coax javac to compile single classes without also compiling everything they depend on.

So, I think that leaves us with 2 practical options:
 1. Accept that we can't have `-Werror` for now.
 2. Implement our own `-Werror` behavior, e.g. by writing a Python script that checks for unacceptable warnings in javac's output.

I think we will end up reverting this change in either case.

lemmy:
-1 (actually -inf) to introducing (Python) wrapper scripts in our build.

PR #937, opened 2024-06-08
Set up a PlusCal universal syntax corpus test framework

ahelwer:
This doesn't add many tests, but it sets up the bones necessary to add a large number of corpus tests similar to what I did for SANY in #886. If any future work wants to modify the PlusCal parser (or fix a bug in it) this will come in useful.

With the exception of some visibility and type annotation changes these changes only touch test code. This generalizes and reuses many of the SANY test utility classes.


tlatools/org.lamport.tlatools/test/pcal/parser/PcalParserTranslator.java
```diff
@@ -0,0 +1,184 @@
...
+		 * Whether the PlusCal block is a fair algorithm.
+		 */
+		boolean isFair = false;
+		
+		/**
+		 * Finds the start of the PlusCal block.
+		 * 
+		 * @param input Text containing PlusCal block, split into lines.
+		 */
+		public PcalStart(Vector<String> input) {
```
Calvin-L, line 51: Nit: I think here you can use the more general `List` type instead of `Vector`.

(Long term we want to get away from `Vector`, since its operations are all `synchronized`, creating a little bit of needless overhead when the list is not actually shared between threads.)

tlatools/org.lamport.tlatools/test/pcal/parser/PcalParserTranslator.java
```diff
@@ -0,0 +1,184 @@
...
+			final String fairStart = "--fair";
+			final String unfairStart = "--algorithm";
```
Calvin-L, line 53: These constants are already defined in `PcalParams`. That file also lists
```java
public static final String BeginFairAlg2 = "algorithm" ;
```
Is that a possible begin token that this class needs to handle?

tlatools/org.lamport.tlatools/test/pcal/parser/PcalParserTranslator.java
```diff
@@ -0,0 +1,184 @@
...
+					} else {
+						this.line = i;
+						this.column = keywordStart + unfairStart.length();
+						this.isFair = false;
+						break;
+					}
+				} else {
+					this.line = i;
+					this.column = keywordStart + fairStart.length();
+					this.isFair = true;
```
Calvin-L, line 70: Nit: all of these class fields can be `final`. That's good style, even for a test-only class, and especially since these fields are non-private. For code like this that contains loops, you'll need to satisfy Java's final-field-initialization-checks by introducing local variables (`line`, `column`, and `isFair`) and writing those into the object as the last few lines in the constructor:
```
this.line = line;
this.column = column;
this.isFair = isFair;
```

I'm not sure how robust this class needs to be; you may also need to detect the case where no begin token was found.

tlatools/org.lamport.tlatools/test/pcal/parser/PcalParserTranslator.java
```diff
@@ -0,0 +1,184 @@
...
+	 * expressions quite simple here. It could be interesting in the future
+	 * to call into SANY to parse these, so we can analyze the logic that
+	 * determines expression extent.
+	 * 
+	 * @param expr The TLA+ expression, as a sequence of lines of tokens.
+	 * @return The corresponding node in normalized form.
+	 */
+	public static AstNode translate(TLAExpr expr) {
+		Assert.assertEquals(1, expr.tokens.size()); // Lines of tokens
+		Assert.assertEquals(1, expr.tokens.get(0).size()); // Tokens on a given line
```
Calvin-L, line 131: I would mention these restrictions in the docstring.

tlatools/org.lamport.tlatools/test/pcal/parser/PcalParserTranslator.java
```diff
@@ -0,0 +1,184 @@
...
+	/**
+	 * Translates the given TLA+ token. We just hardcode a small set of
+	 * tokens used as expression placeholders in the test corpus.
+	 * 
+	 * @param token The TLA+ token to translate.
+	 * @return The corresponding node in normalized form.
+	 */
+	public static AstNode translate(TLAToken token) {
+		switch (token.string) {
+			case "TRUE": {
```
Calvin-L, line 110: The docstring should list which tokens are in the "small set" this method can translate.

tlatools/org.lamport.tlatools/test/pcal/parser/PcalParserTranslator.java
```diff
@@ -0,0 +1,184 @@
...
+			return translate((AST)node.stmts.get(0));
+		} else if (astNode.getClass().equals(AST.SkipObj.getClass())) {
+			return Kind.PCAL_SKIP.asNode();
+		} else if (astNode.getClass().equals(AST.AssertObj.getClass())) {
+			AST.Assert node = (AST.Assert)astNode;
+			AstNode assertNode = Kind.PCAL_ASSERT.asNode();
+			assertNode.addChild(translate(node.exp));
+			return assertNode;
+		} else {
+			throw new ParseAlgorithmException("Unhandled AST class " + astNode.getClass().toString());
```
Calvin-L, line 181: Nit: `Class.toString()` already includes the word "class":
```
jshell> String.class.toString()
$1 ==> "class java.lang.String"
```
I would replace `toString` with `getCanonicalName` here.

tlatools/org.lamport.tlatools/test/pcal/parser/PcalParserTranslator.java
```diff
@@ -0,0 +1,184 @@
...
+	 * bones are now in place for a comprehensive PlusCal syntax corpus test
+	 * similar to what exists for SANY, but now somebody needs to do the real
+	 * work of translating every syntax node to its universal corpus form.
+	 * 
+	 * @param astNode A node in the PlusCal AST.
+	 * @return The corresponding node in normalized universal corpus form.
+	 * @throws ParseAlgorithmException If a construct could not be translated.
+	 */
+	public static AstNode translate(AST astNode) throws ParseAlgorithmException {
+		if (astNode.getClass().equals(AST.UniprocessObj.getClass())) {
```
Calvin-L, line 149: I respect that this matches how the translator does case analysis on AST nodes, but I don't think there's any reason to use the `AST.*Obj` fields in new code. Just use e.g. `AST.Uniprocess.class` instead.

tlatools/org.lamport.tlatools/test/util/CorpusParser.java
```diff
@@ -346,4 +346,25 @@ public static List<CorpusTestFile> getAndParseCorpusTestFiles(Path toolsRoot) th
...
+	public static Path getToolsRoot() {
+		String baseDir = System.getProperty("basedir");
+		if (null == baseDir) {
+			throw new IllegalArgumentException(
+				"Require basedir Java VM environment variable to be provided"
+				+ " and set to tlatools root."
+			);
+		}
+		
+		return Paths.get(baseDir);
```
Calvin-L, line 368: Use `return Paths.get(CommonTestCase.BASE_DIR)` instead. (You may have to make that field public.)

(Actually, what might be even better is a separate PR that replaces all the basedir stuff in the parse tests with proper Java resources so the tests do not depend on system properties or the current working directory.)

tlatools/org.lamport.tlatools/test/util/AstNode.java
```diff
@@ -284,15 +316,31 @@ public static enum Kind {
 		 */
 		private static final Set<Kind> unused = new HashSet<Kind>();
 		
+		/**
+		 * The subset of nodes that are exclusively used by PlusCal.
+		 */
+		private static final Set<Kind> pcal = new HashSet<Kind>();
```
Calvin-L, line 322: Nit: for `enum` types, [`EnumSet`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/EnumSet.html) is far more memory- and cpu-efficient than `HashSet`. Efficiency isn't really a concern here, but `EnumSet` is still a nice thing to know about. :)

tlatools/org.lamport.tlatools/test/pcal/parser/PcalParserTranslator.java
ahelwer: Good catch, will switch to using those.

The weird expected behavior of the `ParseAlgorithm.getAlgorithm` method is that, if we are parsing a fair algorithm (specified with a boolean parameter), it expects to find the `algorithm` keyword. If we are not parsing a fair algorithm, it expects to start parsing the PlusCal code block directly.

tlatools/org.lamport.tlatools/test/util/AstNode.java
ahelwer: Oooh, very nice API!

tlatools/org.lamport.tlatools/test/util/CorpusParser.java
ahelwer: Made the changes here. I can pull them out if necessary; it would be a bit annoying with rebase conflicts but either works.

tlatools/org.lamport.tlatools/test/pcal/parser/PcalParserTranslator.java
lemmy: Please consider adding a corresponding comment to `ParseAlgorithm.getAlgorithm`.

ahelwer: It is currently documented (in non-javadoc style):

https://github.com/tlaplus/tlaplus/blob/a15241e40c56284e36c451a9b65159ab38990aa1/tlatools/org.lamport.tlatools/src/pcal/ParseAlgorithm.java#L308-L311

tlatools/org.lamport.tlatools/test/util/AstNode.java
```diff
@@ -305,12 +339,27 @@ private Kind(String name) {
...
 		
 		/**
-		 * Get list of unused AST node kinds; useful for identifying gaps in testing.
+		 * Get list of unused AST node kinds specific to TLA+, not PlusCal;
+		 * useful for identifying gaps in testing.
+		 * 
+		 * @return Kinds that were never constructed during parsing.
+		 */
+		public static EnumSet<Kind> getUnusedTlaPlusNodeKinds() {
+			EnumSet<Kind> unused = Kind.unused.clone();
```
Calvin-L, line 348: Nit: Here (and in `getUnusedPlusCalNodeKinds` a few lines lower down), use `new EnumSet<>(Kind.unused)` instead of `clone()`. Clone is convenient but it's riddled with surprises; see e.g. ["Copy Constructor versus Cloning" in this interview](https://www.artima.com/articles/josh-bloch-on-design) with Josh Bloch, one of the original bigwig Java architects. See also #836.

Calvin-L (approved):

lemmy (approved):

PR #936, opened 2024-06-07
Expand CONTRIBUTING.md

ahelwer:
Discussed the broad outline of these changes with @juliaferraioli.

Ref #778 
Ref #935

PR #935, opened 2024-06-07
Expand README.md

ahelwer:
Added the following detail to the README:

- Links to releases and notes on release model
- Basic CLI use instructions
- Links to contributing and developing docs

CC @juliaferraioli

Ref #778

PR #938, opened 2024-06-08
Expand DEVELOPING.md

ahelwer:
CC @juliaferraioli

Ref #778
Ref #935
Ref #936

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -1,75 +1,202 @@
-Development Notes
-=================
+Overview
+--------
+This file summarizes basic development knowledge & practices for this project.
+You will learn the repository structure, how to build & test the software it contains from your command line interface (CLI), how to set up your interactive development environment (IDE), and several other tidbits.
+This project is written entirely in Java.
 
-Firstly this is not complete. If this doesn't have information you need, please
-engage with tlaplus contributors on Github
-and add new knowledge, instructions, insights, or tips back to this document to
-help the next potential contributor
-(or even yourself a few months down the track once you've forgotten everything).
+Repository Layout
+-----------------
+Here is a diagram of the repository layout.
+There are other files and directories beyond these, but these are the most important:
+```
+/
+├── LICENSE                     # The project license
+├── README.md                   # Basic info about the project
+├── CONTRIBUTING.md             # Info on contributing to the project
+├── DEVELOPING.md               # This document
+├── pom.xml                     # The Maven build system definition file
+├── toolbox/                    # All files related to the TLA⁺ Toolbox IDE
+├── tlatools/
+│   └── org.lamport.tlatools/   # All files related to the TLA⁺ Tools
+│       ├── .classpath          # The Eclipse IDE classpath file
+│       ├── .project            # The Eclipse IDE project file
+│       ├── customBuild.xml     # The Ant build system definition file
+│       ├── pom.xml             # A Maven build file wrapping the Ant file
+│       ├── src/                # The TLA⁺ Tools source code dir
+│       ├── test/               # Unit tests for the TLA⁺ Tools
+│       ├── test-model/         # TLA⁺ modules and models for use in tests
+│       ├── lib/                # Vendored dependencies; checked-in jar files
+│       └── javacc/             # Source files for the TLA⁺ parser generator
+└───.github/
+    ├── CODE_OF_CONDUCT.md      # The code of conduct
+    └── workflows/              # GitHub CI workflow definition files
+        ├── pr.yml              # The CI workflow that validates PRs
+        └── main.yml            # Build & publish master branch to pre-release
+```
 
+Build & Test TLA⁺ Tools
+----------------------
+Install the following dependencies to your path:
+ * [Java Development Kit](https://adoptium.net/) version 11+
+ * [Apache Ant](https://ant.apache.org/) version 1.9.8+
 
-Building TLC / `tlatools.jar`
-=============================
-
-While the Toolbox is maintained in [Eclipse](https://github.com/tlaplus/tlaplus/tree/master/general/ide), sometimes changes to TLC do not require building the entire GUI. In such cases, it can be beneficial to compile TLC on the command line. 
-
-While `Maven` is used for the Toolbox (see below), `ant` can be used for the command line. It can be installed with `brew` or `apt` with the package name `ant` (e.g. `brew install ant` and `sudo apt-get install ant`).
-
-Make sure that you're using Java 11+. If you're using a lower version you'll see compilation errors.
-
-You can build TLC from a fresh clone of this repository+:
+Clone this repo & open a shell in its root, then run:
 ```bash
-$ git clone https://github.com/tlaplus/tlaplus.git
-$ cd tlaplus/tlatools/org.lamport.tlatools
-$ ant -f customBuild.xml info clean compile compile-test dist
+cd tlatools/org.lamport.tlatools
+ant -f customBuild.xml info compile compile-test dist # Builds tla2tools.jar
+java -cp dist/tla2tools.jar tlc2.TLC test-model/pcal/Bakery.tla # Runs an example
+ant -f customBuild.xml test # Runs unit tests
 ```
 
-The `tlatools.jar` will be created in the `org.lamport.tlatools/dist` directory.
+The unit tests should all succeed.
+The compiled `tla2tools.jar` will be output to `tlatools/org.lamport.tlatools/dist/tla2tools.jar`, and can be used the same as any `tla2tools.jar` you download from the releases.
 
-Test that the build was successful by using an example model:
+Build & Test Toolbox IDE
+------------------------
+Install the following dependencies to your path:
+ * [Java Development Kit](https://adoptium.net/) version 11+
+ * [Apache Ant](https://ant.apache.org/) version 1.9.8+
+ * [Apache Maven](https://maven.apache.org/) version 3.9.7+
+
+Clone this repo & open a shell in its root, then run:
 ```bash
-$ java -jar dist/tla2tools.jar ./test-model/pcal/Bakery.tla
+mvn install -Dmaven.test.skip=true # Builds Toolbox
+mvn verify # Runs tests
 ```
-To run the unit tests, abbreviate the above `ant` command:
+
+The Toolbox distributables will be output in the `toolbox/org.lamport.tla.toolbox.product.product/target/products` directory.
+
+Developing with the Eclipse IDE
+-------------------------------
+You are free to use your own editor, but this project is largely developed in the [Eclipse](https://www.eclipse.org/) integrated development environment (IDE).
+
+### Developing the TLA⁺ Tools in Eclipse
+
+Developing the TLA⁺ Tools in Eclipse requires the [Eclipse IDE for RCP and RAP Developers](https://www.eclipse.org/downloads/packages/release/2024-06/r/eclipse-ide-rcp-and-rap-developers) edition, or any other edition that comes with the OSGi Plugin Development Extension.
+After you have installed Eclipse, use it open the `tlatools/org.lamport.tlatools` project directory.
+You should then be able to build, run, test, and debug the TLA⁺ Tools.
+
+Note that running build or test operations with Ant can occasionally interfere with Eclipse's build system and cause Eclipse to report nonexistent compilation errors in the UI.
+For this reason you should run Ant from within Eclipse by opening Window > Show View > Other > Ant > Ant, then adding the `customBuild.xml` file and running selected targets.
+This will also add the `RunAllTLCTests` target to the external tools runner menu.
+
+If you ran Ant from your CLI for whatever reason and are encountering issues with nonexistent compilation errors in Eclipse, you can resolve them by right-clicking the project in the Workspace pane then selecting "Refresh".
+This can also be set to refresh automatically by turning on the Eclipse > Settings > General > Workspace > "Refresh using native hooks or polling" setting.
+
+### Developing the Toolbox in Eclipse
+
+For instructions on how to set up Eclipse to develop the Toolbox IDE, read [this](general/ide/README.md).
+
+Using Git Effectively
+---------------------
+
+Git was designed to shine in a federated open source development model.
+You are likely to require use of more of its features than you would in a corporate development environment.
+Here is a brief primer on often-used git functionality when developing for this project.
+It assumes a basic familiarity with common git operations like clone, commit, push, and pull.
+There are countless free and paid git guides available online; [here](https://www.git-scm.com/doc) are the docs from the git project itself, and [here](https://wizardzines.com/git-cheat-sheet.pdf) is a useful cheat-sheet from Julia Evans.
+
+### Dealing with multiple remotes
+
+Your first step when developing for this project will be to fork it so you have your own copy to work with under your own GitHub account.
+Then, you will clone that fork to your local development machine.
+However, you will also want to occasionally synchronize your fork with the original (which we might call *upstream*).
+For this you will need to define multiple *remotes* that you can push to and pull from.
+To list all the remotes defined for your local cloned repo, run:
 ```bash
-$ ant -f customBuild.xml test
+git remote
+```
+If you have a remote named `origin`, you can see the repository it is pointing to with:
+```bash
+git remote get-url origin
+```
+You want to have at least two remotes, one pointing to upstream and another pointing to your user fork.
+Add an `upstream` remote with this command:
+```bash
+git remote add upstream git@github.com:tlaplus/tlaplus.git
+```
+If you cloned your repo from your user fork, you can also rename `origin` to `user` or something similarly descriptive as follows:
+```bash
+git remote rename origin user
+```
+If you want to get the latest changes from upstream, you can then run:
+```bash
+git checkout master
+git pull upstream master
+```
+Then, either merge or rebase on those changes to get them into your local development branches.
+
+### Modifying past commits
+
+When you're working with others, you don't want to subject them to your raw git commit history.
+Often you accumulate commits that are only periodic check-ins to avoid losing work instead of logically-contained sets of changes.
+Git's `rebase` command is much-feared but powerful, and you should become very familiar with it.
+If you are developing a feature in a branch, you can enter rebase mode by running:
+```bash
+git rebase -i master
+```
+This will take you to a text editor where you can command git to do all kinds of things with your commits!
+You can:
+ * Reorder them
+ * Squash several of them into a single commit
+ * Select them to be manually edited/amended
+
+Then, git will run through all your commits sequentially and execute whatever manipulations you indicated.
+Squashing is very useful for combining those periodic check-in commits into something meaningful.
+Editing is useful for responding to PR review comments; most people know about using `git --amend` to apply changes to the latest commit, but using the `rebase` edit option lets you apply changes to commits in the past!
+
+### Splitting up commits and PRs
+
+While rebasing, you can also use editing to split a single commit up into multiple commits.
+This is fairly straightforward when the changes are in different files, but if you want some changes within a single file to go in one commit and some to go in another commit, you can accomplish this interactively with:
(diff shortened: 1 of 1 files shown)


DEVELOPING.md
```diff
@@ -1,75 +1,236 @@
...
+    ├── CODE_OF_CONDUCT.md      # The code of conduct
+    └── workflows/              # GitHub CI workflow definition files
+        ├── pr.yml              # The CI workflow that validates PRs
+        └── main.yml            # Build & publish master branch to pre-release
+```
 
 
-Building TLC / `tlatools.jar`
-=============================
+Build & Test CLI Tools
```
lemmy, line 38: Elsewhere it's called "TLA+ Tools", not "CLI Tools".

DEVELOPING.md
```diff
@@ -1,75 +1,236 @@
...
+----------------------
+Install the following dependencies to your path:
+ * [OpenJDK](https://openjdk.org/) version 11+
+ * [Apache Ant](https://ant.apache.org/) version 1.9.8+
 
-While the Toolbox is maintained in [Eclipse](https://github.com/tlaplus/tlaplus/tree/master/general/ide), sometimes changes to TLC do not require building the entire GUI. In such cases, it can be beneficial to compile TLC on the command line. 
+Clone this repo & open a shell in its root, then run:
+```bash
+cd tlatools/org.lamport.tlatools
+ant -f customBuild.xml compile compile-test dist # Builds tla2tools.jar
```
lemmy, line 47: This need the info task to properly set the version number.

DEVELOPING.md
```diff
@@ -1,75 +1,236 @@
...
-
-Test that the build was successful by using an example model:
+The Toolbox distributables will be output in the `toolbox/org.lamport.tla.toolbox.product.product/target/products` directory.
+
+Developing with the Eclipse IDE
+-------------------------------
+
+You are free to use your own editor, but this project is largely developed in the [Eclipse](https://www.eclipse.org/) integrated development environment (IDE).
+
+To develop the TLA⁺ CLI tools in Eclipse, use Eclipse to open the `tlatools/org.lamport.tlatools` project.
```
lemmy, line 77: "TLA+ Tools"

DEVELOPING.md
```diff
@@ -1,75 +1,236 @@
...
+This Eclipse project configuration lives in the `tlatools/org.lamport.tlatools/.classpath` file.
```
lemmy, line 78: .classpath and .project.

DEVELOPING.md
```diff
@@ -1,75 +1,236 @@
...
-On completion you'll find the toolbox distributables in `org.lamport.tla.toolbox.product.product/target/products/`
+Release Channels
+----------------
+
+Nightly builds of the [Toolbox](https://nightly.tlapl.us/products/) and [tla2tools](https://nightly.tlapl.us/dist/) are found at https://nightly.tlapl.us/products/ and https://nightly.tlapl.us/dist/.
+The Toolbox contains the latest version of tla2tools.jar for command-line usage in its root directory.
+
+It is also possible to configure the Toolbox to [automatically update to nightly (experimental) builds](https://nightly.tlapl.us/doc/update/update-preferences.html).  
+
+Note that it is called nightly for historical reasons, but builds are actually rolling releases triggered by commits.
```
lemmy, line 216: "Rolling releases" not (yet) true.

DEVELOPING.md
```diff
@@ -1,75 +1,236 @@
...
+ </classpath>
+```
+Then, run `git apply classpath.patch`.
+You can right-click the root directory in Eclipse's Project Explorer and click Refresh to get Eclipse to pick up the `.classpath` changes.
+You will need to avoid including these changes in whatever commits you produce.
+One method is to run `git checkout .classpath` to revert the changes, create your commits, then again run `git apply classpath.patch`.
+
+When running unit tests in Eclipse, many of the tests depend on TLA⁺ module and model files in the repository.
+These tests need to know where to look for these files.
+For this reason, you will often need to modify the test run configurations by adding `-Dbasedir=${workspace_loc:tlatools}` to the VM arguments (not the program arguments) in the Arguments tab of the run configurations menu.
```
lemmy, line 111: `-Dbasedir...` was not always necessary.  It creeped with some recent-ish Eclipse IDE upgrade.  It would be good to eventually get rid of it.

DEVELOPING.md
```diff
@@ -1,75 +1,236 @@
...
 
-The `tlatools.jar` will be created in the `org.lamport.tlatools/dist` directory.
-
-Test that the build was successful by using an example model:
+The Toolbox distributables will be output in the `toolbox/org.lamport.tla.toolbox.product.product/target/products` directory.
+
+Developing with the Eclipse IDE
+-------------------------------
+
+You are free to use your own editor, but this project is largely developed in the [Eclipse](https://www.eclipse.org/) integrated development environment (IDE).
```
lemmy, line 75: One can develop the TLA+ Tools in pretty much an IDE, but Eclipse is the way to go when it comes to the Toolbox (anything else will be an uphill battle).  Thus, I suggest that this section is split into a TLA+ Tools and a Toolbox section, with a focus on the TLA+ Tools --- we want more TLA+ Tool contributors.

DEVELOPING.md
```diff
@@ -1,75 +1,236 @@
...
+$ cat /etc/apt/sources.list.d/tlaplus.list
+deb https://nightly.tlapl.us/toolboxUpdate/ ./
+$ curl -fsSL https://tla.msr-inria.inria.fr/jenkins.pub | sudo apt-key add -
 ```
 
-Debugging
-======================
+#### macOS
+
+The Toolbox's nightly builds are also made available as a [cask](https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/tla-plus-toolbox-nightly.rb) through [homebrew versions](https://github.com/Homebrew/homebrew-cask-versions#usage):
```
lemmy, line 230: Consider adding: "The Homebrew community makes the Toolbox's nightly builds..."

FedericoPonzi:
Thanks for doing this! I'm wondering if we could cherry pick some information from this doc: https://github.com/tlaplus/tlaplus/blob/fb41fdb5e0c5f6f5cfb63c684354134bc0a6888b/tlatools/org.lamport.tlatools/docs/CodebaseOverview.md (part of https://github.com/tlaplus/tlaplus/pull/756). Some parts my apply only to their branch, but maybe some information is still relevant.

lemmy:
> Thanks for doing this! I'm wondering if we could cherry pick some information from this doc: https://github.com/tlaplus/tlaplus/blob/fb41fdb5e0c5f6f5cfb63c684354134bc0a6888b/tlatools/org.lamport.tlatools/docs/CodebaseOverview.md (part of #756). Some parts my apply only to their branch, but maybe some information is still relevant.

At first glance, there are many good sections in that document. Would you be willing to carry them over one by one so we can give each a proper review?

FedericoPonzi:
Of course, I can create a follow up pr.

DEVELOPING.md
```diff
@@ -1,75 +1,241 @@
...
+ 	<classpathentry kind="lib" path="lib/jline/jline-terminal-3.25.0.jar"/>
+-	<classpathentry kind="lib" path="lib/jpf.jar"/>
+ 	<classpathentry kind="lib" path="lib/jmh/jmh-core-1.21.jar"/>
+ 	<classpathentry kind="lib" path="lib/jmh/commons-math3-3.2.jar"/>
+ 	<classpathentry kind="lib" path="lib/gson/gson-2.8.6.jar"/>
+ 	<classpathentry kind="lib" path="lib/lsp/org.eclipse.lsp4j.debug_0.21.1.v20230829-0012.jar"/>
+ 	<classpathentry kind="lib" path="lib/lsp/org.eclipse.lsp4j.jsonrpc_0.21.1.v20230829-0012.jar"/>
++	<classpathentry kind="lib" path="lib/junit-4.12.jar"/>
++	<classpathentry kind="lib" path="lib/jpf.jar"/>
++	<classpathentry kind="lib" path="lib/easymock-3.3.1.jar"/>
```
lemmy, line 101: The tools use Easymock 2.4.0.

DEVELOPING.md
```diff
@@ -1,75 +1,241 @@
...
+-------------------------------
+
+You are free to use your own editor, but this project is largely developed in the [Eclipse](https://www.eclipse.org/) integrated development environment (IDE).
+
+### Developing the TLA⁺ Tools in Eclipse
+
+To develop the TLA⁺ Tools in Eclipse, use Eclipse to open the `tlatools/org.lamport.tlatools` project directory.
+Within that directory, the Eclipse project configuration lives in the `.classpath` and `.project` files.
+The `.classpath` file is also consumed by the Toolbox IDE build, so some local modifications are required before Eclipse can fully build & test the project.
+Copy & paste the following into a file called `classpath.patch` within your repo:
```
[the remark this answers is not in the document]
ahelwer: It does not, Eclipse cannot find any of the test dependencies.

DEVELOPING.md
ahelwer: the `tlatools/org.lamport.tlatools/lib` directory only contains `easymock-3.3.1.jar`.

DEVELOPING.md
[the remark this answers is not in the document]
lemmy: With what error message does the Oomph setup fail?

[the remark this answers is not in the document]
ahelwer: I don't use the oomph setup at all for developing the TLA+ tools (not the toolbox). I just open the tlatools/org.lamport.tlatools project.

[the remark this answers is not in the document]
lemmy: We should get there eventually but that's not supported right now. Thus,  it shouldn't be in DEVELOPING.md.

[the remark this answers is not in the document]
ahelwer: I think having people be able to start development by opening a single directory in eclipse is much simpler than running the long and failure-prone oomph process, especially since we want to focus development on the tools.

lemmy (changes requested):
See standalone comments.

ahelwer:
@lemmy how about this: I will try to use the oomph process again to set up the development environment on linux and macOS. If either of them do not work, then we should just include these instructions about how to set up the development environment by opening a directory in eclipse.

DEVELOPING.md
lemmy: With https://github.com/tlaplus/tlaplus/pull/953 merged, I can now import the `org.lamport.tlatools` project into a recent Eclipse IDE, and it compiles successfully out of the box.

DEVELOPING.md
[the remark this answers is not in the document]
lemmy: See https://github.com/tlaplus/tlaplus/pull/938#discussion_r1644990978

DEVELOPING.md
ahelwer: That's great! I can confirm the tools build successfully now. However, the unit tests still do not build in Eclipse without the addition of `lib/junit-4.12.jar` to the `.classpath` file, and they must be placed above `lib/jpf.jar` (due to some class name collisions). The unit tests will also fail to run unless `lib/hamcrest-core-1.3.jar` is included in `.classpath`.

lemmy: Please check if your Eclipse IDE comes with (OSGi) Plugin Development Extension (PDE), because junit and hamcrest are provided by its plugin libraries.

image

ahelwer: Oh excellent, I found that the [Eclipse IDE for RCP and RAP Developers](https://www.eclipse.org/downloads/packages/release/2024-06/r/eclipse-ide-rcp-and-rap-developers) edition comes with that plugin - I installed that one instead of the one I have now (I think just the basic "Eclipse for Java" edition) and now I can build & run the tests! I will add this to the instructions.

DEVELOPING.md
```diff
@@ -1,75 +1,210 @@
...
+You should then be able to build, run, test, and debug the TLA⁺ Tools.
 
-While `Maven` is used for the Toolbox (see below), `ant` can be used for the command line. It can be installed with `brew` or `apt` with the package name `ant` (e.g. `brew install ant` and `sudo apt-get install ant`).
+When running unit tests in Eclipse, many of the tests depend on TLA⁺ module and model files in the repository.
+These tests need to know where to look for these files.
+For this reason, you will often need to modify the test run configurations by adding `-Dbasedir=${workspace_loc:tlatools}` to the VM arguments (not the program arguments) in the Arguments tab of the run configurations menu.
+You should then be able to run & debug JUnit tests in the Eclipse UI.
 
-Make sure that you're using Java 11+. If you're using a lower version you'll see compilation errors.
+Note that running build or test operations with Ant can interfere with Eclipse's build system and cause Eclipse to report nonexistent compilation errors in the UI; right-clicking the project in the Workspace pane then selecting "Refresh" should make these go away.
```
lemmy, line 84: Eclipse packages ant and can run ant directly.  We should encourage users to run ant from inside Eclipse to avoid inconsistencies.

DEVELOPING.md
```diff
@@ -1,75 +1,210 @@
...
+For this, switch to a new branch then use:
 ```bash
-$ java -jar dist/tla2tools.jar ./test-model/pcal/Bakery.tla
+git cherry-pick <commit hash>
 ```
-To run the unit tests, abbreviate the above `ant` command:
+This will add the specified commit to the head of your new branch.
+You can then rebase your original feature branch to remove that commit.
+
+### Developer Certificate of Origin (DCO) signing
```
lemmy, line 163: Let's not call this "signing" but "sign-off", because the former sounds too much like GPG signatures...

/cc @Calvin-L

DEVELOPING.md
```diff
@@ -1,75 +1,210 @@
...
 ```bash
-$ ant -f customBuild.xml test
+git commit -s -am "Commit message here"
 ```
+Don't worry too much about forgetting this; the CI will catch it and GitHub provides a page with simple instructions to retroactively add DCO signing to past commits.
+Eventually adding the `-s` flag will become muscle memory.
 
-If you're using Linux or OSX, you can install `entr`,
-then run the following to watch for and automatically recompile changes:
+Other Useful Tools
```
lemmy, line 174: This overhaul of developing seems like a good opportunity to get rid of promoting entr.

DEVELOPING.md
```diff
@@ -1,75 +1,205 @@
...
+You should then be able to build, run, test, and debug the TLA⁺ Tools.
+
+When running unit tests in Eclipse, many of the tests depend on TLA⁺ module and model files in the repository.
+These tests need to know where to look for these files.
+For this reason, you will often need to modify the test run configurations by adding `-Dbasedir=${workspace_loc:tlatools}` to the VM arguments (not the program arguments) in the Arguments tab of the run configurations menu.
+You should then be able to run & debug JUnit tests in the Eclipse UI.
 
-Test that the build was successful by using an example model:
+Note that running build or test operations with Ant can occasionally interfere with Eclipse's build system and cause Eclipse to report nonexistent compilation errors in the UI.
+For this reason you should run Ant from within Eclipse by opening Window > Show View > Other > Ant > Ant, then adding the `customBuild.xml` file and running selected targets.
```
lemmy, line 85: The "RunAllTLCTests" should show up under the run external tool button in Eclipse automatically.

image

lemmy (approved):
LGTM

DEVELOPING.md
lemmy: Coming from tlatools/org.lamport.tlatools/ide/RunAllTLCTests.launch

DEVELOPING.md
```diff
@@ -1,75 +1,205 @@
...
+However, if you ran Ant from your CLI for whatever reason and are encountering issues with nonexistent compilation errors in Eclipse, you can resolve them by right-clicking the project in the Workspace pane then selecting "Refresh".
```
lemmy, line 86: Who knows why that's not on by default, but it is a good idea to enable "native hooks":

image

DEVELOPING.md
lemmy: Bisecting identified ad454838ad3eaa296aad212e5e1d480b450ead37 as the starting commit after which `-Dbasedir=...` became necessary.

lemmy: @Calvin-L Any idea how to get the best in both/all worlds?

ahelwer: Use the new [Path.resolve API](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#resolve-java.lang.String-) instead of concatenating strings ourselves?

lemmy: Are you sure that that will work in Eclipse, ant, maven, IntelliJ, and elsewhere?

Calvin-L: How about this patch instead?

```diff
diff --git a/tlatools/org.lamport.tlatools/test/tlc2/tool/CommonTestCase.java b/tlatools/org.lamport.tlatools/test/tlc2/tool/CommonTestCase.java
index f24b805fb..452581ee9 100644
--- a/tlatools/org.lamport.tlatools/test/tlc2/tool/CommonTestCase.java
+++ b/tlatools/org.lamport.tlatools/test/tlc2/tool/CommonTestCase.java
@@ -54,7 +54,7 @@ import util.IsolatedTestCaseRunner;
 @RunWith(IsolatedTestCaseRunner.class)
 public abstract class CommonTestCase {
 
-       protected static final String BASE_DIR = System.getProperty("basedir", "");
+       protected static final String BASE_DIR = System.getProperty("basedir", System.getProperty("user.dir", "."));
        protected static final String TEST_MODEL = "test-model" + File.separator;
        public static final String BASE_PATH = System.getProperty("basepath", BASE_DIR + File.separator + TEST_MODEL);
 
```

(Generally it is safe to add extra `/` to a path; `/tmp/x` is the same as `/tmp///////x`, at least on Unix-based systems. The one exception is adding `/` to an empty string; `"" + "/" + "x"` is not the same as `"" + "x"`, since the first is an absolute path and the second is relative to the current working directory.)

Calvin-L: [`user.dir` is the magic system property for the current working directory](https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html)

lemmy: @Calvin-L Looks good.  Tests work without manually setting `basedir` with your patch.

Calvin-L: > Use the new [Path.resolve API](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#resolve-java.lang.String-) instead of concatenating strings ourselves?

[Yes](https://github.com/tlaplus/tlaplus/blob/4613109641676389d97b8df209d6cf4d90d31c1c/tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.java#L184). :) But independently of that, we have to decide what `basedir` even means, and how the tests should behave when it is not provided.

The Most Correct Fix might actually be even more invasive: make `test-model` a resource directory so it is available on the classpath during test execution. Do away entirely with file APIs and use Java's resource resolution mechanism instead. That would make the tests completely independent of the directory in which they're run. That is the most canonical way for Java tests to depend on data files, but it may require some careful tweaks to achieve.

lemmy: May I suggest to push @Calvin-L patch from above and remove the paragraph from DEVELOPING.md in this PR.

More substantial work is likely related to https://github.com/tlaplus/tlaplus/issues/719.

lemmy: https://github.com/tlaplus/tlaplus/pull/975

lemmy: With #975 merged, this paragraph can be removed from DEVELOPING.md.

ahelwer:
@Calvin-L @lemmy I think this and #935 have stuck around & been through enough iterations to be safe to merge, and any desired modifications can be caught in future PRs; what do you think?

lemmy:
This PR can be merge if @Calvin-L doesn't want to review this PR.

Calvin-L (approved):
This seems like a clear improvement to me. (For some reason I thought I already approved this one!)

PR #939, opened 2024-06-10
Fix compiler error introduced by recent refactoring that assumes a newer

lemmy:
version of Easymock than what TLAToolbox.target defines.

Refactoring: https://github.com/tlaplus/tlaplus/commit/4d60a40d599ef4c5fcbef573e96af125a0661028

[Build][TLC][Test]


lemmy:
https://github.com/tlaplus/tlaplus/commit/4d60a40d599ef4c5fcbef573e96af125a0661028

FedericoPonzi (approved):

lemmy:
@ahelwer Which version of Easymock are you using?

lemmy:
> @ahelwer Which version of Easymock are you using?

Never mind, found that you use 3.3.1.  Since the Tools use 2.4.0, we cannot use `newInstance`.

PR #940, opened 2024-06-10
Remove everything AspectJ from the TLA Tools+ and Toolbox projects.

lemmy:
These aspects never provided user-visible functionality. Instead, they offered functionality that was only used during development, and have not been utilized for many years.

[Refactor][TLC]


lemmy:
@Calvin-L, I don't believe it's necessary, but would you like to review it?

Calvin-L (commented):
Overall this looks great, and I won't miss the AspectJ stuff. :)

There are some references to AspectJ left over in some comments:
```
$ wtfi aspectj --in tlatools -x target -x '\.idea' -x out
tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java
 -> line 470: // tools/aspectj/...) might have added synthetic members.
tlatools/org.lamport.tlatools/src/tlc2/tool/coverage/CostModelCreator.java
 -> line 111: ...pect-programming (based on e.g. AspectJ used in other places
 -> line 113: ...CostModelAspect are woven into TLC by AspectJ at the defined
 -> line 120: ... would require TLC to include the AspectJ runtime - and even
 -> line 133: ...ars as if the code generation in AspectJ is not optimized to
 -> line 136: ...od invocation. Future versions of AspectJ might produce more
```

The comment in `SpecProcessor` looks worth keeping. I am less certain about the ones in `CostModelCreator`.

lemmy:
Thanks for the review. I actually prefer to keep the comments in CMC.

Issue #883, opened 2024-02-21
Nondeterminism in diameter reported by TLC


ahelwer:
## Description
TLC occasionally reports varying diameter values with multiple workers. As discovered in this PR: https://github.com/tlaplus/Examples/pull/122

## Expected Behavior
Diameter is expected to be deterministic, per Markus.

## Actual Behavior
TLC will occasionally report varying state graph diameter, although the total & distinct states remain the same.

## Steps to Reproduce
1. Check out the tlaplus/examples repo
2. Download a recent version of tla2tools.jar
4. Repeatedly run the model for `specifications/glowingRaccoon/clean.tla` or `specifications/Paxos/MCVoting.cfg` with -workers greater than 1
5. Observe that the reported diameter varies

## Steps Taken to Fix
N/A

## Possible Fix
N/A

## Your Environment
Has been reproduced across multiple OSs and (recent) versions of TLC.

FedericoPonzi:
Yes I'm currently looking into this one. You can also assign it to me, thanks

lemmy:
Assigned (looks like you had to comment in order for the Github UI to let me do that).

FedericoPonzi:
Issue: When running the model with multiple workers, sometime the maxLevel (or max Depth) is larger then usual.
Some findings first:
1. the states are generated correctly. The same number of states is generated. The state graph is the same (I suspect).
2. It doesn't reproduce with _all_ specifications.
3. I've tried to go as back as 2018 and the bug was already there.

The ModelChecker involved is running a BFS. I wanted to try the DFS one, but apparently it just doesn't work yet with multiple workers (https://github.com/tlaplus/tlaplus/issues/548 seems like a good issue to look into). 

Anyways, what's happening is that when a worker is faster then another, a node is discovered from a longer path. 
Suppose we have this tree:
![Untitled Diagram drawio](https://github.com/tlaplus/tlaplus/assets/2124904/ae302dc0-1ecd-430c-a753-6e6e88115fd0)

A sequential BFS would visit the tree in the following order:
1. visit A
2. visit B
4. visit C
5. visit E
6. visit D

E and D are effectively at depth 2. A concurrent visit with two threads (the prefix is the thread id):
1. 0: Visit A
2. 0: Take B, go to sleep
3. 2: Visit C
4. 2: Visit D
5. 2: Visit E
7. 0: Visit B

[...]

PR #942, opened 2024-06-12
Update documentation about non-deterministic diameter with multiple workers

FedericoPonzi:
Persist the learning from issue #883. Diameter is non-deterministic when the BFS is run with multiple workers.

### DIFF
--- a/general/docs/current-tools.md
+++ b/general/docs/current-tools.md
@@ -30,6 +30,10 @@ ISpec == IM!Spec
 
 TLC cannot handle natural numbers greater than 2<sup>31</sup> - 1.
 
+When running the model checking in BFS mode with multiple workers on a medium-small model, the reported diameter might differ across runs.
+This is a limitation we accept to avoid putting additional synchronization that would limit the scalability. 
+For small models, use a single worker. For large models, the diameter will almost always appear deterministic.
+
 ### Additional Features
 
 #### Enhanced Replacement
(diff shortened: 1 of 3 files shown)


toolbox/org.lamport.tla.toolbox.doc/html/model/tlc-options-page.html
```diff
@@ -38,6 +38,11 @@ <h3>Number of worker threads</h3>
 This parameter specifies the number of separate threads that TLC will spawn to
 perform that computation.&nbsp;  You should not set it to be greater than the number of
 separate processors (cores) on your computer; the Toolbox will warn you if you do.
+When using multiple workers on a medium-small model, the reported <i>Diameter</i> might
+differ across runs.
+This is a limitation we accept to avoid putting additional synchronization that would limit
+the scalability of the visit. For small models, use a single worker. For large models,
```
lemmy, line 44: Perhaps just "that would limit scalability"?

toolbox/org.lamport.tla.toolbox.doc/html/model/tlc-options-page.html
```diff
@@ -96,6 +101,8 @@ <h3><a  name="model-mode">Model-checking mode</a></h3>
 queue, so it does not produce any
  <i>Diameter</i> or <i>Queue size</i>
  statistics.
+Additionally, <em>Depth-first</em> option doesn't currently
```
lemmy, line 104: The confusion typically comes from the fact that TLC does *not* plain depth-first but depth-first iterative deepening (DFID).  We should consistently call it by its full name.

general/docs/current-tools.md
```diff
@@ -30,6 +30,10 @@ ISpec == IM!Spec
 
 TLC cannot handle natural numbers greater than 2<sup>31</sup> - 1.
 
+When running the model checking in BFS mode with multiple workers on a medium-small model, the reported diameter might differ across runs.
+This is a limitation we accept to avoid putting additional synchronization that would limit the scalability of the visit.
```
lemmy, line 34: Perhaps just "that would limit scalability"?

toolbox/org.lamport.tla.toolbox.doc/html/model/tlc-options-page.html
FedericoPonzi: Thanks for the insight! I've updated the paragraph.

lemmy (approved):

Issue #905, opened 2024-04-10
`test-verify` and `test-dist-long` Ant targets both fail


ahelwer:
These don't appear to be checked in any CI so it isn't surprising that they've started failing. Should they be fixed or just removed from the Ant build file? The `benchmark` target still works but I won't be surprised if that starts failing eventually since it isn't being exercised in the CI anymore.

`test-verify` error:
```
ant -f tlatools/org.lamport.tlatools/customBuild.xml test-verify
Buildfile: /home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/customBuild.xml

test-verify:
    [junit] Running tlc2.tool.fp.OffHeapDiskFPSetJPFTest
    [junit] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.093 sec
    [junit] Test tlc2.tool.fp.OffHeapDiskFPSetJPFTest FAILED
    [junit] Running tlc2.tool.queue.StateQueueJPFTest
    [junit] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.098 sec
    [junit] Test tlc2.tool.queue.StateQueueJPFTest FAILED

BUILD SUCCESSFUL
Total time: 0 seconds
```

`test-dist-long` error:
```
ant -f tlatools/org.lamport.tlatools/customBuild.xml test-dist-long
Buildfile: /home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/customBuild.xml

test-dist-long:
    [javac] Compiling 8 source files to /home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/test-class
    [javac] /home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/test-long/tlc2/tool/liveness/MultiThreadedSpecTest.java:47: error: cannot find symbol
    [javac] import tlc2.tool.WorkerMonitor;
    [javac]                 ^
    [javac]   symbol:   class WorkerMonitor
    [javac]   location: package tlc2.tool
    [javac] /home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/test-long/tlc2/tool/liveness/MultiThreadedSpecTest.java:157: error: package WorkerMonitor does not exist
    [javac] 		WorkerMonitor.addThreadListener(new WorkerMonitor.ThreadListener() {
    [javac] 		                                                 ^
[...]

PR #944, opened 2024-06-12
Bump JPF dependency

FedericoPonzi:
Solves #905.
I've build JPF using the latest master (0916082fac, the history seems fairly stable) and replaced the corresponding jars.

I've updated the CI To run `test-verify`, though I would like some feedback on this. I've CTRL^C'd out of it after about 10 minutes:
```
    [junit] Running tlc2.tool.fp.OffHeapDiskFPSetJPFTest
    [junit] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.063 sec
    [junit] Running tlc2.tool.queue.StateQueueJPFTest
```
It seems to take a very long time, I'm not sure if we want to enable it for all commit to master. OTHO there is no point to have it if we don't run it. My proposal would be to use Github actions and schedule a run on master once a week.

For the `test-dist-long` I'm not able to reproduce the issue, I guess because the related class (/home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/test-long/tlc2/tool/liveness/MultiThreadedSpecTest.java) was deleted since then.

```
test-dist-long:
    [javac] Compiling 5 source files to /home/fponzi/dev/tlaplus/tlatools/org.lamport.tlatools/test-class
    [junit] Running tlc2.tool.fp.OffHeapDiskFPSetLongTest
```

It was deleted just 3 days ago in this pr:https://github.com/tlaplus/tlaplus/pull/940


lemmy:
Agreed, there is no point in running JPF on every commit.  Could a CI job detect if the git changeset contains modifications to one of the relevant source files?

FedericoPonzi:
makes sense! I've added a new ci job.

FedericoPonzi:
CI job currently running on my fork: https://github.com/FedericoPonzi/tlaplus/actions/runs/9490624576/job/26154561571

.github/workflows/test-jpf.yml
```diff
@@ -0,0 +1,41 @@
...
+  run-tlc-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: '11'
```
lemmy, line 28: Do we need to upgrade jpf if this runs on Java11?

.github/workflows/test-jpf.yml
```diff
@@ -0,0 +1,41 @@
...
+      ##
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: Run TLC tests
+        run: ant -f tlatools/org.lamport.tlatools/customBuild.xml -Dtest.halt=true compile compile-test test-verify
```
lemmy, line 41: Also call the `info` target to include build information in the test output (if any).

.github/workflows/test-jpf.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: '11'
+
+      ##
+      ## Speed-up build with a cached ~/.m2/repository (300+ MB).
+      ##
+      - uses: actions/cache@v1
```
lemmy, line 33: Maven doesn't seem involved.

.github/workflows/test-jpf.yml
FedericoPonzi: do you want me to update [pr.yml](https://github.com/tlaplus/tlaplus/blob/fa4968f60a0a460da4b7e182330680ce93bd96aa/.github/workflows/pr.yml#L118) and [main.yml](https://github.com/tlaplus/tlaplus/blob/fa4968f60a0a460da4b7e182330680ce93bd96aa/.github/workflows/main.yml#L107) as well?

.github/workflows/test-jpf.yml
FedericoPonzi: Yes, in this pr I'm also adding the latest version of JPF jar

FedericoPonzi:
The build failed with timeout after 6 hours. I've increased it to 12 hours and tried again: https://github.com/FedericoPonzi/tlaplus/actions/runs/9495856195/job/26168993161

lemmy:
> The build failed with timeout after 6 hours. I've increased it to 12 hours and tried again: https://github.com/FedericoPonzi/tlaplus/actions/runs/9495856195/job/26168993161

I'm pretty sure that the verification used to finish significantly faster.  Consider reverting the upgrade to the newer version of JPF.

> should I create a new similar ci action for `test-dist-long`? I assume the -long prefix is because they also take a long time to run.

I don't recall the total runtime. More importantly, these tests were primarily designed to assess the scalability of the OffHeapDiskFPSet across multiple cores. The fact that the tests pass is of secondary importance.

FedericoPonzi:
> I'm pretty sure that the verification used to finish significantly faster. Consider reverting the upgrade to the newer version of JPF.
Good to know. The previous JPF version won't run with java 11, it fails with NoClassDefFoundError as reported in the ticket. I'll take a closer look to understand why it's diverging.

FedericoPonzi:
> I'm pretty sure that the verification used to finish significantly faster.
I've tried checking out the commit that change last the test (3117a6) and tried to run test-verify from there, and I've stopped it after almost 1 hour. 

I've tried using java 9 but tlatools [won't build](https://github.com/FedericoPonzi/tlaplus/actions/runs/9501844983/job/26188389173) with it. 

Running the test with a single thread rather than 3 takes:
```
    [junit] Running tlc2.tool.fp.OffHeapDiskFPSetJPFTest
    [junit] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.62 sec
    [junit] Running tlc2.tool.queue.StateQueueJPFTest
    [junit] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 10.126 sec
```

In the end, it looks like most of the time is due to Thread creation and reaping. Moving it outside the critical path helped reduce the total runtime (with 3 threads):
```
    [junit] Running tlc2.tool.queue.StateQueueJPFTest
    [junit] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 167.969 sec
```
I *think* it leads to an equivalent test. See sample run here: https://github.com/FedericoPonzi/tlaplus/actions/runs/9505905709/job/26201890817 (failed on the info step, but the testing succeeds).

lemmy:
To establish a runtime baseline, you could revert to the point in the Git history where my work with JPF concluded and run that version on Java 9.

FedericoPonzi:
I tried that as well, but my last comment was poorly formatted and the sentence was wrapped in the quote. It stopped it after 1hr of execution

FedericoPonzi:
test-dist-long compiles fine because the class that failed to build was just removed in this pr: https://github.com/tlaplus/tlaplus/pull/940 so this is ready for review.

tlatools/org.lamport.tlatools/test-verify/tlc2/tool/queue/StateQueueJPFTest.java
```diff
@@ -1,29 +1,3 @@
-/*******************************************************************************
- * Copyright (c) 2015 Microsoft Research. All rights reserved.
```
lemmy, line 2: Why did you remove the copyright header?

lemmy:
Have you confirmed that the changed StateQueue test fails if you deliberately introduce a deadlock in SQ?

tlatools/org.lamport.tlatools/test-verify/tlc2/tool/queue/StateQueueJPFTest.java
FedericoPonzi: Sorry, It was removed it by mistake I'll add it back!

FedericoPonzi:
I got this:
```
gov.nasa.jpf.vm.NoUncaughtExceptionsProperty
java.lang.IllegalMonitorStateException: notifyAll() without holding lock
	at java.lang.Object.notifyAll(gov.nasa.jpf.vm.JPF_java_lang_Object)
	at tlc2.tool.queue.StateQueue.finishAll(StateQueue.java:211)
	at tlc2.tool.queue.StateQueueJPFTest$WorkerTask.run(StateQueueJPFTest.java:81)
```
By removing a bunch of synchronized:
```
diff --git a/tlatools/org.lamport.tlatools/src/tlc2/tool/queue/StateQueue.java b/tlatools/org.lamport.tlatools/src/tlc2/tool/queue/StateQueue.java
index e37d9287c..9bca9f9d2 100644
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/queue/StateQueue.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/queue/StateQueue.java
@@ -205,7 +205,7 @@ public abstract class StateQueue implements IStateQueue {
 	/* (non-Javadoc)
 	 * @see tlc2.tool.queue.IStateQueue#finishAll()
 	 */
-	public synchronized void finishAll() {
+	public void finishAll() {
 		this.finish = true;
 		// Notify all other worker threads.
 		this.notifyAll();
@@ -219,10 +219,9 @@ public abstract class StateQueue implements IStateQueue {
 		//
 		// synchronized(this.mu) does not block when the main thread waits on
 		// this.mu in suspend all. We merely have to follow proper thread access.
-		synchronized (this.mu) {
 			// Technically notify() would do.
-			this.mu.notify();
-		}
+        this.mu.notify();
+
 	}
 
 	/* (non-Javadoc)
@@ -230,13 +229,12 @@ public abstract class StateQueue implements IStateQueue {
 	 */
 	public final boolean suspendAll() {
 		boolean needWait = false;
-		synchronized (this) {
 			if (this.finish) {
 				return false;
 			}
 			this.stop = true;
 			needWait = needsWaiting();
-		}
+
 		// Wait for all worker threads to stop.
 		while (needWait) {
 			synchronized (this.mu) {
@@ -304,7 +302,7 @@ public abstract class StateQueue implements IStateQueue {
 	/* (non-Javadoc)
 	 * @see tlc2.tool.queue.IStateQueue#resumeAll()
 	 */
-	public final synchronized void resumeAll() {
+	public final void resumeAll() {
 		this.stop = false;
 		this.notifyAll();
 	}
```

lemmy:
IIRC it should detect this: https://github.com/tlaplus/tlaplus/commit/0818132bbf405d9b8f275664a64f98b022042908

FedericoPonzi:
Nice bug :) 
I've reverted the commit https://github.com/tlaplus/tlaplus/commit/0818132bbf405d9b8f275664a64f98b022042908 
and it's failing with:
```
    [junit] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.922 sec
    [junit] Test tlc2.tool.queue.StateQueueJPFTest FAILED
```

```
gov.nasa.jpf.vm.NotDeadlockedProperty
deadlock encountered:
  thread java.lang.Thread:{id:0,name:main,status:WAITING,priority:5,isDaemon:false,lockCount:0,suspendCount:0}
  thread java.lang.Thread:{id:1,name:Main,status:WAITING,priority:5,isDaemon:false,lockCount:1,suspendCount:0}
  thread java.lang.Thread:{id:2,name:Worker0,status:TERMINATED,priority:5,isDaemon:false,lockCount:0,suspendCount:0}
  thread java.lang.Thread:{id:3,name:Worker1,status:TERMINATED,priority:5,isDaemon:false,lockCount:0,suspendCount:0}
  thread java.lang.Thread:{id:4,name:Worker2,status:TERMINATED,priority:5,isDaemon:false,lockCount:0,suspendCount:0}
```

lemmy (approved):

tlatools/org.lamport.tlatools/test-verify/tlc2/tool/queue/StateQueueJPFTest.java
```diff
@@ -35,76 +34,110 @@
...
-	public static void main(String[] args) {
-		new StateQueueJPFTest().test();
-	}
-	
-	@Test
-	public void test() {
-		if (verifyNoPropertyViolation()) {
-			final StateQueue queue = new DummyStateQueue();
-			final TLCState tlcState = new DummyTLCState();
-			queue.enqueue(tlcState);
```
lemmy, line 47: Dang, the refactor lost populating the queue.  Thus, without the initial enqueue, the queue starts empty. Every worker immediately gets null and calls `finishAll()`, so the test only explores shutdown races. It no longer exercises workers processing queued states concurrently with `suspendAll()`/`resumeAll()`.

PR #948, opened 2024-06-14
Protect `BufferedRandomAccessFile` against double-close and use-after-close

Calvin-L:
This follows up on https://github.com/tlaplus/tlaplus/pull/907#discussion_r1630015466

These commits restore the commented-out `closed` field and use it to ensure that
 1. The actions in `close()` only happen once; invoking `close()` a second time is a no-op.
 2. After calling `close()`, the file cannot be used. This is good defensive programming and also a requirement of the general [`RandomAccessFile` contract](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/RandomAccessFile.html).


FedericoPonzi (approved):
Looking good!

ahelwer (approved):
Nice tests!

PR #950, opened 2024-06-14
Supress SecurityManager deprecation warnings

### LINKED ISSUE #949: `SecurityManager` is deprecated in Java 17 with no alternative API to intercept calls to `System.exit`
The class `test/tlc2/tool/distributed/DistributedTLCTestClass.java` uses `SecurityManager` to intercept calls to `System.exit` so the JUnit test is not prematurely terminated. The `SecurityManager` class was marked as deprecated in Java 17; see [JEP 411](https://openjdk.org/jeps/411). The JEP specifically calls out the use case of intercepting calls to `System.exit`; [this](https://bugs.openjdk.org/browse/JDK-8199704) bug tracks the development of a replacement API, which as of mid 2024 has not seen any progress toward standardization & implementation.

I believe we have three options if we want to upgrade to Java 17 or later:
1. Annotate the usages with `@SuppressWarnings("removal")` to silence them, although this just kicks the can down the road
2. Figure out another way to intercept calls to `System.exit`, for example by injecting an override-able object wrapping `System.exit` for TLC to call instead of it calling `System.exit` directly
3. Spin out the Distributed TLC test workers into new processes instead of threads

ahelwer:
These don't suppress anything in Java 11 but once we upgrade to Java 17+ they will suppress some warnings.

Closes #949

### DIFF
--- a/tlatools/org.lamport.tlatools/test/tlc2/tool/distributed/DistributedTLCTestCase.java
+++ b/tlatools/org.lamport.tlatools/test/tlc2/tool/distributed/DistributedTLCTestCase.java
@@ -44,6 +44,7 @@ public abstract class DistributedTLCTestCase extends CommonTestCase {
 	protected final String[] arguments;
 	protected final int fpSets;
 	
+	@SuppressWarnings("removal")
 	private SecurityManager securityManager;
 	
 	public DistributedTLCTestCase(String spec, String path) {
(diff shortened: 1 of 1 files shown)


lemmy (approved):

lemmy:
This PR introduces four new warnings in Eclipse on Java 11 (`Unnecessary @SuppressWarnings("removal")`).

Screenshot 2024-06-20 at 8 30 06 AM

ahelwer:
That will be the case until we upgrade to Java 17 unfortunately. Those are linter warnings though, of which we have a great many.

lemmy (changes requested):
Postpone until Java 17 or a later version becomes the new baseline (no time frame has been established for this yet).

Issue #946, opened 2024-06-14
North star: all tools should be able to run multiple times independently within a single JVM process


ahelwer:
This is related to #891, which can be seen as a necessary prerequisite. As discussed in the last monthly community meeting:

It would be amazing to be able to run all the tools, especially TLC, several times within the same JVM process. Reasons:

1. Greatly improved development experience for other tools wishing to integrate these as a library
2. Opens door to interesting tests where you can compare TLC's output across multiple runs in a single test without having to use runtime class reloading nonsense
3. Much faster unit testing runs without the overhead of having to start up a new JVM process with every test

This should be considered a long term goal that is unlikely to be achieved within the next couple of years. However, it should be top of mind when adding new functionality or refactoring existing functionality.

lemmy:
What are these other tools that wish to integrate TLC as a library?

ahelwer:
Apalache does, I think. And with the difficulty of writing a TLA+ parser the door should be open for tools to do this. Having a language tooling ecosystem is a good thing. Otherwise why do we care about preserving the public API if nobody consumes it?

ahelwer:
I recall some comment from you that we want to preserve public-facing APIs. However, I could be misremembering. I do not particularly care either way.

lemmy:
Yes, Apalache is built with SANY, but I'm not aware that Apalache suffers from the problem/enhancement discussed in this PR>

konnov:
> Yes, Apalache is built with SANY, but I'm not aware that Apalache suffers from the problem/enhancement discussed in this PR.

Apalache does not use TLC itself, though an ability to offload some computations to TLC could lead to something interesting, similar to concolic execution. The only thing that Apalache is using from `tla2tools.jar` is SANY. Indeed, we ran into filesystem-level race conditions when running SANY in multiple threads. I think the issue is currently solved by changed the temporary directory, or by having a mutex on top of SANY.

lemmy:
The race condition mentioned by @konnov is tracked in https://github.com/tlaplus/tlaplus/issues/688, but is orthogonal to this is PR.  A related issue is https://github.com/tlaplus/tlaplus/issues/719.  The proposed removal of some SANY APIs is at https://github.com/tlaplus/tlaplus/issues/718.

Calvin-L:
> What are these other tools that wish to integrate TLC as a library?

We've had a few users allude to such tools (or the desire to build such tools):
 - #756 ("To allow TLC to be building block in a larger system")
[...]

PR #951, opened 2024-06-15
Replace `sun.misc.Unsafe` with the new Java 22 FFM API

ahelwer:
Making this a draft PR because it requires Java 22+ so we will be sitting on this PR for the next couple years until a Java LTS release is issued. Writing this now because I've gone to the trouble of learning it now.

This was pretty much a complete internal rewrite although the API stays the same. The FFM API also zeroes memory automatically, which is nice.

I maintained the old behavior which was to just never deallocate and leak memory. If we want to deallocate this memory it is actually very simple, we can switch to using [`Arena.ofAuto()`](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/foreign/Arena.html#ofAuto()) instead of `Arena.global()` and keep the allocator as a field in the class. Then whenever the LongArray class is deallocated by the garbage collector the owned allocated memory will also be deallocated.


lemmy:
Garbage collection of the fingerprint set happens at the earliest possible time, which is when the JVM process terminates.

What do you mean by "leak memory"?

ahelwer:
In some happy future where #946 comes to pass it will be a memory leak because there will be no way to deallocate the allocated memory. If we make it an auto allocator attached as a field to the LongArray class, then whenever LongArray is garbage-collected the memory will be deallocated.

tlatools/org.lamport.tlatools/src/tlc2/tool/fp/LongArray.java
```diff
@@ -1,310 +1,209 @@
...
-            b.append(", ");
-        }
-	}
-	
-	public static void main(final String[] args) throws IOException {
-		final long elements = 1L << Integer.valueOf(args[0]);
-		System.out.format("Allocating LongArray with %,d elements.\n", elements);
-		final LongArray longArray = new LongArray(elements);
-		longArray.zeroMemory();
+		final StringBuilder b = new StringBuilder();
```
lemmy, line 194: Why remove the main?

tlatools/org.lamport.tlatools/test/tlc2/tool/distributed/DistributedTLCTestCase.java
```diff
@@ -44,6 +44,7 @@ public abstract class DistributedTLCTestCase extends CommonTestCase {
 	protected final String[] arguments;
 	protected final int fpSets;
 	
+	@SuppressWarnings("removal")
```
lemmy, line 47: Changes to DistributedTLCTestCase should not be part of this PR.

tlatools/org.lamport.tlatools/src/tlc2/tool/fp/LongArray.java
ahelwer: Why not? Nobody uses it, it's dead code.

tlatools/org.lamport.tlatools/src/tlc2/tool/fp/LongArray.java
```diff
@@ -1,310 +1,209 @@
...
-		// expect 8 (Long.SIZE / 8) which is the size of a TLC fingerprint
-		// (see FP64).
-		Assert.check(this.unsafe.addressSize() == (Long.SIZE / 8), EC.GENERAL);
-		baseAddress = this.unsafe.allocateMemory(positions << logAddressSize);
+		assert isSupported() : "LongArray requires a 64-bit system architecture.";
+
+		// This allocates the memory in the global arena, which ensures it will
+		// never be garbage-collected and thus guarantees a memory leak. Allocated
+		// memory is zeroed automatically.
+		this.segment = Arena.global().allocate(this.layout, capacity);
```
lemmy, line 66: https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/foreign/Arena.html#thread-confinement suggests a potential performance penalty.  It should be confirmed that `global` provides comparable performance to Unsafe.

tlatools/org.lamport.tlatools/src/tlc2/tool/fp/LongArray.java
lemmy: Please don't unilaterally decide what is dead code.

ahelwer: Do you use it?

lemmy (changes requested):
See standalone comments.

lemmy:
Other files such as the Toolbox's `org.lamport.tla.toolbox.product.product.product` still reference Java 11.

ahelwer:
I would hope by the time we want to merge this (you said you wanted to wait for a LTS release?) the toolbox won't be part of the repo. If you actually want to merge it imminently instead then I can look at fixing that.

Calvin-L:
Today `LongArrayTest` achieves ~50% coverage of the `LongArray` class. We should make an effort to improve that number before we rewrite the class internals, and any test suite improvements can be merged separately well ahead of this PR.

Screenshot 2024-06-17 at 12 48 07 PM

tlatools/org.lamport.tlatools/src/tlc2/tool/fp/LongArray.java
Calvin-L: Maybe this can be moved to a new class in `test-benchmark/`?

lemmy:
> I would hope by the time we want to merge this (you said you wanted to wait for a LTS release?) the toolbox won't be part of the repo. If you actually want to merge it imminently instead then I can look at fixing that.
minimum requirements are transitively also the Toolbox's minimum requirements.

The location of the source code repo of the TLA+ Tools is irrelevant; the tools will always remain a dependency of the Toolbox. Therefore, the minimum requirements of the Tools also apply transitively to the Toolbox.

lemmy:
`tlatools/org.lamport.tlatools/META-INF/MANIFEST.MF` also still references 11.

lemmy:
> Today `LongArrayTest` achieves ~50% coverage of the `LongArray` class. We should make an effort to improve that number before we rewrite the class internals, and any test suite improvements can be merged separately well ahead of this PR.
> 

In this particular case, it would be worthwhile to add [JMH](https://github.com/openjdk/jmh) tests to establish a performance baseline, like it is done for e.g. [module overrides](https://github.com/tlaplus/tlaplus/tree/master/tlatools/org.lamport.tlatools/test-benchmark).  Fine if that includes converting LongArray#main to a JMH test.

ahelwer:
Seems like JavaSE-22 is not yet a valid OSGi required execution environment name: https://docs.osgi.org/reference/eenames.html

ahelwer:
> Today `LongArrayTest` achieves ~50% coverage of the `LongArray` class. We should make an effort to improve that number before we rewrite the class internals, and any test suite improvements can be merged separately well ahead of this PR.

@Calvin-L It should be noted that LongArraysTest.java (different file from LongArrayTest.java) has quite a few more tests that exercise more of those lesser-exercised functions. Still happy to write more tests though.

Issue #943, opened 2024-06-12
Flaky unit tests


Calvin-L:
## Description
The unit tests sometimes fail in GitHub CI. Two examples:
 - https://github.com/tlaplus/tlaplus/pull/925#issuecomment-2115885840
 - CI failure for a commit that passed PR CI and local tests on my machine: https://github.com/tlaplus/tlaplus/actions/runs/9487536859/attempts/1 --- unfortunately the log does not include the test output :(

I don't know if the examples are related, but they may be pointing at some obscure race condition in TLC or in the test infrastructure.

lemmy:
Can you check if this is limited to Trace* tests?  That's a relatively new set of tests and feature of the test suite.

Calvin-L:
I don't think we have enough data points to say for sure. It's true that both of the failures we've seen so far have been Trace* tests, but that's hardly conclusive.

Calvin-L:
I set up a script to run the tests over and over on my laptop overnight. Here's an example where `tlc2.TraceExpressionSpecSafetySimTest` failed:

```
TLC2 Version 2.18 of Day Month 20??
Running Random Simulation with seed 1 with 1 worker on 8 cores with 7282MB heap and 0MB offheap memory [pid: 6197] (Mac OS X 14.5 aarch64, Azul Systems, Inc. 21.0.2 x86_64).
Parsing file /Users/cloncari/src/tlaplus/tlatools/org.lamport.tlatools/test-model/TESpecTest/TESpecTest.tla
Parsing file /Users/cloncari/src/tlaplus/tlatools/org.lamport.tlatools/class/tla2sany/StandardModules/Naturals.tla
Parsing file /Users/cloncari/src/tlaplus/tlatools/org.lamport.tlatools/class/tla2sany/StandardModules/_TLAPlusDebugger.tla
Semantic processing of module Naturals
Semantic processing of module _TLAPlusDebugger
Semantic processing of module TESpecTest
Starting... (2024-06-13 09:30:58)
[...]

PR #954, opened 2024-06-17
Simple, robust, fail-fast unit test invocation.

lemmy:
Partially reverts 704bd4d190553bc65d307cac862a5c41395c8ff8, because the script `.github/scripts/parse-unit-test-output.py` has been found to be unreliable at identifying failed tests (see https://github.com/tlaplus/tlaplus/pull/902#issuecomment-2174477956).

[Tests][TLC]


.github/workflows/pr.yml
```diff
@@ -24,22 +24,9 @@ jobs:
...
-        echo "Unit test exit code: $ut_exit_code"
-        echo "UT_EXIT_CODE=$ut_exit_code" >> $GITHUB_ENV # Export UT result to CI env var
-    - name: Summarize unit tests
-      run: |
-        echo "Unit test exit code: $UT_EXIT_CODE"
-        python .github/scripts/parse-unit-test-output.py        \
-          test-output.txt                                       \
-          tlatools/org.lamport.tlatools/target/surefire-reports
-        exit $UT_EXIT_CODE
+      run: ant -f tlatools/org.lamport.tlatools/customBuild.xml -Dtest.halt=true test
```
Calvin-L, line 29: When a test fails, is there a way to get its output from the `target/surefire-reports/` directory? `parse-unit-test-output.py` may not be perfect, but (when it works) it does make the test output available in the CI log.

I was sad not to have that output for [one of the recent flaky test failures](https://github.com/tlaplus/tlaplus/actions/runs/9487536859/attempts/1) (#943), and I was considering making `parse-unit-test-output.py` part of the build in `main.yml` so we would have that output in the future.

lemmy: Perhaps, we should set `showoutput` to `yes`.

lemmy: https://ant.apache.org/manual/Tasks/junitreport.html

ahelwer:
This seems unnecessarily aggressive. Here is the fix: #955

.github/workflows/pr.yml
lemmy: @Calvin-L This is what a JunitReport looks like for a test suite run that was aborted after the first test failures:

image

lemmy:
> This seems unnecessarily aggressive. Here is the fix: #955
> 
> I really don't understand why your immediate reaction was to revert this work when you have an active maintainer who can respond to this issue with fixes.

[Parsing Ant's output is a bad idea.](https://github.com/tlaplus/tlaplus/pull/902#issuecomment-2174477956)

.github/workflows/pr.yml
lemmy: This seems like even nicer reporting for the pr.yml: https://github.com/marketplace/actions/junit-report-action#sample-%EF%B8%8F

lemmy: image

ahelwer:
Markus, I want to be very clear here: I told you, verbatim, that I felt you were being unnecessarily unfriendly. Your immediate reaction to me telling you this was to open a PR reverting my work because it had an easily fixed bug in it. Do you believe you are cultivating a friendly, welcoming environment for contributors with this behavior?

ahelwer:
Please reply inline instead of editing your comments retroactively.

lemmy:
> Please reply inline instead of editing your comments retroactively.

All edits are transparent because they are listed under the dropdown next to "edited".

lemmy:
Uploading `TEST-*.xml` and `junit-noframes.html` as workflow artifacts almost achieves feature parity with the old, defunct [Jenkins build](https://github.com/tlaplus/tlaplus/blob/534788509c9501b41ca62e20570f55ebc6aab34c/.jenkins.groovy#L69-L71).

lemmy:
FWIW: I tried to use https://github.com/pavi2410/html-preview-action to render `junit-noframes.html` as a pull request comment, but it did not work.

.github/workflows/pr.yml
lemmy: The individual TEST-*.xml as well as the aggregated TESTS-*.xml can be imported in Eclipse.  I imagine similar support exists in IntelliJ and other IDEs:

Screenshot 2024-06-17 at 9 35 35 PM

lemmy: I believe this can be resolved.

lemmy:
If need be, showing the longest running tests can be done by extracting the time from the aggregated `TESTS-TestSuites.xml` that is generated by junitreport:

```shell
basex -i TESTS-TestSuites.xml 'for $ts in //testsuite return concat($ts/@time, " ", $ts/@name )' | sort -gr
```

Alternatively, it could also be done as a custom style for JUnitReport.

.github/workflows/pr.yml
```diff
@@ -23,23 +23,36 @@ jobs:
...
```
Calvin-L, line 39: Is `-Dtest.halt=true` desirable? We might want to see all failing tests instead of just the first one. Because JUnit runs tests in an arbitrary order, the first failing test is not even deterministic.

.github/workflows/main.yml
```diff
@@ -103,8 +103,33 @@ jobs:
...
+      ## runs the test suite in parallel, aborting the run and failing the workflow upon the first
+      ## test failure (test.halt=true). Regardless of any test failures, a JUnitReport is generated.
+      ## Both the report and the test results (in XML format) are uploaded as workflow artifacts for
+      ## potential local download and further analysis. Additionally, mikepenz/action-junit-report
+      ## publishes the stack traces from the test results. If desired, the test.halt setting can be
+      ## removed to allow the test suite to run to completion.
+      ## (The same sequence of steps is executed by pr.yml)
+      ## 
     - name: Run TLC tests
       run: ant -f tlatools/org.lamport.tlatools/customBuild.xml -Dtest.halt=true compile compile-test test
```
Calvin-L, line 116: Should this line include the `info` target to match the one in `pr.yml`?

.github/workflows/pr.yml
Calvin-L: This applies to the similar line in `main.yml` as well.

Calvin-L (commented):
This makes the CI output very nice indeed. 😃

.github/workflows/main.yml
lemmy: Good point, wasn't there before but has been added in https://github.com/tlaplus/tlaplus/pull/954/commits/608a4b8e5657668d63c9cbe77ad53c71693e4228

.github/workflows/pr.yml
lemmy: I'm not married to it, but I've set `test.halt=true` in `pr.yml` and changed it to `test.halt=false` in `main.yml`. However, one can override `test.halt` per run via a workflow input in `main.yml` (I don't believe inputs work for PR workflows).

https://github.com/tlaplus/tlaplus/pull/954/commits/acbf907949a4d98aa566f1411f5262b96e3a6735

ahelwer:
Here's a first opportunity to see it in action, a flaky unit test was triggered here! https://github.com/tlaplus/tlaplus/actions/runs/9584727990

PR #953, opened 2024-06-17
Remove Easymock dependency from TLA+ Tools (TLC test suite).

lemmy:
The tests have been marked @Ignore for many years anyway.  Reactivate test cases in SymmetryTableauLiveCheckTest when actively working on symmetry reduction under liveness.

Related to
https://github.com/tlaplus/tlaplus/pull/938#discussion_r1633912962

[Tests][TLC]

PR #955, opened 2024-06-17
Fix handling of integral time elapse in UT parse/summary script

ahelwer:
Bug exposed in PR #953

The source of the bug was the expectation that elapsed time would always take the decimal form `123.123 sec`. The test failures in question elapsed in `0 sec` because it was an entirely commented-out file.

This will also print out a warning if test failures were expected but could not be found.


Calvin-L:
Thanks for the quick fix!

I have to ask, though: do you see any use for this script going forward? As far as I can tell, the publish tasks added in #954 produce better output with less maintenance overhead. `parse-unit-test-output.py` did a good job for a brief window, but I think its time might be through. :)

ahelwer:
I am happy for the script to be deleted since it is being replaced with largely equivalent functionality!

PR #958, opened 2024-06-18
Revert -Werror changes

ahelwer:
This reverts the changes introduced in PR #931, as @Calvin-L observed the pre-compile step was pulling in more files than we wanted. Alas we will just have to get rid of the warnings the hard way.


Calvin-L:
Thanks for following up on this!

PR #961, opened 2024-06-20
Running liveness checking with multiple workers can cause unsoundness: TLC fails to report a violation of a property

lemmy:
Small test case that reproduces the issue by forcing a thread schedule.

Part of Github issue #971
https://github.com/tlaplus/tlaplus/issues/971

[Bug][TLC]


lemmy:
Confirmed that the thread schedule also causes TLC as of [tag v1.4.0](https://github.com/tlaplus/tlaplus/commit/9a3879ada2672f229255d6320eefb722d29e94e3) to miss the violation:

```
1. t1 picks x=F from state queue
2. t2 blocks on the empty state queue
3. t1 adds x=T to the state queue (having evaluated Next for x=F)
4. t2 picks x=T from the state queue
5. t2 adds x=T to the behavior graph before t1 adds x=F to the graph
```

lemmy:
Superseded by https://github.com/tlaplus/tlaplus/pull/972

PR #965, opened 2024-06-20
Output unit test TE specs to unique directories

### LINKED ISSUE #943: Flaky unit tests

ahelwer:
Addresses #943


Calvin-L:
With this change I was able to do 50 consecutive runs of the unit test suite without error, and it's still going. (My previous record was 14.)

PR #966, opened 2024-06-21
Use correct keyword.

lemmy:
[Build]


ahelwer:
Does the build/UTs not pass on windows? It would be more valuable to test that IMO. macOS VMs are pretty slow and similar to linux anyway.

PR #968, opened 2024-06-21
Fix grammar sync check in Windows CI

ahelwer:
This step currently fails on Windows because it uses powershell, and also because the line endings are not renormalized.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -21,9 +21,11 @@ jobs:
         distribution: adopt
         java-version: 11
     - name: Check tla+.jj grammar/code sync
+      shell: bash
       run: |
         ant -f tlatools/org.lamport.tlatools/customBuild.xml generate
         git status
+        git add --renormalize .
         diff_count=$(git status --porcelain=v1 2>/dev/null | wc -l)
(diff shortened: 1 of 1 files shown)


lemmy:
This check should be done by the ant build; otherwise, it should be omitted entirely. Considering how rarely the grammar file is modified, it may be best to skip this check altogether.

ahelwer:
@markus this seems like an excellent place to put our email discussion about reversion steps to use, don't you think?

lemmy:
Please keep comments to technical discussions. Thank you!

ahelwer:
This is a technical discussion. The process by which we decide these things is a technical discussion.

lemmy (changes requested):
This check should be performed as part of the Ant build to a) ensure platform independence and b) to be enforced locally (pre-push).  However, given the stability of the language grammar, it can likely be skipped/removed altogether.

ahelwer:
I believe #1013 resolves this issue as the difference was due to git checking out the file with CRLF line endings on windows, which the Ant generate task overwrote. Now with those changes the generate task leaves CRLF line endings unchanged on windows.

Issue #776, opened 2023-01-06
Procedure argument name causes RuntimeException


ben-Draeger:
## Description

When the following PlusCal algorithm is translated to TLA+

```
(* --algorithm Algo {

variables messages = <<>>

procedure sendMatches(matches) {
sendMatches:       messages := Append(messages,
                                         [matches |-> matches]
                                      );
}

fair process (device \in 0..1)
{
    send: call sendMatches(<<>>);
}

}*)

```

and then ModelChecked for the following Temporal Property using TLC via the TLA+ Toolbox

`Property == ~<>(pc[0]="transition")
`

one gets the following Error:

```
TLC threw an unexpected exception.
This was probably caused by an error in the spec or model.
See the User Output or TLC Console for clues to what happened.
The exception was a java.lang.RuntimeException
: Attempted to check equality of the set {} with the value:
<<>>

```
the User Output is empty and the Progress Output is
```
TLC2 Version 2.16 of 31 December 2020 (rev: cdddf55)
Running breadth-first search Model-Checking with fp 20 and seed 8848499862621567840 with 8 workers on 16 cores with 3852MB heap and 8662MB offheap memory [pid: 2620] (Windows 10 10.0 amd64, AdoptOpenJDK 14.0.1 x86_64, OffHeapDiskFPSet, DiskStateQueue).
Starting SANY...
Parsing file C:\safe\Models\SDC3.toolbox\Model_1\MC.tla
Parsing file C:\safe\Models\SDC3.toolbox\Model_1\SDC3.tla
Parsing file C:\safe\install\TLA+\toolbox\plugins\org.lamport.tlatools_1.0.0.202012311918\tla2sany\StandardModules\TLC.tla
Parsing file C:\safe\Models\SDC3.toolbox\Model_1\System_Base.tla
Warning: symbols were renamed.
Parsing file C:\safe\install\TLA+\toolbox\plugins\org.lamport.tlatools_1.0.0.202012311918\tla2sany\StandardModules\Naturals.tla
Parsing file C:\safe\install\TLA+\toolbox\plugins\org.lamport.tlatools_1.0.0.202012311918\tla2sany\StandardModules\Integers.tla
Parsing file C:\safe\install\TLA+\toolbox\plugins\org.lamport.tlatools_1.0.0.202012311918\tla2sany\StandardModules\Sequences.tla
Parsing file C:\safe\Models\SDC3.toolbox\Model_1\SequencesExt.tla
[...]

Issue #899, opened 2024-03-31
SANY should exit with nonzero error code when parsing invalid PlusCal code and `-error-codes` is given


ahelwer:
## Description
Currently SANY will print out the PlusCal parse error but still exit with error code 0, even with the `-error-codes` flag.

## Expected Behavior
I expected SANY to exit with a nonzero error code when encountering an error and given the `-error-codes` flag.

## Actual Behavior
SANY prints out the syntax error but exits with error code 0, indicating success.

## Steps to Reproduce
Parse the following module with SANY using the `-error-codes` flag; it has the plausible PlusCal syntax error `x = x + 2` instead of `x := x + 2`.
```tla
------------------------------ MODULE Test --------------------------------
(****************************************************************************
--algorithm Increase {
  variable x = 0; {
    while (TRUE) {
      x = x + 2
    }
  }
}
****************************************************************************)
\* BEGIN TRANSLATION (chksum(pcal) = "b4b07666" /\ chksum(tla) = "8adfa002")

\* END TRANSLATION 
====
```
SANY will output the following:
```
****** SANY2 Version 2.2 created 08 July 2020

Parsing file /home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/Test.tla

Unrecoverable error:
 -- Expected ":=" but found "}"
    line 7, column 5.

!! WARNING: The PlusCal algorithm and its TLA+ translation in module Test filename since the last translation.
Semantic processing of module Test
```
SANY will not output any PlusCal parse error if the begin/end translation comments are not present, which is perhaps grounds for its own issue.

## Steps Taken to Fix
N/A

## Possible Fix
The relevant logic seems to be the function `validateParseUnit()` in `tla2sany/modanalyzer/SpecObj.java`, roughly line 1022. However the PlusCal parser also doesn't seem to percolate the parse error up to that level, instead swallowing it and printing out the error directly.

## Your Environment
Current tlaplus/tlaplus head commit of master, Linux but looking at the code this should be a cross-platform issue.

PR #976, opened 2024-06-28
Rearrange position of var "pc" in pluscal translations

FedericoPonzi:
to avoid RuntimeException, see #776.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/pcal/PcalTLAGen.java
+++ b/tlatools/org.lamport.tlatools/src/pcal/PcalTLAGen.java
@@ -1830,28 +1830,48 @@ public class PcalTLAGen
         Vector lVarsSource = new Vector();
         Vector gVarsSource = new Vector();
 
+        /*
+         * Consistently put the "pc" variable first.
+         * In this way, tlc will shortcut evaluation on pc
+         * instead of potentially failing on a user defined variable
+         * that contains an incomparable, different type of value across steps.
+         * For example:
+         * vars == << retval, output, pc, stack, x, y, int >>
+         * the comparison would fail if output get assigned a different type:
+         * <<..., 10, ..., "step2", ...>> vs
+         * <<..., "10", ..., "step3", ...>>
+         * by putting pc first:
+         * vars == << pc, retval, output, stack, x, y, int >>
+         * The comparison becomes:
+         * <<"step2", ..., 10, ...>> vs
+         * <<"step3", ..., "10", ...>>
+         * and tlc would stop at the pc and declare the two tuples different.
+         * More info on Github: #776.
+         */
+        if (! ParseAlgorithm.omitPC) {
+            gVars.addElement("pc");
+            /**
+             * For added variables, create a VarDecl with null origin.
+             */
+            AST.VarDecl pcVarDecl = new AST.VarDecl();
+            pcVarDecl.var = "pc";
+            gVarsSource.addElement(pcVarDecl);
+            vars.addElement("pc");
+        }
+
         /*******************************************************************
         * Set gVars to the global variables, including pc and `stack' if   *
         * there are procedures, and add these variables to vars.           *
         *******************************************************************/
-        if (globals != null)
-            for (int i = 0; i < globals.size(); i++)
-            {
+        if (globals != null) {
+            for (int i = 0; i < globals.size(); i++) {
                 AST.VarDecl decl = (AST.VarDecl) globals.elementAt(i);
                 gVars.addElement(decl.var);
                 gVarsSource.addElement(decl);
                 vars.addElement(decl.var);
             }
-        if (! ParseAlgorithm.omitPC) {
-          gVars.addElement("pc");
-          /**
-           * For added variables, create a VarDecl with null origin.
-           */
-          AST.VarDecl pcVarDecl = new AST.VarDecl();
-          pcVarDecl.var = "pc";
-          gVarsSource.addElement(pcVarDecl);
-          vars.addElement("pc");
         }
+
         if (procs != null && procs.size() > 0)
         {
             gVars.addElement("stack");
--- a/tlatools/org.lamport.tlatools/test-model/CallGotoUnlabeledTest.tla
+++ b/tlatools/org.lamport.tlatools/test-model/CallGotoUnlabeledTest.tla
@@ -23,9 +23,9 @@ C:
 
 end algorithm *)
 \* BEGIN TRANSLATION - the hash of the PCal code: PCal-4786c31224fe8555dc7ad50128a92fc6
-VARIABLES x, pc, stack
+VARIABLES pc, x, stack
 
-vars == << x, pc, stack >>
+vars == << pc, x, stack >>
 
 Init == (* Global variables *)
         /\ x = 0
--- a/tlatools/org.lamport.tlatools/test-model/CallGotoUnlabeledTest.tla
+++ b/tlatools/org.lamport.tlatools/test-model/CallGotoUnlabeledTest.tla
@@ -22,7 +22,7 @@ C:
   goto A;
 
 end algorithm *)
-\* BEGIN TRANSLATION - the hash of the PCal code: PCal-4786c31224fe8555dc7ad50128a92fc6
+\* BEGIN TRANSLATION (chksum(pcal) = "1fe457ba" /\ chksum(tla) = "a7227a89")
 VARIABLES pc, x, stack
 
 vars == << pc, x, stack >>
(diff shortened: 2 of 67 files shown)


tlatools/org.lamport.tlatools/test-model/CallGotoUnlabeledTest.tla
```diff
@@ -23,9 +23,9 @@ C:
 
 end algorithm *)
 \* BEGIN TRANSLATION - the hash of the PCal code: PCal-4786c31224fe8555dc7ad50128a92fc6
```
lemmy, line 25: The hashes should change

FedericoPonzi: I've committed my local diff, but shouldn't the hash change when pluscal code is changed? In this case, the tla+ translated code has changed not the input source

lemmy: The test specs/models might be outdated. There should be two hashes: one for PlusCal and one for the generated TLA+.

lemmy: See e.g. https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/src/pcal/Validator.java

lemmy: Retranslating the specifications and removing the old hashes should result in generating two new hashes.

FedericoPonzi: ack, let's sed (ah!) what I can do

lemmy: https://github.com/tlaplus/tlaplus/assets/88777/4919fde8-55c5-4d59-bf7b-044ed1eed178

lemmy: \* BEGIN TRANSLATION (chksum(pcal) = "5fd933e9" /\ chksum(tla) = "bff29291")

FedericoPonzi: I've pushed a new commit. I tried different routes but in the end replacing the "begin translation" and "end translation" strings and setting the cksum as empty string was the easiest way

tlatools/org.lamport.tlatools/src/pcal/PcalTLAGen.java
```diff
@@ -1843,14 +1843,14 @@ private void GenVarsAndDefs(Vector globals, Vector procs, Vector processes, TLAE
                 vars.addElement(decl.var);
             }
         if (! ParseAlgorithm.omitPC) {
```
lemmy, line 1845: Have you considered keeping `addElement` and moving the if block up instead? Additionally, it would be helpful if you could add a comment explaining the rationale behind why we want `pc` to occur first.

lemmy (approved):
LGTM

Calvin-L (approved):

ahelwer:
The filetype filter on the code review certainly came in useful on this one!

ahelwer (approved):

lemmy:
This PR caused the tlaplus/examples CI to fail: https://github.com/tlaplus/Examples/actions/runs/9735586555/job/26864989173

I recommend not including the pcal translator in the tlaplus/examples CI tests because the translator is not part of that repository.

ahelwer:
I think this might have caught a bug of sorts with our PlusCal upgrade strategy. Correct me if I'm wrong, but I believe the logic around whether to re-run PlusCal translation is to always re-run it if no checksum is provided. Otherwise, if a checksum is present:

|        Re-run generation?                  | PlusCal checksum match | PlusCal checksum wrong |
|-----------------------------------------------|------------------------------------|-----------------------------------|
| Generated TLA+ checksum match |             (A) NO                   |           (B) YES                  |
| Generated TLA+ checksum wrong |             (C) NO                   |           (D) ???                  |

Some notes:
1. Case B is the simplest, what we might call the "happy path", when the user has modified the PlusCal code and then re-runs translation to update the corresponding generated TLA+ code.
1. The reason for not running generation in case C is that the user must have hand-modified the generated code, so we should respect that and not overwrite it; I am not sure what happens in case D (probably generation is skipped for similar reasoning?)
2. The reason for not running it in case A is one of simple optimization; since the source & destination checksums match, and *we assume the translator always produces the same output on the same input* (ominous noises play), we do not run generation.

This seems to indicate we should add a PlusCal generator version number that is incorporated into the checksum. Then, whenever we change the way PlusCal generates code (as we did here) this version number should be incremented, and then code generation will be re-run in case A. What do you think about that?

Separately, the reason the tlaplus/examples CI failed is because generation was automatically re-run on all the PlusCal specs that omit checksums:
```
specifications/Bakery-Boulangerie/Bakery.tla
specifications/Bakery-Boulangerie/Boulanger.tla
specifications/DiningPhilosophers/DiningPhilosophers.tla
specifications/KeyValueStore/KVsnap.tla
specifications/LearnProofs/FindHighest.tla
specifications/LeastCircularSubstring/LeastCircularSubstring.tla
specifications/LoopInvariance/BinarySearch.tla
specifications/LoopInvariance/Quicksort.tla
specifications/LoopInvariance/SumSequence.tla
specifications/MisraReachability/ParReach.tla
specifications/MisraReachability/Reachable.tla
specifications/MultiPaxos-SMR/MultiPaxos.tla
specifications/N-Queens/QueensPluscal.tla
specifications/N-Queens/QueensPluscal.toolbox/FourQueens/QueensPluscal.tla
specifications/SlushProtocol/Slush.tla
specifications/TLC/TLCMC.tla
specifications/TeachingConcurrency/Simple.tla
specifications/TeachingConcurrency/SimpleRegular.tla
specifications/chang_roberts/ChangRoberts.tla
specifications/dijkstra-mutex/DijkstraMutex.tla
specifications/dijkstra-mutex/DijkstraMutex.toolbox/LSpec-model/DijkstraMutex.tla
specifications/dijkstra-mutex/DijkstraMutex.toolbox/Safety-4-processors/DijkstraMutex.tla
specifications/echo/Echo.tla
specifications/transaction_commit/2PCwithBTM.tla
```
Then the CI checks to ensure all the generated specs were unchanged, which ensures people don't accidentally check in specs with stale PlusCal code. This was actually the case for a few specs before I added this check, and the stale translations had runtime bugs exposed when I added models for all the specs, so this check isn't coming out of nowhere.

lemmy:
> I think this might have caught a bug of sorts with our PlusCal upgrade strategy. Correct me if I'm wrong, but I believe the logic around whether to re-run PlusCal translation is to always re-run it if no checksum is provided.

I am not aware of any upgrade strategy for PlusCal beyond the [`-version` command-line flag](https://github.com/tlaplus/tlaplus/blob/f869da5800a0443501f1d037bb1405be331d913f/tlatools/org.lamport.tlatools/src/pcal/trans.java#L218-L222).  Furthermore, I believe the translator has no logic to perform any automatic actions; it delegates the decision what to do on checksum inconsistencies to the IDE.  For the Toolbox, it raises a dialog that defers the decision to the user.  The rational being that users may make deliberate changes to the TLA+ or PlusCal *after* running the translator.  For example, one might add debug instructions to the generated TLA+ that one doesn't want the translator to blindly replace:

image

By the way, TLC will warn users about inconsistencies:

image

This PR should have included an increment in the translator's [minor version number](https://github.com/tlaplus/tlaplus/blob/f869da5800a0443501f1d037bb1405be331d913f/tlatools/org.lamport.tlatools/src/pcal/PcalParams.java#L27).  However, I don't think this PR requires an extension of the `-version` flag because it doesn't change the semantics of the generated TLA+.

> which ensures people don't accidentally check in specs with stale PlusCal code.

How can you be certain that any outdated PlusCal code was accidentally checked in? It is completely acceptable for users to modify the generated TLA+ code, and many of the specifications in the list were created before the translator's checksum feature existed. The translator should not be run on these specs until we have manually confirmed that the generated TLA+ code has not been altered. After this confirmation, checksums can be added to those specifications.

ahelwer:
>I believe the translator has no logic to perform any automatic actions

This seems to be true, and I was thinking of this logic to print warnings in SANY (which is probably what the toolbox is picking up in your screenshots):

https://github.com/tlaplus/tlaplus/blob/f869da5800a0443501f1d037bb1405be331d913f/tlatools/org.lamport.tlatools/src/tla2sany/modanalyzer/SpecObj.java#L1021-L1058

>The translator should not be run on these specs until we have manually confirmed that the generated TLA+ code has not been altered. After this confirmation, checksums can be added to those specifications.

So I guess what we want is to ensure the PlusCal specs in the tlaplus/examples repo have valid syntax without actually running code generation on them. Is there any command line flag for the `pcal.trans` tool to do this? Or perhaps this functionality should be added to SANY as suggested in #899.

Secondarily it is worth considering what value we can wring out of the tlaplus/examples repo as regression tests for the PlusCal translator. What do you think?

PR #977, opened 2024-06-28
Locate the the "StandardModules" folder more reliably

Calvin-L:
If a given `path` exists on the classpath in multiple places, `class.getResource(path)` will return an arbitrary one of them.  This behavior creates a potential problem for `SimpleFilenameToStream` when running the TLA+ tests:

 - Main classes and test classes are placed in separate folders, and both folders are on the classpath.
 - The path "tla2sany/" exists in both locations, since there are some test classes in the `tla2sany` package.

As a result, `SimpleFilenameToStream` may decide that the installation base path is actually "test-classes/tla2sany", which does not contain a "StandardModules" folder.

This is not a problem in most cases, since in practice `getResource` returns the _first_ place it found the requested path, and the main classes go on the classpath first.  But that behavior is not guaranteed, and furthermore some IDEs like IntelliJ actually put the test classes first on the classpath.

This patch changes the internals of `SimpleFilenameToStream` to look for "tla2sany/StandardModules" instead of looking for "tla2sany" and assuming the arbitrarily-chosen result has a "StandardModules" folder.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
+++ b/tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
@@ -1,6 +1,6 @@
 // Copyright (c) 2003 Compaq Corporation.  All rights reserved.
 // Portions Copyright (c) 2003 Microsoft Corporation.  All rights reserved.
-// Copyright (c) 2023, Oracle and/or its affiliates.
+// Copyright (c) 2023, 2024, Oracle and/or its affiliates.
 // Last modified on Sat 28 June 2008 at  0:23:49 PST by lamport
 
 package util;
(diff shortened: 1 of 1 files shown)


lemmy:
I have been burned when touching `SimpleFilenameToStream` and its subclasses (OS compatibility, Toolbox usage, distributed TLC, ...).  Have you considered implementing a more defensive approach by introducing a test-specific `FilenameToStream`?

Calvin-L:
Despite the danger, I still think this is a good change. `SimpleFilenameToStreamTest` gives us good coverage of this class, reducing the chance of regressions. In addition:
 - The `installationBasePath` parameter is always equal to `getInstallationBasePath()`, and the only use of `getInstallationBasePath()` is to provide a value for `installationBasePath`.
 - Today all uses of `installationBasePath` have "/StandardModules/" appended; with this PR that will happen in `getInstallationBasePath()` instead.

Calvin-L:
> Have you considered implementing a more defensive approach by introducing a test-specific `FilenameToStream`?

I both love and hate this idea. On one hand it would certainly give us more control over test behavior. But on the other hand, `SimpleFilenameToStream` is _not_ simple, and I think we get some value out of covering it in the tests. I think the best case for a test-specific `FilenameToStream` would be if we could make `SimpleFilenameToStream` an order of magnitude simpler by removing any functionality that only supports the test environment---but I doubt that's possible since a lot of the complexity revolves around loading data out of JAR files, which is still necessary in the `tla2tools.jar` distribution.

Calvin-L:
I'm going to go ahead and merge this. I've been using this branch without incident for a few days, and every test I can perform suggests the behavior is identical (except when "tla2sany" appears on the classpath in multiple places, in which case the behavior is better).

Issue #981, opened 2024-07-02
Upload code-coverage-report in PR and CI


FedericoPonzi:
Followup to #932 , we need to upload the generated code-coverage-report in PR and CI builds.

FedericoPonzi:
We can use: https://github.com/marketplace/actions/jacoco-reporter

FedericoPonzi:
just a quick update:
1. the state of jacoco reporter actions doesn't look very bright, the above extension [failed](https://github.com/tlaplus/tlaplus/actions/runs/9800114266/job/27061509863). It's written in powershell apparently.


2. I tried to use `[madrapps/jacoco-report](https://github.com/Madrapps/jacoco-report)`. This extension will post a message in the pr with code coverage (see for example [here](https://github.com/thsaravana/jacoco-playground/pull/24). There's an option to update the message rather than creating a new one every time so no risk of spamming. It currently does not working in my PR, but I suspect it's because I'm working from a fork of the repo.


3. The test `[tlc2/tool/InliningTest.java](https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/test/tlc2/tool/InliningTest.java)` is not playing well with Jacoco and fails with:
```
    [junit] junit.framework.AssertionFailedError
    [junit]     at tlc2.tool.InliningTest.notIn(InliningTest.java:130)

```

I assume the instrumentation inserted by Jacoco gets in the way. I will have to add another junit block without jacoco and use it to run this test. By the way, I've added a `System.out.println("methodNameMatches.size = " + methodNameMatches.size());` [here](https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/test/tlc2/tool/InliningTest.java#L128) and it always print methodNameMatches.size = 0 when running _without_ Jacoco. Not sure if this is expected or not.

FedericoPonzi:
It is working on my pr to my fork: https://github.com/FedericoPonzi/tlaplus/pull/1#issuecomment-2209602384 it's a bit funny because the comment gets written once from the macos run and another time by the ubuntu run, but at least it works.

lemmy:
> 3\. The test `[tlc2/tool/InliningTest.java](https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/test/tlc2/tool/InliningTest.java)` is not playing well with Jacoco and fails with:
> 
> ```
>     [junit] junit.framework.AssertionFailedError
[...]

PR #982, opened 2024-07-02
Upload code-coverage report to GitHub and add a comment with the results

FedericoPonzi:
Still WIP.


tlatools/org.lamport.tlatools/customBuild.xml
```diff
@@ -467,219 +473,131 @@
...
-			</batchtest>
-			<batchtest fork="yes" todir="${test.reports}">
-				<fileset dir="${test.dir}">
-					<include name="tlc2/tool/UserModuleOverrideTest.java" />
-				</fileset>
-			</batchtest>
-		</junit>
-
-		<!-- remove copied class.dir -->
+        <jacoco:coverage destfile="${code-coverage-seq.exec}"
```
ahelwer, line 477: I suspect this large diff is due to the replacement of tabs with spaces; the customBuild.xml file uses tabs for indentation, could you check that?

.github/workflows/pr.yml
```diff
@@ -53,12 +59,23 @@ jobs:
...
             tlatools/org.lamport.tlatools/target/surefire-reports/junit-noframes.html
+            tlatools/org.lamport.tlatools/target/code-coverage.exec
+            tlatools/org.lamport.tlatools/target/code-coverage-report/report.xml
     - name: Publish unit test results
       uses: mikepenz/action-junit-report@v4
       if: always()
       with:
           check_name: JUnit Test Report on ${{ matrix.operating-system }}
           report_paths: 'tlatools/org.lamport.tlatools/target/surefire-reports/TEST-*.xml'
+    - name: Publish code coverage results
```
ahelwer, line 70: So the difference between uploading and publishing is that uploading makes them available as artifacts, but publishing makes the github actions bot comment on your PR?

ahelwer (changes requested):

tlatools/org.lamport.tlatools/customBuild.xml
FedericoPonzi: thanks I'll take a look and keep diff minimal! I think the diff is weird also due to the fact that I've removed some tests to make the ci complete faster while I'm still working on this

.github/workflows/pr.yml
FedericoPonzi: The upload will make the .exec file with the coverage available to download in the test assets page inside the related zip file (see the bottom of this page: https://github.com/FedericoPonzi/tlaplus/actions/runs/9800330483?pr=1).

The "publish" step will invoke a github action to publish a comment with the code coverage (see comment: https://github.com/FedericoPonzi/tlaplus/pull/1#issuecomment-2209602384). I've added more info to the issue: https://github.com/tlaplus/tlaplus/issues/981#issuecomment-2209598609 and inputs to the discussion are welcome!

.github/workflows/pr.yml
```diff
@@ -5,6 +5,9 @@ jobs:
   tlatools-build-and-test:
     name: TLA+ Tools Build & Test
     runs-on: ${{ matrix.operating-system }}
+    permissions:
+      contents: read
```
lemmy, line 9: Can you please add a short note why these two permissions are needed?

tlatools/org.lamport.tlatools/customBuild.xml
```diff
@@ -1046,6 +964,7 @@
                 </sourcefiles>
             </structure>
             <html destdir="${code-coverage.dir}"/>
+            <xml destfile="${code-coverage.dir}/report.xml" />
```
lemmy, line 967: Consider naming this file something more descriptive, such as `coverage-result.xml`, to make it clear what the file contains if someone downloads the artifact to their local machine.

FedericoPonzi: The xml coverage result is the report supported by the github actions (both the actions I've tried), it won't be uploaded as artifact, but I can rename it regardless. Also, this file is created inside the code-coverage folder.

.github/workflows/pr.yml
```diff
@@ -43,7 +47,9 @@ jobs:
     - name: Generate JUnitReport
       if: always()
       run: ant -f tlatools/org.lamport.tlatools/customBuild.xml test-report
-    - name: Upload raw unit test results
+    - name: Run unit tests on dist and generate coverage report
+      run: ant -f tlatools/org.lamport.tlatools/customBuild.xml test-dist -Dtest.halt=true code-coverage-report
```
ahelwer, line 51: How long does this take to run? Ideally we want the CI to complete within ten minutes.

lemmy: > How long does this take to run? Ideally we want the CI to complete within ten minutes.

Where is this number coming from?

ahelwer: That's how fast the CI was running before two weeks ago. I put in quite a bit of effort to get it there!

lemmy: OK, so "we" was used in the sense of the royal "we."

ahelwer: So do you care about how fast the CI runs or not?

FedericoPonzi: > Edit: I see from the CI run that this takes as long as the original unit tests. I don't think we should run the unit tests twice in a row, which doubles the total CI runtime.

I think there are two improvements that can be done here:
1. Run test-dist using threads. I just tried a test run, it succeeds in less time and the coverage is same
2. Steps cannot run in parallel, but I could create a new job to run test-dist and upload report + publish comment. This should help further reduce the total ci run time.

How does these sound?

ahelwer: Why does it have to run twice? Is it just because of that one problematic test case? We could pull that out into its own `junit` task invocation where we don't track code coverage, right?

FedericoPonzi: Yes, that's an option but I think the priority is on keeping the customBuild file as simple as possible

ahelwer: I guess you can count me as advocating for adding 50 more lines of XML instead of ensuring all future CI runs take twice as long while adding no additional test value (how much extra CO2 will that produce?) - we also already have multiple `junit` task invocations in the `test` target for the non-parallel tests.

FedericoPonzi: As an alternative, we could skip this work altogether. I think keeping an eye on coverage is good but it's not really vital as long as new contributions include tests. The code-coverage-report ant target is already useful by itself, especially if anyone wants to start contributing some tests. I'll resolve this conversation for now. Let's keep discussing in the [ticket](https://github.com/tlaplus/tlaplus/issues/981) to avoid losing good comments in PRs. Thanks for the feedback!

PR #985, opened 2024-07-06
Show the changed variables as part of the transition if `-difftrace` is given and the successor has previously been added to the GraphViz graph.

lemmy:
/cc @heidihoward

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/util/DotStateWriter.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/util/DotStateWriter.java
@@ -251,16 +251,24 @@ public class DotStateWriter extends StateWriter {
 			maintainRanks(state);
 		} else {
 			// Add the transition edge label.
-			final String labelFmtStr = " [label=\"%s%s\",color=\"%s\",fontcolor=\"%s\"]";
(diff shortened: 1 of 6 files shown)


lemmy:
What do others think about this visualization?

Untitled


Regardless, [7e8080a](https://github.com/tlaplus/tlaplus/pull/985/commits/7e8080a158097049afece58301f2bd3f948c1bcf) and [d0b7a6a](https://github.com/tlaplus/tlaplus/pull/985/commits/d0b7a6a2c8fc9169ad03cdcb79c16efa5a0d046d) should be integrated into master.

ahelwer:
Does it duplicate information? I think all you would need on the transition arrows is the set of changed variables rather than their values right?

lemmy:
Imagine two actions, `s -> r` and `t -> r`. The state `s` is defined as `x = 42 /\ y = "a"`, while the state `t` is defined as `x = 23 /\ y = "b"`. The successor state `r` is defined as `x = 42 /\ y = c`. 

If the TLC processes the action `s -> r` first, the label of state `r` will display `y = c` but will omit `x` since it remains unchanged. If the node representing state `t` is distant in the rendered graph, or, -moreover- the value of `x` will also be omitted from `t` because it is unchanged compared to its predecessor, it would be difficult for users to determine the value of `x`.

ahelwer:
Oh, I am talking about the labels on the transitions - not the states themselves.

PR #988, opened 2024-07-12
Add note about Eclipse interference with Ant CLI build

ahelwer:
Ran into some very weird Ant compilation failures in the CLI and found they were caused by the Eclipse "Refresh using native hooks or polling" setting; added a note about this interference to the developing docs.


lemmy (commented):
Is it correct that this happens iff Eclipse is open while ant runs on the command line?  If yes, it's because the ant changes cause Eclipse to recompile.  This should not be an issue if ant is executed from inside Eclipse.

ahelwer:
Yes, this happens if Ant is run from the command line while Eclipse is open. Do you think the wording on the note makes sense?

lemmy:
I recommend mentioning this and highlighting the possibility of running Ant directly from within Eclipse.

lemmy (approved):

Issue #627, opened 2021-05-19
REPL: definition support


cpacejo:
Support definitions in REPL, to aid exploration of TLA by new users.  Requested already in #586; here is a proposed design.

The REPL should maintain a sequence of operator definitions and (optionally named) module instantiations, which, when evaluating an expression, are emitted in order as part of the `tlarepl` module.  (Supporting module instantiations also allows the REPL to be used to explore existing constant-level modules.)

New definitions and instantiations are input by the user using the same syntax as in a TLA module.  If the definition or instantiation results in a parse error, it should be rejected with the error shared to the user.

Since the REPL now maintains state, it is helpful for users to see and record that state.  A new `LIST` command should display (in order) all existing definitions and instantiations.  A new `SAVE <filename>` command should write out a module containing the definitions and instantiations to a file.  (Probably `<filename>` can be either a proper filename, or just a bare module name.)

Additionally, it should be possible to modify or remove existing definitions.  Prefixing a definition or instantiation with the word `REDEFINE` should replace the named definition or instantiation of the named module from the sequence with the new definition or instantiation, without changing the relative position of the item in the sequence.  As with new definitions, if the replacement results in an error, it should be rejected.

Finally, it should be possible to remove existing definitions.  The new commands `FORGET DEFINITION <name>` and `FORGET INSTANCE <name>` should remove the named operator or instance from the sequence, again rejecting the operation if it results in an invalid module.  (Is `OPERATOR` a better word to use here than `DEFINITION`?)  To permit easily returning to a clean slate, `FORGET ALL` simply wipes all recorded operators and instances.

Optional: Leverage TLC to detect incomplete operator definitions, and prompt the user to complete the definition on one or more subsequent lines.  This is similar behavior to Python's REPL, and especially helps when pasting a multiline definition.

lemmy:
Thanks for opening the issue!

Before anything, I'd suggest to first implement proper error-reporting, which is *the* shortcoming of the current REPL to me.  That said, one can already define operators and instantiate modules in the REPl via LET/IN:

```tla
Welcome to the TLA+ REPL!
TLC2 Version 2.16 of Day Month 20??
Enter a constant-level TLA+ expression.
(tla+) LET Range(T) == { T[x] : x \in DOMAIN T } IN Range([i \in 1..10 |-> i*i])
{1, 4, 9, 16, 25, 36, 49, 64, 81, 100}
```

```tla
Welcome to the TLA+ REPL!
[...]

PR #1000, opened 2024-08-20
Add a `:help` command to the REPL

Calvin-L:
Originally requested in a comment:

https://github.com/tlaplus/tlaplus/issues/627#issuecomment-844538528

I chose the `:` prefix for `:help` because

 - it is already used in at least one other REPL (ghci)
 - it cannot appear at the start of any legal TLA+ expression.

This commit also lays out a bit of infrastructure for adding additional commands.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/REPL.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/REPL.java
@@ -27,6 +27,9 @@ import tla2sany.semantic.ModuleNode;
 import tla2sany.semantic.OpDefNode;
 import tlc2.output.EC;
 import tlc2.output.MP;
+import tlc2.repl.Command;
+import tlc2.repl.CommandParseException;
+import tlc2.repl.EvaluateExpressionCommand;
 import tlc2.tool.EvalException;
 import tlc2.tool.impl.FastTool;
 import tlc2.tool.impl.Tool;
@@ -187,15 +190,13 @@ public class REPL {
      */
     public void runREPL(final LineReader reader) throws IOException {
         // Run the loop.
-    	String expr;
         while (true) {
             try {
-                expr = reader.readLine(prompt);
-                String res = processInput(expr);
-                if (res.equals("")) {
-                    continue;
-                }
-                System.out.println(res);
+                String line = reader.readLine(prompt);
+                Command command = Command.parse(line);
+                command.apply(this, System.out);
+            } catch (CommandParseException e) {
+                System.err.println("Failed to parse command: " + e.getMessage());
             } catch (UserInterruptException e) {
                 return;
             } catch (EndOfFileException e) {
@@ -218,10 +219,7 @@ public class REPL {
             // TODO: Allow external spec file to be loaded into REPL context.
  
             if(args.length == 1) {
-                String res = repl.processInput(args[0]);
-                if (!res.equals("")) {
-                	System.out.println(res);
-                }
+                new EvaluateExpressionCommand(args[0]).apply(repl, System.out);
                 //TODO Return actual exit value if parsing/evaluation fails.
                 System.exit(0);
             }
@@ -236,7 +234,7 @@ public class REPL {
 
 			System.out.println("Welcome to the TLA+ REPL!");
             MP.printMessage(EC.TLC_VERSION, TLCGlobals.versionOfTLC);
-        	System.out.println("Enter a constant-level TLA+ expression.");
+            System.out.println("Enter a constant-level TLA+ expression or type :help for help.");
 
             repl.runREPL(reader);
         } catch (Exception e) {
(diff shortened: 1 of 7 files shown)


ahelwer:
Do you think it makes sense to print this out if given the `-help` command line parameter at startup too, for uniformity with the other tools?

lemmy:
Have you considered defining a `REPL` module that defines a `help` or `Help` operator?

Calvin-L:
> Do you think it makes sense to print this out if given the `-help` command line parameter

Maybe! Although I think `-help` usually gives information about command-line flags, not REPL interaction. (E.g. compare `python3 --help` to `python3 <<<'help()'`.)

Calvin-L:
> Have you considered defining a `REPL` module that defines a `help` or `Help` operator?

That's an interesting approach I hadn't thought of. It could certainly work, but I am imagining we want to add other commands that do something to the REPL itself, like `:exit`. It's not clear that such functionality could be implemented in a TLA+ module.

lemmy:
An `exit` command should only require a Java module override that calls `System#exit`. :-)

Calvin-L:
In the future the REPL might spawn TLC in a separate process (see [comment](https://github.com/tlaplus/tlaplus/issues/627#issuecomment-844538528): "there might be subtle bugs lurking"). If so, `System#exit` won't work as intended. (Although, I suppose we could certainly define `Quit == FALSE` and then special-case `Quit` when typed in the REPL.)

I read the discussion in #627 over the weekend. Maybe I missed something, but my takeaway was that (1) the biggest deficiency in the REPL is poor error reporting but (2) several users are also interested in new features. I think a help command is orthogonal to (1) and a prerequisite for (2), otherwise nobody will be able to discover and use the new features.

----

Regarding a dedicated TLA+ `REPL` module, here are two points of comparison:

"Python-style": all the REPL functionality is written in Python. `help()` and `quit()` are Python procedures, and you can call them from any Python code. A TLA+ `REPL` module would build toward a Python-style REPL.

"GHCi-style": the REPL has lots of extra commands prefixed with a special character, like `:help`, `:quit`, `:load <module>`, `:reload`, `:show bindings`, `:type <expr>`, and dozens more.

I am more inclined toward "GHCi-style" because there are some kinds of actions I want to take in the REPL that don't "feel" like part of the TLA+ language to me. When would I ever write a TLA+ module that uses `Help` or `Quit`? We might also want to implement a command like `:help OperatorName`, and it's hard to see how that might be shoehorned into TLA+. But, maybe I can be swayed. :)

lemmy:
I suggest to collect concrete REPL commands and then assess if we are better served by the GHCI or the Python style.

lemmy:
What would `:help OperatorName` show?

Calvin-L:
I'm imagining `:help OperatorName` would show [the comments associated with the definition of `OperatorName`](https://github.com/tlaplus/tlaplus/blob/0dbe98d51d6f05c356303b5e0e7d8ae4a85d144b/tlatools/org.lamport.tlatools/src/tla2sany/st/TreeNode.java#L19)---but I'm not yet sure how practical it would be to build such a feature.

lemmy:
> I'm imagining `:help OperatorName` would show [the comments associated with the definition of `OperatorName`](https://github.com/tlaplus/tlaplus/blob/0dbe98d51d6f05c356303b5e0e7d8ae4a85d144b/tlatools/org.lamport.tlatools/src/tla2sany/st/TreeNode.java#L19)---but I'm not yet sure how practical it would be to build such a feature.

I once prototyped a SANY module that exposed some of SANY's functionality...

Calvin-L:
> I suggest to collect concrete REPL commands and then assess if we are better served by the GHCI or the Python style.

Here are the commands I have in mind (because I would use them if they existed):

 - Show help
 - Show help for a given name (e.g. if it is an operator then its associated comment, its arity, where it is defined; if it is a module, then its associated comment and list of exported operators; etc.)
 - Quit (the REPL already supports Ctrl+D, which is common to all major shells, but I'm not sure it's universal)
 - `DEFINE Op == Expr` (this was the original request in #627; I would use the TLAPS `DEFINE` syntax for it, pending discussion)

Here are a few that we might consider:

 - List definitions (e.g. to copy-paste into a .tla file)
 - Import (`EXTENDS`) another constant module
 - Remove or alter a previous `DEFINE`
 - Remove an imported module
 - Ask TLAPS to prove something

We should probably limit ourselves to thinking about the top list for now.

lemmy:
Some preliminary considerations:

- If deciding between Python and Haskell styles is difficult, consider supporting both through a configuration option.
  => Implement REPL commands as TLA+ operators and implement generic support to wire haskell-style commands?!
- How should we differentiate between the functionality of a REPL and the evaluation of expressions within a scratch-pad module? For instance, I personally prefer defining operators in the editor (which offers more editing features) and then using those definitions dynamically within the REPL (already possible via `LET M==INSTANCE MyModule IN M!Op`)
- Will the REPL eventually support formulas that aren't constant?

PR #1002, opened 2024-09-04
Remove old parser files

ahelwer:
In my opinion as someone who works on the parser these files are not useful and are just clutter. Predominantly they are old versions of the auto-generated grammar code. There are also old versions of the grammar files themselves, which are also not very useful; I can think of no reason why I would want to know how something was parsed a decade ago when we have a fairly thorough standard & tests for how TLA+ should be parsed.

One other deleted file is `javacc/README/methods-sany1.txt` which is just a copy of the file (which is being kept) named `javacc/README/methods.txt` but with some added incorrect speculation about the purpose of various files.


lemmy:
I believe these files predate our use of CVS -> SVN -> git.

lemmy (approved):

Issue #1005, opened 2024-09-05
PlusCal parse validation fails non-catastrophically when running SANY on Unicode specs on Windows GitHub CI runners


ahelwer:
Annoyingly I cannot reproduce this on my local Windows install so it's possible this is not actually a problem and is just Windows Unicode weirdness.

You can see the problem by going to any tlaplus/examples CI run and looking at the logs of the `(windows-latest, true)` "Check small models" step. You'll see something like:
```
2024-09-04T21:09:40.9721893Z DEBUG:root:java -enableassertions -Dtlc2.TLC.ide=Github -Dutil.ExecutionStatisticsCollector.id=abcdef60f238424fa70d124d0c77ffff -XX:+UseParallelGC -cp deps\tools\tla2tools.jar;deps\apalache\bin\apalache-mc\lib\apalache.jar;deps\community\modules.jar;deps\tlapm\library tlc2.TLC specifications\Bakery-Boulangerie\MCBakery.tla -config specifications\Bakery-Boulangerie\MCBakery.cfg -workers auto -lncheck final -cleanup
2024-09-04T21:09:40.9724124Z TLC2 Version 2.20 of Day Month 20?? (rev: 2039099)
2024-09-04T21:09:40.9725641Z Running breadth-first search Model-Checking with fp 72 and seed 7814885188961549382 with 4 workers on 4 cores with 3641MB heap and 64MB offheap memory [pid: 4696] (Windows Server 2022 10.0 amd64, Eclipse Adoptium 17.0.12 x86_64, MSBDiskFPSet, DiskStateQueue).
2024-09-04T21:09:40.9727138Z Parsing file D:\a\Examples\Examples\specifications\Bakery-Boulangerie\MCBakery.tla
2024-09-04T21:09:40.9727945Z Parsing file D:\a\Examples\Examples\specifications\Bakery-Boulangerie\Bakery.tla
2024-09-04T21:09:40.9728365Z 
2024-09-04T21:09:40.9728466Z Unrecoverable error:
2024-09-04T21:09:40.9728955Z  -- Illegal lexeme `�' found at
2024-09-04T21:09:40.9729269Z     line 61, column 22.
2024-09-04T21:09:40.9729436Z 
[...]

PR #1006, opened 2024-09-05
Fix PlusCal Unicode handling on Windows

ahelwer:
I have hopes that this will address #1005, although I cannot actually test that it does on my Windows install locally.


Calvin-L (approved):
This is a clear improvement.

I will add that there are many other places where we don't specify the charset for an `InputStreamReader`:

Screenshot 2024-09-09 at 10 40 05 AM

Any of these locations might be a source of bugs on platforms where UTF-8 is not the default charset.

lemmy:
I don't have to time to contribute to this PR.

ahelwer:
This change is basically trivial so can be safely merged with only @Calvin-L's signoff, I think.

PR #1007, opened 2024-09-06
Add support for negative syntax corpus parser tests

ahelwer:
These changes only touch test code.

This adds support for negative syntax corpus tests: inputs which we expect to cause a parse error. One such test is added as an example. It also adds support for skipping syntax corpus tests.

Since these changes complicated the syntax corpus test running logic which is shared by the TLA+ and PlusCal parser test suites, I abstracted the runner code into its own function that these suites customize by implementing an interface.

I also renamed a bunch of related files to make the naming scheme consistent.


lemmy:
I won't be able to review this soon. If @Calvin-L is also unavailable, let's proceed with the merge since the risk of regressions is minimal.

Calvin-L (approved):

Issue #1008, opened 2024-09-08
Level parameters are not set for built-in dot infix operator (record access operator)


ahelwer:
It is unclear whether this is a bug because I do not know the level-checking system very well, so just close this issue if this is expected behavior.

The built-in TLA+ operators (`SUBSET`, `DOMAIN`, etc.) are loaded into a global context by parsing operator names & arities listed [here](https://github.com/tlaplus/tlaplus/blob/2039099c27a5310618d70a2ed9f58975dae77607/tlatools/org.lamport.tlatools/src/tla2sany/configuration/ConfigConstants.java#L60). Then, their level constraint data is set drawing from a list [here](https://github.com/tlaplus/tlaplus/blob/2039099c27a5310618d70a2ed9f58975dae77607/tlatools/org.lamport.tlatools/src/tla2sany/semantic/BuiltInLevel.java#L62). These two lists of operators correspond almost exactly, with two exceptions:

1. The `\neg` operator is defined in list of built-ins but does not have its levels set; this is probably not an issue because it's just a synonym of the `\lnot` operator, which *does* have its levels set (the other synonym `~` operator is neither defined in the list of built-ins nor does it have its level set).
2. The record access dot operator `.` is defined in the list of built-ins but is totally missing from the list of level constraints

I don't know whether this is an oversight or is simply not necessary, but noticed this while writing a test for this functionality and thought it should be brought up.

ahelwer:
Okay so the level of the `.` operator itself should be 0 (constant), how about the max levels of its two parameters? I am guessing `(2,2)` (both step-level) for max level, since you might want something like `rec'.name'`. What about the weights? I will try to take a look at the code to figure out why the lack of this info has not caused a bug in the level checker.

ahelwer:
Upon further examination the level of this operator is actually set as `$RcdSelect` instead of `.`. I will experiment to see whether the `.` version even needs to be set as as built-in operator.

PR #1009, opened 2024-09-08
[Patch 1/4] Add tests for built-in operator initialization process

ahelwer:
Before SANY can parse a module, there is a process that initializes the various built-in operators and sets their levels. This is done by calling `Configuration.load()` to set the operator name & arity then `BuiltInLevel.load()` to set the level constraints. This work adds a unit tests covering the operation of these two functions both to check that their action is as expected, and as protection to ensure any future refactoring of these functions preserves their action.

This is all test code with the exception of a visibility change on an internal class (which was probably meant to be publicly-visible, because instances of it are returned from a publicly-visible method) and addition of two methods exposing some internal level constraint information for testing purposes.

Issue #1008 was found during this work.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
@@ -30,6 +30,7 @@
 
 package tla2sany.semantic;
 
+import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.Hashtable;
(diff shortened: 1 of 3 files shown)


tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
```diff
@@ -1196,6 +1197,32 @@ public final boolean getOpLevelCond(int i, int j, int k) {
...
+   * Returns an array giving the maximum levels of the parameters to this
+   * operator, or null if no such information is available.
+   * 
+   * @return The maximum arg levels.
+   */
+  public final int[] getArgMaxLevels() {
+    return
+      this.maxLevels == null
+      ? null
+      : Arrays.copyOf(this.maxLevels, this.maxLevels.length);
```
lemmy, line 1210: Is returning a copy done primarily as a precaution? If so, let's add a comment explaining this to help our future selfs understand it.

tlatools/org.lamport.tlatools/test/tla2sany/TestBuiltInOperatorInitialization.java
```diff
@@ -0,0 +1,226 @@
...
+    new BuiltInOperator("$Take",                1,  0,  mk(3),      mk(1)),
+    new BuiltInOperator("$Pick",                1,  0,  mk(3),      mk(1)),
+    new BuiltInOperator("$Witness",             -1,  0, mk(2),      mk(1)),
+    new BuiltInOperator("$Suffices",            1,  0,  mk(3),      mk(1)),
+  };
+
+  /**
+   * Ensures all built-in operators are initialized as expected.
+   */
+  public static void testCorrectAndComplete() {
```
lemmy, line 173: Does this have to be public? Does this have to be static?

ahelwer:

PR #1010, opened 2024-09-08
[Patch 2/4] Statically initialize built-in operators instead of using parser

ahelwer:
This is the second patch in the series after #1009 and is based on top of those changes.

This is similar to work in #895 and completes the transition from using the JavaCC config.jj parser to process string operator data to using static Java initialization code.


ahelwer:

PR #1011, opened 2024-09-08
[Patch 3/4] Remove JavaCC config.jj parser

ahelwer:
This is the third patch in the series after #1010 and is based on top of those changes.

Now that we don't use the config.jj parser for anything, this removes it from the codebase. I know @Calvin-L was looking forward to this one.


ahelwer:

PR #1012, opened 2024-09-08
Statically initialize built-in operator properties

ahelwer:
The goals of this patch set are primarily:
1. Remove the vestigial `config.jj` parser from the SANY initialization code path, and indeed remove the entirety of the `tla2sany/configuration` directory
2. Remove the need to call two static initialization methods before SANY parsing works, in line with goals of #891
3. Add tests for the built-in operator initialization code, which found issue #1008; then added some level-checking tests to prove that the issue was not actually a problem.

The cumulative PR is fairly large but each commit is broken down into such a way that the CI will pass on all of them. There are also detailed commit messages.


lemmy:
Next time, you don't have to create four PRs if you make each PR description the commit message.

ahelwer:
Sure, I will do that. Felt like it would have been a bit of a monster PR. Wish github supported patch sequences.

lemmy:
https://github.com/tlaplus/tlaplus/pull/1012/commits

ahelwer:
Okay, I put all the PR descriptions into the commit messages so it's all just in this PR!

tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
Calvin-L, line 1: There is a now-invalid comment on line 414:
https://github.com/tlaplus/tlaplus/blob/ccacbf5ad2f1a898a746632c15ae61d59a0d4df1/tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java#L414

Calvin-L (commented):
Added a few surface-level comments. I plan to do a closer read later in the week.

tlatools/org.lamport.tlatools/src/tla2sany/parser/Operators.java
```diff
@@ -148,4 +122,213 @@ public static void printTable() {
...
+    new String[] {"||", "‖"},
+    new String[] {"\\wr", "≀"},
+    new String[] {"^+", "⁺"},
+  };
+
+  /**
+   * Statically initializes all operators in the definition table. This will
+   * be executed when this class loads into memory.
+   */
+  static {
```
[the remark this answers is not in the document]
ahelwer: It was moved since it needs to be below the new static fields. I also changed the exception to a RuntimeException to match the built-in op initialization code.

Calvin-L (approved):

lemmy:
I don't have to time to contribute to this PR.

ahelwer:
That is okay, it is a pretty big one. We can discuss what to do at the next community meeting.

tlatools/org.lamport.tlatools/test/tla2sany/TestBuiltInOperatorInitialization.java
```diff
@@ -0,0 +1,78 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import tla2sany.semantic.ASTConstants;
+import tla2sany.semantic.BuiltInOperators;
+import tla2sany.semantic.BuiltInOperators.BuiltInOperator;
+import tla2sany.semantic.Context;
+import tla2sany.semantic.OpDefNode;
+import tla2sany.semantic.SymbolNode;
```
FedericoPonzi, line 14: This test references classes from tla2sany.semantic. If you moved it to the tla2sany.semantic package, you would not need to change Context.Pair visibility from package private to public.

tlatools/org.lamport.tlatools/src/tla2sany/parser/Operators.java
```diff
@@ -148,4 +122,213 @@ public static void printTable() {
...
+    new Operator("..",            90,  90,  Operators.assocNone, Operators.infix),
+    new Operator("|",             100, 110, Operators.assocLeft, Operators.infix),
+    new Operator("||",            100, 110, Operators.assocLeft, Operators.infix),
+    new Operator("&&",            130, 130, Operators.assocLeft, Operators.infix),
+    new Operator("&",             130, 130, Operators.assocLeft, Operators.infix),
+    new Operator("$$",            90,  130, Operators.assocLeft, Operators.infix),
+    new Operator("$",             90,  130, Operators.assocLeft, Operators.infix),
+    new Operator("??",            90,  130, Operators.assocLeft, Operators.infix),
+    // Removed as requested by Leslie (16 Feb. 01)
+    //new Operator("?",             90,  130, Operators.assocLeft, Operators.infix),
```
FedericoPonzi, line 181: This could be removed as well (IMO)

tlatools/org.lamport.tlatools/src/tla2sany/semantic/BuiltInOperators.java
```diff
@@ -0,0 +1,155 @@
...
+   * @return An array of the given values.
+   */
+  private static int[] mk(int... vals) {
+    return vals;
+  }
+
+  /**
+   * A list of all built-in operators along with their arities and levels.
+   */
+  public static final BuiltInOperator[] Properties = {
```
FedericoPonzi, line 80: How did you generate this table? By hand? I've compared the first two columns from tlatools/org.lamport.tlatools/src/tla2sany/configuration/ConfigConstants.java#defaultConfig and they look good to me.

tlatools/org.lamport.tlatools/test/tla2sany/TestLevelChecking.java
```diff
@@ -0,0 +1,155 @@
+package tla2sany;
```
FedericoPonzi, line 1: What is the test coverage of the new code?

FedericoPonzi (approved):
Minor comments, overall LGTM. I think it would have broken very loudly if it was not safe to merge, but what are the possible risks associated to this change? Pruning a codebase is always good :) nice work!

tlatools/org.lamport.tlatools/src/tla2sany/semantic/BuiltInOperators.java
ahelwer: I don't remember how I generated it. If you look at the sequence of commits it was part of a unit test to ensure it conforms with how the system functioned originally before it was moved to become core to the system itself.

tlatools/org.lamport.tlatools/test/tla2sany/TestLevelChecking.java
ahelwer: What do you mean?

tlatools/org.lamport.tlatools/src/tla2sany/parser/Operators.java
ahelwer: I think it is interesting for historical reasons and @lemmy likes to keep comments of this sort around.

tlatools/org.lamport.tlatools/test/tla2sany/TestLevelChecking.java
ahelwer: Oh, you mean like is the new table and the logic reading it in exercised. Yes that is tested.

tlatools/org.lamport.tlatools/test/tla2sany/TestBuiltInOperatorInitialization.java
ahelwer: I do think sorting the `tla2sany` tests into submodules modules in this way is a good idea though, so I added another commit which makes this change and also restricts the visibility of `Context.Pair` again.

ahelwer:
@FedericoPonzi if you think this looks good then I think @Calvin-L can reasonably merge it. It should probably be merged as a rebase commit instead of a squash commit to preserve the patch sequence structure in the git history.

ahelwer:
@FedericoPonzi are you satisfied with the changes?

ahelwer:
@Calvin-L if you aren't on vacation yet then @FedericoPonzi has signed off if you have time to merge this! If you are on vacation then enjoy vacation and see you in the new year.

Calvin-L:
Thanks @FedericoPonzi. Glad to have a second set of eyes on this. :)

PR #1016, opened 2024-09-27
Print a log message whenever simulator has finished to compute initial state

FedericoPonzi:
This will add a log line:
```
Computed 1 initial states...
Finished computing initial states: 1 distinct state generated at 2024-09-28 00:26:25.
```
It's not super informative, but it's helpful for the ux of the vscode extension. Without it, users need to wait about a minute to get another log line before seeing something on the table:
image
This adds the initial line:
image
I think it's helpful to know that something is happening behind the scenes.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/Simulator.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/Simulator.java
@@ -302,6 +302,8 @@ public class Simulator {
 			return MP.printError(EC.TLC_NO_STATES_SATISFYING_INIT_AND_CONSTRAINT);
 		}
 
+		MP.printMessage(EC.TLC_INIT_GENERATED1, String.valueOf(initStates.size()), "");
+
 		// It appears deepNormalize brings the states into a canonical form to
 		// speed up equality checks.
 		initStates.deepNormalize();


lemmy:
Correct me if I'm wrong but it also better aligns the output of simulation and modelchecking?

lemmy:
Did you check if the Toolbox handles the new log output gracefully?

FedericoPonzi:
Toolbox seems running fine:

image

> Correct me if I'm wrong but it also better aligns the output of simulation and modelchecking?

Yes, you're correct. This is the same output printed during modelchecking.

ahelwer (approved):

tlatools/org.lamport.tlatools/src/tlc2/tool/Simulator.java
```diff
@@ -302,6 +302,8 @@ public int simulate() throws Exception {
 			return MP.printError(EC.TLC_NO_STATES_SATISFYING_INIT_AND_CONSTRAINT);
 		}
 
+		MP.printMessage(EC.TLC_INIT_GENERATED1, String.valueOf(this.numOfGenStates.longValue()), "");
+
```
lemmy, line 306: Should we get the number of distinct states from the data structures `initStates` rather than from a statistics variable?

lemmy (approved):

lemmy (approved):

PR #1015, opened 2024-09-20
Simplified & documented junction list parsing code

ahelwer:
Handling vertically-aligned conjunction & disjunction lists (junction lists) is one of the most important & complicated parts of parsing TLA+. These changes are a functionally-equivalent refactor which simplifies & improves documentation for the code that tracks the stack of nested junction lists. I also removed some old commented-out parsing code that no longer touches the junction list parsing code; this commented-out code served no pedagogical purpose other than confusion. Overall these changes will make life easier for whoever comes after me that tries to understand junction list parsing.

As a bonus, this also frees up parser users to parse finer-grained units than an entire module. They can now call `Expression()` for example to parse a single expression.

The syntax corpus has a very large number of tests for junction list parsing so test coverage of these changes is very good.

PR #1017, opened 2024-09-30
Removed dead code from JavaCC-generated parser

ahelwer:
I used Eclipse's nice test coverage highlighting to identify code in the JavaCC-generated parser which is not exercised, then confirmed that no reference to any of these functions exists. Removing them increases test coverage of `TLAplusParser.java` from 69.9% to 76.8%, in addition to making the grammar easier to understand. All told about 700 lines are removed.

There were some ASCII page break characters I also removed. I also took the liberty of replacing the exactly three tab characters in the entire file with spaces.

Finally, I also automatically regenerate the `javacc/tla+.html` file in the Ant `generate` task now.

This will cause merge conflicts with #1015 and #1019 so I'll resolve them in whichever gets merged last.

### DIFF
--- a/tlatools/org.lamport.tlatools/javacc/tla+.jj
+++ b/tlatools/org.lamport.tlatools/javacc/tla+.jj
@@ -56,19 +56,15 @@ config.jj.
 
 // Copyright (c) 2003 Compaq Corporation.  All rights reserved.
 
-// WARNING - order of definition of tokens *does* matter : check ClosedExpressionOrOp
   /*************************************************************************
-  * ClosedExpressionOrOp is irrelevant here.  The "order of definition of  *
-  * tokens" in this file determines the actual numbers assigned to the     *
-  * `kind' fields of the tokens.  (Those assignemnts are put by the        *
-  * parser in TLAplusParserConstants.java.)  These values are used in the  *
-  * definitions of isOp and is...fixOp, which I imagine are called         *
-  * somewhere in methods called by methods called by ...  methods called   *
-  * by ClosedExpressionOrOp.  The other thing to note is that the values   *
-  * of token `kind' fields must be different from those of nodes' kind     *
-  * fields, which are defined in st/SyntaxTreeConstants.java.              *
+  * The order of definition of tokens in this file determines the actual   *
+  * numbers assigned to the 'kind' fields of the tokens (those assignments *
+  * are put by the parser in TLAplusParserConstants.java). These values    *
+  * are used in the definitions of isOp and is(Pre|In|Post)fixOp. The      *
+  * other thing to note is that the values of token 'kind' fields must not *
+  * overlap with those of syntax tree nodes' 'kind' fields, which are      *
+  * defined in st/SyntaxTreeConstants.java.                                *
   *************************************************************************/
-  
 
 // Handling of FieldName : when the token is a keyword, the type of the token is changed to IDENTIFIER, and then we proceed with regular Identifier() handling.
 
@@ -199,24 +195,6 @@ public class TLAplusParser implements tla2sany.st.SyntaxTreeConstants, ParseTree
 //   ParseTree.setNumberFlags( numberFlag, decimalFlag );
     return PErrors.empty();
   }
-
-  private SyntaxTreeNode anchor = null;
-    /***********************************************************************
-    * This is set to a non-null value only in Expression().  If a          *
-    * ClosedExpressionOrOp node is found when parsing an expression,       *
-    * anchor is set to that node.  It is set to null in the following      *
-    * places:                                                              *
-    *                                                                      *
-    *  - When beginning to parse a Substitution()                          *
-    *                                                                      *
-    *  - When OpSuite() or Substitution() finds that anchor equals an      *
-    *    Op Symbol, as described below.                                    *
-    *                                                                      *
-    * The Substitution() and OpSuite() procedures look for an              *
-    * Expression or Op Symbol by calling Expression() and, if that throws  *
-    * an exception, catching the expression and checking if anchor is the  *
-    * desired Op Symbol.                                                   *
-    ***********************************************************************/
 
 // Lookahead mechanisms for definitions
   /*************************************************************************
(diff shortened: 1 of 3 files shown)


Calvin-L:
I just merged #1015. Please resolve the conflicts when you get a chance.

tlatools/org.lamport.tlatools/javacc/tla+.jj
```diff
@@ -1064,7 +978,7 @@ private boolean correctLevel(Token tok) {
 
 
 PARSER_END(TLAplusParser)
-
+
```
Calvin-L, line 981: Note for self: these apparent no-op whitespace changes are removal of "form-feed" characters (ASCII 0x0C). Form-feed is meant to act like a page break. It has fallen out of fashion, and most text editors (including mine) do not render them well, so I'm in favor of their removal.

ahelwer: Aha, so they are actually ASCII!

ahelwer:
@Calvin-L done, thanks!

tlatools/org.lamport.tlatools/javacc/tla+.jj
```diff
@@ -413,38 +413,6 @@ public class TLAplusParser implements tla2sany.st.SyntaxTreeConstants, ParseTree
...
-      t = getNextToken();
-    }
-  }
-
-/***************************************************************************
-* Note: the non-terminal ClosedStart was commented out, apparently to be   *
-* replaced by this boolean-valued method.                                  *
-***************************************************************************/
-boolean
-ClosedStart( Token t ) {
```
Calvin-L, line 431: Farther down there is a comment that mentions this method:
```
// op_119 : dernier compteur : impact sur ClosedStart et autres !!!
```

The comment is in French. It should be updated, and if we're doing that we might as well also translate it to English, something like:

```
// NOTE: op_119 is the last operator token, and it is used in helper methods to detect certain classes of tokens.
// We have to be careful to adjust those uses if we ever add new operator tokens.
```

tlatools/org.lamport.tlatools/javacc/tla+.jj
```diff
@@ -610,17 +542,6 @@ if (opArgs.kind != N_OpArgs) { ToolIO.out.println("Bug: not N_OpArgs node"); };
      } ;
     }
 
-// predicate used in lookahead to discriminate between the Case Separator and
-// the box operator. Returns true if it is most likely not the separator.
-// This is a weak mechanism.
-  boolean boxDisc() {
```
Calvin-L, line 616: This dead code is mentioned in a (presumably) outdated comment below:
```
// boxDisc (riminate) uses preInEmptyTop
// note that junction is processed separately.
```

tlatools/org.lamport.tlatools/javacc/tla+.jj
```diff
@@ -2304,72 +2218,6 @@ expecting = "Expression or Op. Symbol"; }
...
-*     ( Identifier | NonExpPrefixOp | InfixOp | PostfixOp)                 *
-*     <SUBSTITUTE> ( <op_76> | Lambda | Expression )                       *
-*                                                                          *
-* Note: <op_76> is "-.", the prefix - operator.                            *
-*       <SUBSTITUTE> is "<-"                                               *
-*                                                                          *
-* Modified 27 March 2007 by LL to allow Lambda substitutions.              *
-***************************************************************************/
-SyntaxTreeNode
-OldSubstitution() : {
```
Calvin-L, line 2318: This is mentioned in `tla+.html`. It isn't used there either.

Is that file autogenerated? It looks handwritten. If it's handwritten, let's remove this production from it by hand, as well as any others removed by this PR.

tlatools/org.lamport.tlatools/javacc/tla+.jj
```diff
@@ -3192,217 +3029,6 @@ AssertStep() : {
     }
 }
 
-// SyntaxTreeNode
-// NumerableStep() : {
```
Calvin-L, line 3196: If we want SANY to do better proof parsing, are we going to need this production?

It's mentioned in a few other places:
 - SyntaxTreeConstants.java has an unused `N_NumerableStep` constant
 - Generator.java has an unused `generateNumerableStep` method

Is there any chance we'll miss those later?

tlatools/org.lamport.tlatools/javacc/tla+.jj
```diff
@@ -3413,28 +3039,6 @@ ParenthesesExpression() : {
   {  return tn; }
 }
 
-SyntaxTreeNode
-ClosedExpressionOrOp() : {
```
Calvin-L, line 3417: This production is mentioned in a big comment up top:

    check ClosedExpressionOrOp [...] ClosedExpressionOrOp is irrelevant here

as well as another:

    If a ClosedExpressionOrOp node is found when parsing an expression, [...]

I'm not entirely sure how those comments should be updated, but they need attention.

tlatools/org.lamport.tlatools/javacc/tla+.jj
ahelwer: Unlikely, looking at the places in `TlaPlusParserOutputTranslator.java` that required more work to draw structure out of the parser output, I see these trouble areas:
1. The use-or-hide construct
2. The assume/prove construct
3. The TAKE proof step type
4. The PICK proof step type

None of which have much to do with what this rule appears to do.

tlatools/org.lamport.tlatools/javacc/tla+.jj
ahelwer: I believe tla+.html is auto-generated by JavaCC. It may have been subsequently modified by hand. I will have to look into how to generate it to keep it up to date.

tlatools/org.lamport.tlatools/javacc/tla+.jj
ahelwer: I rewrote the first comment and looking at the second comment & the generated code, it turns out that anchor element is totally unused (value set but never read) so I just deleted it.

tlatools/org.lamport.tlatools/javacc/tla+.jj
ahelwer: Deleted

tlatools/org.lamport.tlatools/javacc/tla+.jj
ahelwer: I figured out how to do it and added another commit in this PR adding the functionality to the `generate` ant task.

ahelwer: I split this out into this PR: https://github.com/tlaplus/tlaplus/pull/1043

This does mean that if this or that PR is merged first, the other will have to be updated or else the CI will fail when both are merged.

tlatools/org.lamport.tlatools/javacc/tla+.jj
Calvin-L: Ok, as far as I can tell, the functionality of this production has been superseded by the more general `Step()` production. We should make a point to clean up SyntaxTreeConstants.java and Generator.java in a separate PR.

Calvin-L:
This is, as far as I can tell, a safe change that does not remove anything we'll miss. (It's always in the Git history if we do.)

PR #1019, opened 2024-09-30
Removed parser support for archaic set-except syntax

ahelwer:
Currently, SANY will successfully parse this file if run without semantic or level checking using the `-s` and `-l` flags like `java -cp dist/tla2tools.jar tla2sany.SANY -s -l Test.tla`:
```tla
---- MODULE Test ----
x == {| S EXCEPT !.name = e |}
====
```
This is an old alternative syntax for sets that was never implemented above the syntax parser level. However, the syntax parser itself does accept it. These changes remove the code accepting this syntax and adds a corpus test to ensure it correctly results in a parse error.


lemmy:
@konnov I don't remember if Apalache uses SANY's semantic and level checking.  Can you confirm that Apalache has never supported this alternative syntax for sets?

tlatools/org.lamport.tlatools/src/tla2sany/st/SyntaxTreeConstants.java
```diff
@@ -107,12 +107,12 @@ public interface SyntaxTreeConstants {
   int  N_RcdConstructor  =  409 ;
   int  N_RecordComponent  =  410 ;
   int  N_SetEnumerate  =  411 ;
-  int  N_SetExcept  =  412 ;
+//  int  N_SetExcept  =  412 ;        // Removed by AH 05 Oct 2024
```
lemmy, line 110: Please use your full name, as future team members may not recognize who "AH" refers to.

konnov:
> @konnov I don't remember if Apalache uses SANY's semantic and level checking. Can you confirm that Apalache has never supported this alternative syntax for sets?

We never used this syntax. This is the first time I see it :)

lemmy (approved):

ahelwer:
@Calvin-L here's another quick one

PR #1025, opened 2024-10-02
`TLCExt!TLCEvalDefinition(name)` equals the value of the definition (reachable from the root module) with that name.

lemmy:
[Feature][TLC]


tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
```diff
@@ -355,4 +357,30 @@ public static synchronized IntValue tlcFingerprint(final Value val) {
...
+		
+		// Try to find the definition (not declaration, ...) starting at the root module.
+		final OpDefNode opDef = mn.getOpDef(sv.val);
+		if (opDef == null) {
+			throw new EvalException(EC.TLC_MODULE_ONE_ARGUMENT_ERROR,
+					new String[] { "TLCEvalDefinition", "name of a definition reachable from the root module", Values.ppr(v.toString()) });
+		}
+		
+		// Evaluate (the body of) the requested definition in the existing scope (context & states).
+		return tool.eval(opDef.getBody(), c, s0, s1, control, cm);
```
Calvin-L, line 384: Should this also check the operator's arity? It's not clear what `TLCEvalDefinition("F")` should do if `F` takes arguments.

lemmy: A non-zero arity operator causes an evaluation error later, because I didn't want to duplicate too much error checking in  TLCExt.java:

```
Error: Evaluating assumption line 78, col 8 to line 78, col 31 of module XXX failed.
Attempted to apply the operator overridden by the Java method
public static tlc2.value.impl.Value tlc2.module.TLCExt.tlcDefByName(tlc2.tool.impl.Tool,tla2sany.semantic.ExprOrOpArgNode[],tlc2.util.Context,tlc2.tool.TLCState,tlc2.tool.TLCState,int,tlc2.tool.coverage.CostModel),
but it produced the following error:
In evaluation, the identifier b is either undefined or not an operator.
line 76, col 5 to line 76, col 5 of module XXX
```

lemmy: Happy to add :

```java
...

if (opDef.getArity() != 0) {
  throw new EvalException(EC.TLC_MODULE_ONE_ARGUMENT_ERROR,
	new String[] { "TLCEvalDefinition", "a zero-arity definition", opDef.getSignature() });
}

return ...
```

```
Error: Evaluating assumption line 78, col 8 to line 78, col 31 of module SIMccfraft failed.
Attempted to apply the operator overridden by the Java method
public static tlc2.value.impl.Value tlc2.module.TLCExt.tlcDefByName(tlc2.tool.impl.Tool,tla2sany.semantic.ExprOrOpArgNode[],tlc2.util.Context,tlc2.tool.TLCState,tlc2.tool.TLCState,int,tlc2.tool.coverage.CostModel),
but it produced the following error:
The argument of TLCEvalDefinition should be a a zero-arity definition, but instead it is:
Foo(b)
```

Calvin-L: Yes, I think that's a little clearer.

lemmy: Done https://github.com/tlaplus/tlaplus/commit/f9b0a2fca6936bce93c8ef6e978899002e770c51

Issue #718, opened 2022-03-01
RFC: Remove features and API from TLA+ tools


lemmy:
Candidates for removal (number in brackets shows supporters of the feature):

### Feature
- Attach files of all modules in TLC's result email
- DFID (depth-first iterative deepening) model-checking
- Distributed TLC & Cloud TLC
- Checkpointing & recovery of the state queue, fingerprint set, and the trace (`-checkpoint` & `-recover` parameter)
- `-DTLA-Library` Java system property
- [1] Resuming reachability checking after a counter-example has been found (`-continue`  parameter)
- Overriding of built-in standard modules with custom ones
- Loading Java module overrides from class files in spec or `TLA-Library`
- ~SANY#XMLExport~
- Toolbox (in favor of [VSCode extension](https://github.com/tlaplus/vscode-tlaplus))

### API
- `tla2sany.semantic.FrontEnd.getToolId()`
- `tla2sany.modanalyzer.ParseUnit.writeParseTreeToFile(boolean, Errors)` (implemented by Jean-Charles Gregoire)
- `tla2sany.semantic.ModuleNode.isStandard()`, `util.FilenameToStream.TLAFile.isLibraryModule()`, `tla2sany.modanalyzer.ParseUnit.isLibraryModule()`

Related:
- #688 #378
- `tla2sany.modanalyzer.SpecObj.validateParseUnit(ParseUnit)` (remove `FileInputStream` from API) (`ParseUnit#NamedInputStream` is closed when `validateParseUnit` is called)

eras:
So these listed features are candidates for removal?

I suppose it's a bit sad to see all these features go, granted I had never used any of them but for two of them: the Toolbox (nowadays I use Emacs mostly), and the `-continue` switch. I do wonder if newcomers would find the VSCode Extension easier to approach, but I've never given it a try.

The `-continue` switch I've used to create a bunch of "example cases" from a state machine. I built support for handling that kind of output to my newly released tool for generating sequence diagrams out of state traces: https://github.com/eras/tlsd . Sometimes you might want to see more examples than just the first one (so basically random) and maybe gain some insight that way.

Maybe Distributed TLC is not that useful nowadays when you can get big machines affordably from e.g. Amazon for short runs, and if I recall correctly, DTLC was not able to check for liveness properties anyway.

zwergziege:
I tend to agree with eras. Would the removal of those features really do much in terms of maintainability / code quality?
[...]

PR #1033, opened 2024-10-06
Convert MODEL_PART_OF_JAR static variable to class variable

ahelwer:
Here's an easy fix to advance the goal of #891


lemmy:
https://github.com/tlaplus/tlaplus/issues/718

Issue #783, opened 2023-01-28
Feature request: completely redo TLC and SANY command line interfaces


ahelwer:
This would be a good project for a new contributor. Leave the legacy command line interfaces around if people want to keep using them, but provide alternative front-ends with the following:

1. Parsed with a standard command line argument parsing library, following conventions people expect; for example:
   * short `-h` and long `--help` versions of commands
   * automatically print out usage and help if commands are incorrect
   * use subcommands instead of mutually-incompatible flags and options
2. Use reasonable defaults; for example, use the equivalent of the following by default:
   * `-workers auto`
   * `-lncheck final`
3. Return well-known error codes for easier integration with CI and scripts (TLC seems to mostly do this but SANY needs work; see #781)
4. Expose a simplified set of parameters; power users can continue to use the legacy command line if they want

lemmy:
The current command-line is not great from a UX perspective and a more principled approach would make TLC more accessible to new users.  However, I propose for this issue to be more ambitious.  The major advantage of wrappers such as tlacli and tlabin is that they allow users to (mostly) work around TLC's config files.   Although we don't want users to write TLA+ on the command-line, it would be a major improvement if we find a more flexible replacement for TLC's config file that is exposed via the command-line.

ahelwer:
I actually like TLC config files, they are quite simple. It is just difficult to know the syntax due to lack of learning material; everything uses the toolbox to generate them. I often find myself referring to the actual cfg file grammar in specifying systems when I write them myself.

lemmy:
Indeed, config files are simple... actually they are too simple, forcing us to write an MC.cfg *and* an MC.tla file for anything serious.

lemmy:
Related: https://github.com/tlaplus/tlaplus/issues/781

ahelwer:
Wanted to highlight one example of how this issue affects me day-to-day: I cannot for the life of me remember all these names

- `tlc2.TLC`
- `tla2sany.SANY`
- `pcal.trans`

I have written them down in my notebook and have to look them up every time I want to run the tools. They are not easy to remember nor discover.

ahelwer:
@lemmy what do you think about changing the `java -jar tla2tools.jar` behavior to invoke this more advanced command line interface instead of just wrapping `tlc2.TLC`?

lemmy:
[...]

PR #1038, opened 2024-10-08
Convert TLC.traceNum to a non-static class variable

ahelwer:
Another easy fix to advance the goal of https://github.com/tlaplus/tlaplus/issues/891 and simplify testing.

Similar to #1033


lemmy:
I recommend keeping `tlc2.TLC` unchanged and creating a new entry point for `tla2tools.jar`, such as `TLA.java`, to specifically address thttps://github.com/tlaplus/tlaplus/issues/783. The existing `tlc2.TLC` would continue to be the default entry point until the new one is fully ready, which may take a while.

lemmy (approved):

PR #1046, opened 2024-10-19
Add syntax tests to improve parser code coverage

ahelwer:
Focused on filling a few red patches in `TLAplusParser.java` in the Eclipse test coverage highlighter. This only adds tests so cannot break anything.


Calvin-L:
Thanks for this. I always love to see more tests. :)

PR #848, opened 2023-11-20
Move resource files to a dedicated folder

Calvin-L:
Instead of collecting resource files from various places using complicated inclusion/exclusion rules, the tools now have a dedicated folder for them.

This change has (almost) no effect on the generated `tla2tools.jar` artifact, but should make development nicer in a few small ways:

 - It is easier to audit which resource files ship with the tools.  See the list of changes below for a list of subtle problems that this change corrects.

 - The project structure is now slightly more canonical.  Java projects generally separate the Java source folder from the resource files folder.

 - The build system can be slightly simpler: those complex inclusion/exclusion rules can be reduced to a blanket copy.

This commit is not a total no-op; there are a few small changes to the set of resource files that ship with `tla2tools.jar`, detailed below.

These files were NOT included, but now they are:

 - gson-LICENSE.txt (there was a bug in an inclusion rule; this file previously did not have the .txt extension and so it was not being included)

These files WERE included, but now they are not:

 - model/README.txt (this is dev documentation, and likely not useful for end users)
 - org/apache/commons/math3/CommonsMath-LICENSE.txt, org/apache/commons/math3/CommonsMath-NOTICE.txt (these licenses were being included twice; now they are only included once at the root)
 - pcal/OldPlusCal.cfg, pcal/PlusCal.cfg, pcal/PlusCal2.cfg, pcal/XPlusCal.cfg (these config files have no use for end users, especially since the corresponding TLA+ files were not included)
 - tla2sany/drivers/drivers - Shortcut.lnk (I suspect this file serves no function and should be removed; this commit excludes it from the final `tla2tools.jar` artifact but retains the file just in case)
 - tlc2/value/impl/SubsetValue.tla (should have been excluded like the other TLA+ files in `src`)

PR #1047, opened 2024-10-19
Move JavaCC files to tla2sany.parser module

ahelwer:
This moves all the JavaCC files next to where they are actually used. The Ant build file has been modified to account for this. I also wrote a preamble to the README file explaining the general layout and development methods of the syntax parser.

No source code is modified in these changes so they cannot introduce any bugs.


Calvin-L (changes requested):
We shouldn't move things into `src/` if we can help it. All non-java files in that folder are included in the distribution jar, meaning that after this change the distribution would include the javacc zip file and documentation. Not ideal!

Java projects canonically sort files by _type_ rather than _function_ (ref [Maven](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html), [Gradle](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html)), which can be awkward for humans but is great for build systems: "Everything here goes through javac, everything here gets copied, everything here gets run as a test, etc." See also #848, which we reverted but I always meant to revisit.

I could imagine renaming the `javacc/README/` folder to `lib/javacc/` (to live with its corresponding binary distribution) and adding a new `README.md` to the `javacc/` folder, but I wouldn't go further than that.

ahelwer:
Alas, the great dream is dead. How about things like READMEs? Where do you think those should go?

ahelwer:
cc @lemmy for more input on where to put READMEs, I would like to sprinkle some README files throughout the source tree to give a general overview of development tactics for that area of the codebase. Does this seem like a good idea? Where should the docs go if that is not a good place for them?

lemmy:
I don't mind having `README.md` files in Java source folders, as that's where people typically look for documentation. During the build process, the `.md` files can be excluded from the final JAR by configuring the filtering rules in the `customBuild.xml`.  That said, I would recommend using `package-info.java` instead. It is more compatible with Java build systems and offers better integration with IDEs.

PR #1055, opened 2024-10-31
SANY XML Exporter: added CLI help text and pretty-print option

ahelwer:
This will print out help text if CLI arg parsing fails or the `-h` parameter is passed in. Added a test for CLI arg parsing.

Also added a pretty-printer option so the XML will be output with indentation for human readability; by default it all goes out without any linebreaks.


ahelwer:
Going to split this up because it is getting a bit large

PR #1056, opened 2024-11-01
SANY XML: silence stdout during parsing then restore

ahelwer:
Before these changes the XML exporter was playing fast & loose with the state of stdout. These redirect only SANY's logging output to null without touching the state of stdout generally.

I split this out of #1055 because that was getting a bit large.


lemmy:
Does SANY output errors to stderr? I’m asking because it might be helpful for users to see SANY’s output when troubleshooting a failed export.

ahelwer:
That's true, I think it was right to pass in stderr to SANY. However, the general messing with stdout should be removed. I guess the concern here is that since stdout is being used to dump the XML, if anything else gets put to stdout by accident it will form invalid XML. I am pretty sure SANY does not randomly print to stdout but it could happen somewhere.

ahelwer:
Closing since I need to rework this with additional tests.

PR #1057, opened 2024-11-01
SANY XML: pretty-print by default and add terse CLI option

ahelwer:
These quick changes pretty-print the output XML by default, but add the `-t` CLI flag for terse output.

I split this out of #1055 because that was getting a bit large.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
@@ -17,6 +17,7 @@ import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerException;
 import javax.xml.transform.TransformerFactory;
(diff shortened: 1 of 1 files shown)


lemmy:
Under what circumstances would we need the non-pretty-printed output? In other words, is there a reason why the pretty-printed version couldn't fully replace the current output? Or is it that pretty-printing is too resource-intensive?

ahelwer:
I'm imagining that most of the consumers of the XML output are other programs for which extra newlines & tabs are not necessary; it's only in the initial integration that people might want to see the format.

I do think the extra tabs & newlines make a difference. I put in a very simple incrementing counter spec and the resulting XML was like 40,000 characters long.

lemmy:
I recommend making the pretty-printed version the default, while providing an option for consumers to switch to the shorter, faster variant if needed.

ahelwer:
That makes sense, I added a `-t` flag for terse output.

lemmy (approved):

Calvin-L:
FYI presence or absence of the new `-t` flag alters the meaning of the XML output: [whitespace is text in XML](https://www.oracle.com/technical-resources/articles/wang-whitespace.html), and XML parsers expose it in the parse tree. So
```
<a><b>value</b></a>
```
means
```
tag:a
  |- tag:b
    |- text:"value"
```
while
```
<a>
  <b>value</b>
</a>
```
means
```
tag:a
  |- text:"\n  "
  |- tag:b
    |- text:"value"
  |- text:"\n"
```

Calvin-L:
(I don't think we need to take any action here, but the distinction between `-t` and not `-t` will be crucial for downstream tools to keep in mind.)

PR #1058, opened 2024-11-01
SANY XML: added CLI help text

ahelwer:
It's difficult to add this in a way that helps in all cases without totally rewriting the very flawed parse logic but this is the minimal diff. Now outputs the following help text:
```
NAME

	tla2sany.xml.XMLExporter - Emit SANY's parse tree as XML - Version 2.2 created 08 July 2020


SYNOPSIS

	tla2sany.xml.XMLExporter [-Iot] [-help] FILE

DESCRIPTION

	Given a TLA+ file, parse that file with SANY then translate the module's semantic parse tree to XML, including all the modules depended on. The XML is printed to stdout and its output format is given by an XML Schema file (.xsd) found at https://proofs.tlapl.us/doc/web/sany.xsd.

OPTIONS

	-I
		Include; use given directory path to resolve module dependencies.
	-help
		Print this usage information.
	-o
		Offline mode; skip XML schema validation step.
	-t
		Terse; format XML output without tabs or newlines.

TIPS

	Only one root TLA+ file can be parsed per run.

	Multiple directory search paths can be given by providing multiple -I arguments.

	XML schema validation does not require network access.
```

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
@@ -11,7 +11,9 @@ import java.io.ByteArrayOutputStream;
 
 import java.io.PrintStream;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.LinkedList;
+import java.util.List;
 
 import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilder;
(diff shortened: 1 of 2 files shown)


tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
```diff
@@ -43,14 +43,31 @@
 
 public class XMLExporter {
 
+  /**
+   * CLI help text listing various options.
+   */
+  private static final String HELP_TEXT =
+    "SANY XML Exporter\n" +
+    "Usage:\n\n" +
+    "  java tla2sany.xml.XMLExporter [-o|-t] [-I <dir_path>]* <input_file>.tla\n\n" +
```
FedericoPonzi, line 52: Assuming the * means "0 or more instances of -I dir_path", do you have other examples of similar usages in man pages / helps?

ahelwer: Switched to using the UsageGenerator; see initial comment for example of help text generated.

ahelwer:
Added tests!

lemmy (approved):

PR #1062, opened 2024-11-05
[Provisional] Support running the TLA+ debugger with multiple workers.

lemmy:
Previously, the debugger could only run with a single worker (`-workers 1`), limiting its usefulness for large models. This commit introduces support for running the debugger with multiple workers (`-workers N`).

The debugger attaches to one worker, while the other workers continue without being directly debugged. This setup reduces debugging overhead, such as the creation of stack frames, on the non-debugged workers.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/util/SetOfStates.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/util/SetOfStates.java
@@ -29,6 +29,7 @@ package tlc2.util;
 import java.util.HashSet;
 
 import tlc2.tool.Action;
+import tlc2.tool.ITool;
 import tlc2.tool.ModelChecker;
 import tlc2.tool.StateVec;
 import tlc2.tool.TLCState;
@@ -60,6 +61,13 @@ public final class SetOfStates {
 		this.states = new TLCState[length];
 	}
 
+	public SetOfStates(final StateVec sv, final ITool tool) {
+		this(sv.size());
+		for (int i = 0; i < sv.size(); i++) {
+			put(sv.elementAt(i), tool);
+		}
+	}
+
 	public SetOfStates(final StateVec sv) {
 		this(sv.size());
 		for (int i = 0; i < sv.size(); i++) {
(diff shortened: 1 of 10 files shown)


lemmy:
I'm not pleased with the duplication in `SetOfStates` and the addition of `TLCState#fingerprint(Tool)`, but I haven't been able to come up with a better solution.  Thoughts anybody?

ahelwer:
I can take a look this weekend

tlatools/org.lamport.tlatools/src/tlc2/util/SetOfStates.java
```diff
@@ -72,6 +80,22 @@ public final void clear() {
...
+		for (int i = 0; i < old.length; i++) {
+			final TLCState s = old[i];
+			// This is where we have to redundantly compute the state's
+			// fingerprint. Thus, try to minimize the number of grow operations.
+			if (s != null) {
+				this.put(s.fingerPrint(tool), s, tool);
+			}
+		}
+	}
+
```
FedericoPonzi, line 98: unless I'm missing something, tool parameter is unused. It is passed down to this.put, which in turn passes it down to grow(tool). This duplication seems unneeded

lemmy: Worker#doNextCheckLiveness > SetOfStates#put > SetOfStates#grow

FedericoPonzi: What SetOfStates#put does with its `tool` argument is to pass it down to SetOfStates#grow, and what grow does with it is to pass it to put. It's doesn't seem to be used otherwise

lemmy: Note `s.fingerPrint(tool)` on 94 above.

FedericoPonzi: ah thanks! Somehow I've missed that reference

ahelwer:
Nothing comes to mind on a quick look but I haven't dealt with this code personally before so I don't have a model in my head of how it all relates.

Issue #947, opened 2024-06-14
SANY should output command line help text


ahelwer:
Related to #783.

SANY should output help text for its command line parameters in the following cases:

1. No parameters are given
2. Incorrect or unrecognized parameters are given

Currently the only way to figure out how the SANY command line parameters work is to read the code that processes them.

There is a `-help` parameter, but the documentation it prints out is incomplete. Also, it does not document the debug spec explorer shell commands.

FedericoPonzi:
Though SANY supports -help:
```
java -cp ./dist/tla2tools.jar tla2sany.SANY -help

NAME

        SANY - provides parsing, semantic analysis, and level-checking for a TLA+ spec - Version 2.2 created 08 July 2020


SYNOPSIS

        SANY [-ls] [-error-codes] [-stat]
        SANY [-ls] [-error-codes] [-stat]

DESCRIPTION

        SANY is a parser and syntax checker for TLA+ specifications.
        It catches parsing errors and some "semantic" errors such as
        priming an expression containing primed variables.

OPTIONS

        -error-codes
                If enabled, error level will be reported as the tools'
                return value.
        -l
                Turns off level checking. Level checking won't be
                used, if the semantic analysis is disabled. 
        -s
                Turns off semantic analysis
        -stat
                Turns off reporting statistics about builtin operator
                usage.
```
It's incomplete but most of the parameters are there

ahelwer:
Thanks for checking! I think I remembered being able to get it to print some things but not others (in particular dumping the parse tree, which I often use). This existing functionality will make implementing this fix a lot easier.

MathieuBordere:
I'll take a stab at this one as suggested by @ahelwer via lobste.rs

MathieuBordere:
`-help` would now output the following, still need to take a look at the `Semantic Graph Explorer` options.

- added -d option.
- minor rewording of existing options.
- added mandatory `FILE...` argument.

```
NAME

	SANY - provides parsing, semantic analysis, and level-checking for a TLA+ spec - Version 2.2 created 08 July 2020


SYNOPSIS

	SANY [-dls] [-error-codes] FILE...

DESCRIPTION

	SANY is a parser and syntax checker for TLA+ specifications.
	It catches parsing errors and some "semantic" errors such as
	priming an expression containing primed variables.

OPTIONS

	-d
		Opens the semantic graph exploration prompt for debugging.
	-error-codes
		The tools' return value will be the error-level.
	-l
[...]

PR #1064, opened 2024-11-07
SANY: Remove unused stats feature from commandline.

MathieuBordere:
As suggested in #947 , a little PR to get me started.

I also took the liberty to remove a line of code that was causing the synopsis to be shown twice.

```
SYNOPSIS

	SANY [-ls] [-error-codes]
	SANY [-ls] [-error-codes]
```

now looks like 

```
SYNOPSIS

	SANY [-ls] [-error-codes]
```


lemmy:
@MathieuBordere You will have to amend and sign the git commit, and then force-push it to this branch:

```
git commit --amend --sign
git push origin update-description-of-dump-option -f
```
The DCO checks needs to find something like Signed-off-by: Markus Alexander Kuppe <github.com@...> in the commit message.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -436,8 +420,6 @@ else if (args[i].equals("-L") || args[i].equals("-l"))
            doLevelChecking    = !doLevelChecking;
       else if (args[i].equals("-D") || args[i].equals("-d")) 
            doDebugging        = !doDebugging;
-      else if (args[i].equals("-STAT") || args[i].equals("-stat"))
```
lemmy, line 439: Please continue to accept the parameter. While it can be ignored, we should avoid breaking any existing scripts that may still pass this parameter, regardless of the reason.

lemmy (approved):

ahelwer (approved):

PR #1068, opened 2024-11-09
Level-checking: fix false acceptances in OpApplNode

ahelwer:
There are a number of places in `OpApplNode` where level-checking errors are logged but the level-checking process itself is not marked as failed. I fixed this and added positive & negative tests for all of these cases.

Originally this testing was motivated by noticing that this code was branching on ASCII versions of operator names like `[]` but that turned out not to be a problem.

### DIFF
--- a/tlatools/org.lamport.tlatools/test/tla2sany/semantic/TestLevelChecking.java
+++ b/tlatools/org.lamport.tlatools/test/tla2sany/semantic/TestLevelChecking.java
@@ -21,6 +21,17 @@ public class TestLevelChecking {
   private static class LevelCheckingTestCase {
 
     /**
+     * Interpolate expressions into this string to form semantically-valid
+     * TLA+ parse inputs.
+     */
+    private static final String INPUT_FORMAT_STRING =
+      "---- MODULE Test ----\n" +
+      "CONSTANT c\n" +
+      "VARIABLE x\n" +
+      "op == %s\n" +
+      "====";
+
+    /**
      * The string parse input.
      */
     public final String Input;
@@ -37,10 +48,10 @@ public class TestLevelChecking {
      * @param expected The expected level-checking result on the input.
      */
     public LevelCheckingTestCase(String input, boolean expected) {
-      this.Input = String.format("---- MODULE Test ----\n%s\n====", input);
+      this.Input = String.format(INPUT_FORMAT_STRING, input);
       this.ExpectedLevelCheckingResult = expected;
     }
-    
+
     /**
      * Summarize this test for error reporting purposes.
      *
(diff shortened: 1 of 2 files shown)


tlatools/org.lamport.tlatools/test/tla2sany/semantic/TestLevelChecking.java
```diff
@@ -67,7 +78,7 @@ public String summarize(Errors log) {
      * operator synonym resolution should happen before any operator details
      * are fetched. Here we test whether that hypothesis is true.
      */
-    new LevelCheckingTestCase("x == \\neg TRUE", true),
+    new LevelCheckingTestCase("\\neg c",         true),
```
lemmy, line 81: Why does the expression change? Additionally, there appears to be unnecessary indentation.

tlatools/org.lamport.tlatools/test/tla2sany/semantic/TestLevelChecking.java
```diff
@@ -77,14 +88,79 @@ public String summarize(Errors log) {
      * this hypothesis is true. For further info, see this issue:
      * https://github.com/tlaplus/tlaplus/issues/1008
      */
-    new LevelCheckingTestCase("x == 5.foo", true),
+    new LevelCheckingTestCase("c.foo",           true),
```
lemmy, line 91: Why does the expression change? Additionally, there appears to be unnecessary indentation.

tlatools/org.lamport.tlatools/test/tla2sany/semantic/TestLevelChecking.java
ahelwer: I just intended so all the booleans are lined up in the array. The expression changed because the string I was using for interpolation changed, adding constant & variable definitions as well as supplying an operator body to fill in.

lemmy (approved):

ahelwer:
It should be noted that SANY *does* actually return nonzero exit code in all these cases (when given `-error-codes` parameter), since it uses both the `levelCorrect` value returned by the function and also the `log.isSuccess()` function:
https://github.com/tlaplus/tlaplus/blob/50b00377c5a36eb5eb385d03d18c6faae0b6eaac/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java#L336-L342

So I suppose these changes could just be stated to bring the `LevelNode.levelCorrect` value in line with the `log.isSuccess()` value, for whatever future consumer mistakenly believes that they will always be the same.

Calvin-L:
Looks like a good fix! (Sorry for the delay; this one got lost in my inbox while I was away in November.)

PR #1067, opened 2024-11-09
SANY commandline enhancements

MathieuBordere:
Fixes #947 

- Show usage when no argument is provided.
- Show error and refer to `-help` when no file argument is provided.
- Slight rewording of existing options.
- Added mandatory `FILE...` argument.
- Added explanation for commands that can be used in the Semantic Graph Explorer.

This results in:

```
NAME

	SANY - provides parsing, semantic analysis, and level-checking for a TLA+ spec - Version 2.2 created 08 July 2020


SYNOPSIS

	SANY [-dls] [-error-codes] FILE...

DESCRIPTION

	SANY is a parser and syntax checker for TLA+ specifications.
	It catches parsing errors and some "semantic" errors such as
	priming an expression containing primed variables.

OPTIONS

	-d
		Opens the semantic graph explorer prompt. The prompt accepts the following commands:
		'cst' Prints out the concrete syntax tree.
		'dot' Emits the semantic graph to a ModuleName.dot file in the DOT graph description language.
		'mt'  Prints the module table, a list of all imported modules and their top-level definitions.
		'mt*' Prints the extended module table including built-in operator definitions.
		
		Optionally, skip the prompt and run a command directly by appending it like 'SANY -d ModuleName.tla cst'
	-error-codes
		Return a descriptive exit code in case of error.
		
		'1' Error during initialization.
		'2' Error during parsing.
		'3' Error parsing built-in operator tables.
		'4' Error during semantic analysis or level-checking
	-l
		Turns off level-checking.
	-s
		Turns off semantic analysis and level-checking.
```

PR #1069, opened 2024-11-10
SANY: Removed init error logging since it's static now

ahelwer:
PR #1067 made me realize that SANY has retained its error reporting machinery for initialization failures when those failures are no longer possible due to #1012. So I removed this machinery.

This did result in one functional change: when extending a spec, the new spec's symbols are read into the symbol context of the current spec (see `mergeExtendContext`). If any symbol conflicts occurred these were recorded in the initialization error log path (code comments reveal this behavior was an object of some confusion as these errors were getting "lost" so special code was necessary to report these errors). These errors will now be reported through the well-trodden semantic error log path.


Calvin-L (approved):

ahelwer:
@Calvin-L thanks for reviewing! I think @lemmy does not have time to review this so we are probably good to merge if you want.

Issue #996, opened 2024-08-13
`Value` API improvements


Calvin-L:
(This would be a big initiative; please chime in with feedback about how this should look and what should be in-scope.)

## Motivation

TLC supports Java module overrides, which is a fantastic feature that enables extending the model checker with new types and optimized operator implementations.

We should make it easy to support new kinds of `Value` objects, since some clever overrides may require new kinds of `Value`s---for instance, an optimized representation of an exponentially-sized set or function. Today it is entirely possible to write custom `Value` subclasses. However, it is not _simple_ to do so. It requires expert knowledge of TLC's internals, and there isn't a lot of support for the use case.

Because of #995, I do not think any module overrides currently implement their own `Value` types. Therefore now is a good time to make these changes, since many of these would be breaking changes for any existing `Value` types in module overrides.

(There is an existing `UserValue` type that I think was meant to fill this use case. However, `UserValue` only seems suitable for implementing infinite-size sets. Extending it to handle additional use cases essentially amounts to copying all the methods from `Value` to the `UserObj` type; I think a simpler approach is to simply make `Value` easier to extend.)

## Guiding Principles

 - `Value` is open (non-[sealed](https://docs.oracle.com/en/java/javase/17/language/sealed-classes-and-interfaces.html)); TLC should be able to work with any well-written implementation.
 - Implementing a new `Value` type should essentially amount to
   1. Implement all abstract methods
   2. Follow recommendations for handling exceptions and reporting cost information

## Proposed Changes

Documentation

 - Add docstrings to all of the `IValue`/`Value` methods
 - High-level advice about how to write a new `Value` type
 - Document how exceptions are handled (see common `catch (RuntimeException | OutOfMemoryError e)` pattern)
 - Document when and how to increment cost model counts
 - Document how values are compared

Major

 - Sort out the responsibilities of `IValue` versus `Value`.
   - I think these are intended to act like [`Collection`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collection.html) and [`AbstractCollection`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/AbstractCollection.html), where `Value` is just a convenient superclass with sensible default implementations for many `IValue` methods. However, the reality is that TLC makes frequent downcasts to `Value`, meaning that today all `IValue` implementations have to be `Value` subclasses.
   - For practical reasons I think we will have to stick with "all `IValue` implementations must extend `Value`".
[...]

PR #1077, opened 2024-11-20
Remove unused methods {has,get,set}Data()

Calvin-L:
It appears that these were never used.

See #996.


Calvin-L:
Got it, I will close this one. :) But, it might be nice to have a comment on these methods explaining why they are useful.

PR #1079, opened 2024-11-20
Add unicode support to state / value serialization

### LINKED ISSUE #1076: Deserializing non-ASCII strings from disk causes error
## Description

Reading non-ascii chars from disk causes an error. I tested this with IOUtils, but I suspect this is true of any time TLC stores states on disk and tries to read them back.

## Steps to reproduce

- Make `test.tla` with the contents below, and an empty `test.cfg`.
- Run TLC on that file with CommunityModules available. Default settings should be fine.

## Expected behavior

Evaluating the module  should print `"ü"`, produce a file `tmp.bin` that contains that serialized string, and pass the assertion that reading that file back gives the same string.

```tla
---- MODULE test ----
EXTENDS TLC, IOUtils

umlaut == "ü"

\* correctly prints ü
ASSUME PrintT(umlaut)

\* crashes with a message that looks like "%2%"
ASSUME
  /\ IOSerialize(umlaut, "tmp.bin", FALSE)
  /\ IODeserialize("tmp.bin", FALSE) = umlaut

====
```

## Actual behavior

Step 1 succeeds, step 2 appears to succeed, and step 3 fails with an error from the deserializer.

The problem is that TLC's binary serializer uses the deprecated version of String.getBytes, which is documented to strip the high bytes of all chars in the string. Similarly, the deserializer casts the bytes back to chars with no expectation that the high bytes mean anything.

## Steps taken to fix

There is no workaround for this, aside from not serializing non-ASCII strings. Fortunately, I can just do that for now.

## Possible fix

The on-disk format for strings is a length-prefixed byte sequence, so we have a lot of flexibility in terms of what could be in that byte sequence. The trick is to ensure existing files still work.

If we replaced the current encoding with UTF-8, all strings that already work with serialization with continue to work, since UTF-8 acts as a superset of ASCII.
As for the possibility that existing files contain valid Latin1 (the one-byte encoding that usually extends ASCII), as I showed in my example, Latin1 characters do not work currently. They are stored in-memory as modified UTF-16, so casting that to bytes breaks those chars too.

Switching to a UTF-8 encoding would therefore accept all existing binary files that work (since Latin1 is broken), and would also support arbitrary unicode text in serialized strings, fixing the problem.

I think the proposed change would impact about 4 methods that I found to be involved in the encode/decode steps (2 versions of encode, 2 versions of decode), as well as needing some additional tests.

I might have time to PR this, if there's interest.

## Environment

Linux, running TLC 1.8.0 pre-relase with latest CommunityModules as of writing. I've used both VSCode and a test script that launches the tools from the shell.

fhackett:
Fixes #1076.

This change replaces the serialization of TLA+ string values, so that TLC's binary serialization format supports unicode. The existing approach uses length-prefixed ASCII, and this change rewrites those methods to read and write length-prefixed UTF-8 instead.

Remind me to squash before merge. Currently the commits mark a history of the conversation here, but I can consolidate once we're sure things make sense.

## Incidental changes

1. Change the interface of `IDataInputStream#readString()` and `IDataOutputStream#writeString()`. They originally relied on the length prefix being read/written by the caller, which assumes a specific encoding. As I change the encoding, rather than update the callsites (though our UTF-8 encoding coincidentally keeps the same length prefix value), I moved the choice of length prefix inside those functions.
2. Delete `BufferedDataInputStream#readLine()` and `BufferedDataOutputStream#writeLine()`, which are ASCII-only and are unused throughout the codebase.
3. Delete `BufferedDataOutputStream#writeAnyString()` and `BufferedDataOutputStream#writeChars()`, which have no read- equivalent and are unused throughout the codebase. They are also ASCII-only.
4. Make `DiskByteArrayQueue.ByteValueInputStream` and `DiskByteArrayQueue.ByteValueOutputStream` public, so they can be instantiated during testing without dealing with `TLCState` objects.
5. Add `IValueInputStream#atEOF()`, which exposes a check for whether the entire input has been consumed. This enables asserting that no trailing bytes are left over after reading a value back during test, and is trivial to implement for both implementations of that interface.
6. Update the comments on the checkpointing test in `CodePlexBug08`, adding some notes explaining more details on how to do that. One version of these changes required a fair bit of problem solving there, so even though no change is required anymore, the comment seems valuable.

## Attempted alternatives

1. Encode the strings as UTF-8 using `ByteBuffer` and co, which is naturally compatible with ASCII. The content-dependent number of bytes the output would contain means that emitting a length-prefixed byte stream with the correct length is not possible. The resulting buffer manipulation and workarounds were complicated, and turned out to be unacceptably slow under test.
2. Write ASCII when the string could be represented using ASCII, and UTF-16 otherwise. This can be done by using the sign bit in the length prefix to indicate which encoding it is. Advantage: backwards compatible, acceptably efficient, and theoretically saves 50% space on many common strings. Disadvantage: 2 code paths per read/write, saves no space over always using UTF-16 and relying on gzip compression, backwards compatibility not a priority since serialization is generally short-lived.
3. Always encode UTF-16, which leads to very simple code that mirrors the original version but stores both bytes of every char. It is a mystery to me why this turned out to be measurably slower than calling `String#getBytes(StandardCharsets.UTF_8)`. Perhaps the JVM optimizes the array allocation in that method, or perhaps the overhead of writing 2x the bytes, gzip or no, dominates somehow. The benchmarks have spoken, so `getBytes` seems to be the way to go, as it is both faster and backwards-compatible.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/queue/DiskByteArrayQueue.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/queue/DiskByteArrayQueue.java
@@ -10,6 +10,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.Arrays;
 
@@ -589,7 +590,7 @@ public class DiskByteArrayQueue extends ByteArrayQueue {
 		  }
 	}
 	
-	static final class ByteValueOutputStream implements IValueOutputStream, IDataOutputStream {
+	public static final class ByteValueOutputStream implements IValueOutputStream, IDataOutputStream {
 
 		private byte[] bytes;
 		
@@ -738,19 +739,15 @@ public class DiskByteArrayQueue extends ByteArrayQueue {
 		 */
 		@Override
 		public final void writeString(String str) throws IOException {
-			final int length = str.length();
-			ensureCapacity(idx + length);
-			
-			final char[] c = new char[length];
-			str.getChars(0, length, c, 0);
-			
-			for (int i = 0; i < c.length; i++) {
-				this.bytes[idx++] = (byte) c[i];
-			}
+			final byte[] strBytes = str.getBytes(StandardCharsets.UTF_8);
+			this.writeInt(strBytes.length);
+			ensureCapacity(this.idx + strBytes.length);
+			System.arraycopy(strBytes, 0, this.bytes, this.idx, strBytes.length);
+			this.idx += strBytes.length;
 		}
 	}
 	
-	static final class ByteValueInputStream implements ValueConstants, IValueInputStream, IDataInputStream {
+	public static final class ByteValueInputStream implements ValueConstants, IValueInputStream, IDataInputStream {
 
 		private final byte[] bytes;
 		
@@ -925,12 +922,21 @@ public class DiskByteArrayQueue extends ByteArrayQueue {
 		 * @see util.IDataInputStream#readString(int)
 		 */
 		@Override
-		public final String readString(int length) throws IOException {
-			final char[] s = new char[length];
-			for (int i = 0; i < s.length; i++) {
-				s[i] = (char) this.bytes[idx++];
+		public final String readString() throws IOException {
+			final int length = readInt();
+			if(length == 0) {
+				return "";
 			}
-			return new String(s);
+			assert length > 0;
+			// UTF-8 unicode
+			final String result = new String(this.bytes, this.idx, length, StandardCharsets.UTF_8);
+			this.idx += length;
+			return result;
+		}
+		
+		@Override
+		public final boolean atEOF() {
+			return this.idx == this.bytes.length;
 		}
 	}
 
--- a/tlatools/org.lamport.tlatools/src/util/IDataInputStream.java
+++ b/tlatools/org.lamport.tlatools/src/util/IDataInputStream.java
@@ -32,6 +32,7 @@ public interface IDataInputStream {
 
 	int readInt() throws IOException, EOFException;
 
-	String readString(int slen) throws IOException;
+	/** Reads a length-prefixed UTF-8 string **/
+	String readString() throws IOException;
 
 }
--- a/tlatools/org.lamport.tlatools/src/util/IDataOutputStream.java
+++ b/tlatools/org.lamport.tlatools/src/util/IDataOutputStream.java
@@ -31,6 +31,7 @@ public interface IDataOutputStream {
 
 	void writeInt(int varLoc) throws IOException;
 
+	/** Write the string as length-prefixed UTF-8 **/
 	void writeString(String s) throws IOException;
 
 }
--- a/tlatools/org.lamport.tlatools/src/tlc2/value/IValueInputStream.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/value/IValueInputStream.java
@@ -58,4 +58,9 @@ public interface IValueInputStream {
 	IDataInputStream getInputStream();
 
 	UniqueString getValue(int idx);
+	
+	/**
+	 * @return whether the underlying data source is exhausted
+	 */
+	boolean atEOF();
 }
--- a/tlatools/org.lamport.tlatools/src/tlc2/value/ValueInputStream.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/value/ValueInputStream.java
@@ -222,6 +222,11 @@ public final class ValueInputStream implements ValueConstants, IValueInputStream
 	public final UniqueString getValue(int idx) {
 		return (UniqueString) this.handles.getValue(idx);
 	}
+	
+	@Override
+	public final boolean atEOF() {
+		return dis.atEOF();
+	}
 
   // @see ValueOutputStream#put
   private static class HandleTable {
--- a/tlatools/org.lamport.tlatools/src/util/BufferedDataInputStream.java
+++ b/tlatools/org.lamport.tlatools/src/util/BufferedDataInputStream.java
@@ -8,6 +8,7 @@ import java.io.FileInputStream;
 import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 
 import tlc2.output.EC;
 
@@ -28,6 +29,7 @@ public final class BufferedDataInputStream extends FilterInputStream implements 
     private int len;        /* number of valid bytes in "buff" */
     private int curr;       /* position of next byte to return */
     private byte[] temp;    /* temporary array used by various methods */
+    
 
     /* Object invariants:
        this.out == null <==> ``stream is closed''
@@ -218,73 +220,16 @@ public final class BufferedDataInputStream extends FilterInputStream implements 
         res <<= 8; res |= (temp[7] & 0xff);
         return res;
     }
-    
-    /** REQUIRES LL = SELF */
-    /** Reads and returns the next line of text from this stream, or
-        <code>null</code> if the stream is exhausted. Any end-of-line
-        character(s) are not included in the result. A line is termined
-        by a carriage return character (<code>'\r'</code>), a newline 
-        character (<code>'\n'</code>), a carriage return immediately 
-        followed by a newline, or by the end of the stream. */
-    public final String readLine() throws IOException {
-        String res = null;
-        while (this.len > 0) {
-            for (int i = this.curr; i < this.len; i++) {
-                if (this.buff[i] == (byte)'\n' || this.buff[i] == (byte)'\r') {
-                    // remember EOL character
-                    byte eol = this.buff[i];
-                    
-                    // create new substring
-                    String s = new String(this.buff, /*offset=*/ this.curr,
-					  /*count=*/ i - this.curr);
-                    if (res == null) res = s; else res += s;
-                    
-                    // skip over bytes in stream
-                    this.skip(i + 1 - this.curr);
-                    
-                    // skip '\n' if it follows '\r'
-                    if (eol == (byte)'\r' && this.len > 0
-                        && this.buff[this.curr] == (byte)'\n') {
-                        this.readByte();
-                    }
-                    return res;
-                }
-            }
-            // hit end of buffer -- append rest of buffer to "res"
-            String s = new String(this.buff, /*offset=*/ this.curr,
-				  /*count=*/ this.len - this.curr);
-            if (res == null) res = s; else res += s;
-            
-            // skip over bytes in stream
-            this.skip(this.len - this.curr);
-        }
-        // hit EOF without seeing EOL chars
-        return res;
-    }
 
-    public final String readString(int n) throws IOException {
-      char[] b = new char[n];
-      int off = 0;
-      while (n > 0) {
-	if (this.len < 0) throw new EOFException();
-	int offInit = off;
-	while (n > 0 && this.len > 0) {
-	  int toCopy = Math.min(n, this.len - this.curr);
-	  for (int i = 0; i < toCopy; i++) {
-	    b[off+i] = (char)this.buff[this.curr+i];
-	  }
-	  this.curr += toCopy; off += toCopy; n -= toCopy;
-	  if (this.curr == this.len) {
-	    // refill buffer from underlying input stream
-	    this.len = this.in.read(this.buff);
-	    Assert.check(this.len != 0, EC.SYSTEM_STREAM_EMPTY);
-	    this.curr = 0;
-	  }
-	}
-	int numRead = off - offInit;
-	off += numRead; n -= numRead;
-      }
-      return new String(b);
+    public final String readString() throws IOException {
+        final int strLen = readInt();
+        if (strLen == 0) {
+            return "";
+        }
+        assert strLen > 0;
+        final byte[] strBytes = new byte[strLen];
+        this.readFully(strBytes, 0, strBytes.length);
+        return new String(strBytes, StandardCharsets.UTF_8);
     }
   
     /** REQUIRES LL = SELF */
--- a/tlatools/org.lamport.tlatools/src/util/BufferedDataOutputStream.java
+++ b/tlatools/org.lamport.tlatools/src/util/BufferedDataOutputStream.java
@@ -7,6 +7,7 @@ import java.io.FileOutputStream;
 import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
 
 /** A <code>BufferedDataOutputStream</code> is an optimized
     combination of a <code>java.io.BufferedOutputStream</code>
@@ -184,60 +185,9 @@ public final class BufferedDataOutputStream extends FilterOutputStream implement
 	    this.writeLong(Double.doubleToLongBits(d));
     }
 
-    /** Write the characters of the string <code>s</code> to this
-        stream as a sequence of bytes. */
     public final void writeString(String s) throws IOException {
-        int n = s.length();
-        int off = 0;
-        while (n > 0) {
-            int toCopy = Math.min(n, this.buff.length - this.len);
-            s.getBytes(off, off + toCopy, this.buff, this.len);
-            this.len += toCopy; off += toCopy; n -= toCopy;
-            if (this.buff.length == this.len) {
-                // write buffer to underlying stream
-                this.out.write(this.buff, 0, this.len);
-                this.len = 0;
-            }
-        }
+        final byte[] strBytes = s.getBytes(StandardCharsets.UTF_8);
+        writeInt(strBytes.length);
+        this.write(strBytes, 0, strBytes.length);
     }
-    
-    /** Write <code>n</code> characters of <code>chars</code> starting 
-        at offset <code>off</code> to this stream as a sequence of bytes. */
-    public final void writeChars(char[] chars, int off, int n) throws IOException {
-        int finOff = off + n;
-        while (off < finOff) {
-            // Copy (part of) chars to this.buff
-            int endOff = Math.min(finOff, off + this.buff.length - this.len);
-            while (off < endOff) this.buff[this.len++] = (byte)chars[off++];
-
-            // If this.buff is full, write it out
-            if (this.buff.length == this.len) {
-                // write buffer to underlying stream
-                this.out.write(this.buff, 0, this.len);
-                this.len = 0;
-            }
-        }        
-    }
-    
-    /** Write the string <code>s</code> to the stream in such a way that it
-        can be read back by <code>BufferedDataInputStream.readAnyString</code>,
-        even if <code>s</code> is <code>null</code> or if it contains newline 
-        characters. */
-    public final void writeAnyString(String s) throws IOException {
-      if (s == null) {
-	this.writeInt(-1);
-      }
-      else {
-	this.writeInt(s.length());
-	this.writeString(s);
-      }
-    }
-    
-    /** Write the characters of the string <code>s</code> to this
-        stream as a sequence of bytes, followed by a newline. */
-    public final void writeLine(String s) throws IOException {
-        this.writeString(s);
-        this.writeByte((byte) '\n');
-    }
-
 }
--- a/tlatools/org.lamport.tlatools/src/util/UniqueString.java
+++ b/tlatools/org.lamport.tlatools/src/util/UniqueString.java
@@ -312,7 +312,6 @@ public final class UniqueString implements Serializable
         dos.writeInt(this.tok);
         dos.writeInt(this. getVarLoc()); 
          // Above changed from dos.writeInt(this.loc); by Yuan Yu on 17 Mar 2010
-        dos.writeInt(this.s.length());
         dos.writeString(this.s);
     }
 
@@ -328,8 +327,7 @@ public final class UniqueString implements Serializable
     {
         int tok1 = dis.readInt();
         int loc1 = dis.readInt();
-        int slen = dis.readInt();
-        String str = dis.readString(slen);
+        String str = dis.readString();
         return new UniqueString(str, tok1, loc1);
     }
     
@@ -337,8 +335,7 @@ public final class UniqueString implements Serializable
     {
         dis.readInt(); // skip, because invalid for the given internTbl
         dis.readInt(); // skip, because invalid for the given internTbl
-        final int slen = dis.readInt();
-        final String str = dis.readString(slen);
+        final String str = dis.readString();
         return UniqueString.uniqueStringOf(str);
     }
 
--- a/tlatools/org.lamport.tlatools/test/tlc2/tool/liveness/CodePlexBug08EWD840FL2FromCheckpointTest.java
+++ b/tlatools/org.lamport.tlatools/test/tlc2/tool/liveness/CodePlexBug08EWD840FL2FromCheckpointTest.java
@@ -66,6 +66,14 @@ public class CodePlexBug08EWD840FL2FromCheckpointTest extends ModelCheckerTestCa
 			 * 4) Locate the directory with the checkpoint data
 			 * 5) Replace the content of checkpoint.zip with the content of 4)
 			 * 6) Update the number below on states found...
+			 * 
+			 * Note from last updater:
+			 * To get a checkpoint, it may be more reliable to hardcode TLCGlobals.forceChkpt = true.
+			 * 
+			 * Worker count may also vary; if TLC fails with a missing file error, it means the worker count
+			 * is different between checkpoint and recover. You can adjust using -workers.
+			 * As a result, a detail or two of the counter-example may change.
+			 * 
 			 */
 			String prefix = BASE_DIR + File.separator + TEST_MODEL + "CodePlexBug08" + File.separator;
 			ZipFile zipFile = new ZipFile(prefix + "checkpoint.zip");
@@ -105,11 +113,11 @@ public class CodePlexBug08EWD840FL2FromCheckpointTest extends ModelCheckerTestCa
 	@Test
 	public void testSpec() {
 		assertTrue(recorder.recorded(EC.TLC_CHECKPOINT_RECOVER_START));
-		// Recovery completed. 1032 states examined. 996 states on queue.
+		// Recovery completed. 1510 states examined. 39 states on queue.
 		assertTrue(recorder.recordedWithStringValues(EC.TLC_CHECKPOINT_RECOVER_END, "1510", "39"));
 		// ModelChecker has finished and generated the expected amount of states
 		assertTrue(recorder.recorded(EC.TLC_FINISHED));
-		assertTrue(recorder.recordedWithStringValues(EC.TLC_STATS, "2334", "1566","0"));
+		assertTrue(recorder.recordedWithStringValues(EC.TLC_STATS, "2334", "1566", "0"));
 		assertFalse(recorder.recorded(EC.GENERAL));
 
 		assertNoTESpec();
--- a/tlatools/org.lamport.tlatools/test/tlc2/value/ValueInputOutputStreamTest.java
+++ b/tlatools/org.lamport.tlatools/test/tlc2/value/ValueInputOutputStreamTest.java
@@ -26,10 +26,14 @@
 package tlc2.value;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 import org.junit.Test;
 
@@ -39,6 +43,7 @@ import tlc2.value.impl.RecordValue;
 import tlc2.value.impl.StringValue;
 import tlc2.value.impl.Value;
 import util.UniqueString;
+import tlc2.tool.queue.DiskByteArrayQueue;
 
 public class ValueInputOutputStreamTest {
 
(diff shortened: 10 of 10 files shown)


fhackett:
@lemmy I would be particularly interested in any comments you have on the single broken test; any history if you remember, or if you know someone else who remembers.
I haven't looked in huge detail, but I've seen enough to know that it will take some digging - for starters, I can't easily find the .zip file it says is corrupted.

fhackett:
The functionality is there now, and I'm more confident in my tests.

I think the other test breakage was because my code still had multiple bugs yesterday, that I have now found and fixed.

Mysteriously, I am now dealing with pseudo-random test failures on my machine. I launch the test suite, and one different test fails each time. I will watch to see if CI does this too...

This is close to "ready for review", anyhow. Just need to update + improve the description a bit.

lemmy:
@fhackett I haven't had the chance to review your PR yet. Please note that while SANY now supports Unicode, this doesn't necessarily mean that TLC must also support Unicode or non-ASCII characters in TLA+ string values. If enabling Unicode support for string values creates significant challenges for TLC, it could instead raise an error whenever it encounters a non-ASCII character.

fhackett:
@lemmy That's a reasonable point. At minimum, I'm fine with this being a what-if PR. If I make no progress on "doing it right", then sure I can convert this into more of a small clean-up + error code addition.

--- rest is more specific discussion of both your comment and the CI results

Regarding potential other correctness issues, I would say the issue may not be fundamental. I think most of TLC already supports unicode perfectly, on account of "normal" usage of Java's string APIs trivially granting this.

Thinking about things like `tlc2.value.impl.StringValue.fingerPrint(long)`, for example, I don't think the "wrong" length value isn't actually a problem. It counts how many shorts are in the underlying array, and then that array gets individually fingerprinted element by element. Problems would arise if people start trying to imagine string indices as indicating "letters" in an alphabetic sense (outside of the domain where they actually do), which I am fine calling an error. Most operations, however, do not make this assumption. Maybe only string indexing / slicing might need to be careful, if people use them while believing they are handling "letters". Anything else can just believe the possibly-textually-inaccurate length (except maybe the TeX formatter if it uses it for alignment judgements?).

Theoretical considerations aside, the CI results are actually quite interesting this time. My code passes all the feature tests, but causes a huge performance regression on the integration tests. It seems my "optimizations" may be incorrectly titled... or maybe they made it less bad, but it's still unacceptably slow.

The next logical step is a small performance study of some sort. My big trip this week-end is approaching and moving preparations are increasing, so I may not have time to do that immediately, but I should try it soon.

My surface-level thought is that text encoding is probably slower than copying the low 8 bits of every element of an array of shorts into a buffer. However, the degradation seems too bad for this to be just the overhead of using Java's (well used and perf tuned I hope?) text encoders. Maybe it's also to do with the inability to predict the length of the resulting byte stream? (see my code for all sorts of attempted work-arounds)  Perhaps I am just doing something wrong performance-wise. I would be interested to investigate this for my own learning anyhow, as this is not the only place I end up writing low-level JVM text handling code.

fhackett:
Note to self: it may be possible to not use UTF-8.

The string length is stored as a signed integer, and since no valid string length will ever be negative, the sign bit is available for bit stealing!

So, if a non ASCII compatible binary format ends up being a good approach (e.g., directly writing a string's 2 byte chars and then relying on gzip to save the size overhead?), then backwards compatibility can be maintained by reading positive lengths as before, and negative lengths using some new format.

Needs actual experiments etc, but it just hit me that this would be possible.

fhackett:
I had a bit more time, and while I shouldn't be too sure of my self, I wanted to see what happened if I replaced the `ByteBuffer` and related manipulation with mostly for loops (that are also much simpler to reason about).

I wonder what the performance results will be for this one... leaving this here for now.

lemmy:
It seems reinstating our performance benchmark runs would be beneficial. @juliaferraioli, do you have any updates on the status of the sponsorship request I submitted a few months ago?

fhackett:
Interesting - I didn't know there were more precise benchmarking tools.

For now though, I guess my most recent changes pass the integration test check at least. Yay!

Calvin-L:
> TLC will transparently continue to read its existing disk format

Is backwards-compatibility actually important? If no, then this change could be substantially simpler. If yes, then we should make an effort to prevent regressions by adding tests that deserialize a few example byte arrays.

lemmy:
A user might attempt to recover from an old checkpoint, but I don't believe it's a priority for us to provide support for that.

Calvin-L:
The only other case I can think of are `*TTrace.bin` files. Could a user want to deserialize a trace from an older version of TLC?

lemmy:
It’s possible, but the .bin file can be regenerated from its textual representation in the *TTrace.tla file. Additionally, if needed, we could provide a standalone tool to convert an old .bin file to the new format, should that issue arise.

fhackett:
Compatibility concerns aside, the other concern I had in mind was binary file size.

Keeping the ASCII option means a lot of strings take 1/2 the space relative to UTF-16. Interestingly, this mirrors current day JVM, which internally uses Latin1 encoding for strings when it can get away with it, compressing a lot of common strings by 2x. Of course, we can't directly access this, and it may be neither here nor there.

It may be that the bzip compression option makes this a pointless concern. In that case, I'm happy to just encode as UTF-16 and drop the ASCII option if everyone else is fine with it.

Once I find more cycles, I could check the size overhead with/without bzip, to see how much this matters.

ahelwer:
Why is the string being serialized in UTF-16 instead of UTF-8? UTF-16 is also variable-width.

fhackett:
@ahelwer because in UTF-16 the byte count == `String#length() * 2`. Sure some characters are surrogates, but in terms of the actual bytes (well, shorts) going in the file, we can directly deduce the length prefix from the String. With UTF-8, some chars (in the Java sense) are 1 byte, some are 2, etc. The chars are essentially _variable-width wrt. Strings in Java_ because it's a non-native encoding*.

Note*: never mind that internally the String object's array is actually often in Latin1.

In a streaming context, this means we'd have to serialize the string in-memory and then look at the length of the byte array to figure how many bytes we will actually write.

My first commit to this PR was a UTF-8 attempt, but it ended up being complicated and slow. Some of that complication was due to misguided optimization attempts. I have a suspicion that using `ByteBuffer` / `CharBuffer` was causing some de-optimization (after reading around on StackOverflow and seeing related discussions). It's possible that just calling `String#getBytes(StandarCharsets.UTF_8)` does not cause problems in practice, despite transiently duplicating the entire string.

All that said, unless there's a specific need for UTF-8 or ASCII compatibility (as I was surprised to see above, there is actually no space overhead with gzip enabled), it seems that copying the String's contents char-for-char onto disk as UTF-16 is the most "obvious" approach that is least likely to slow down model checking.

ahelwer:
I don't understand. UTF-16 is variable-width. Some codepoints are more than two bytes. Assuming that each codepoint is two bytes is a notorious source of bugs, because it is so rarely exercised.

fhackett:
@ahelwer the `char` data type is not variable-width, however.

The bugs I think you're talking about are when people assume that 1 `char` === 1 code point, which I agree is false.

The code I've written makes no pretense of expecting that each `char` in the string has meaning on its own. It just reads/writes `String.length` instances of the `char` datatype. These may contain multi-`char` code points split into 2-byte parcels.

Consider this string `"🔥"`, which is a 2-`char` String (but containing one code point) encoded as `short[] {0xd83d, 0xdd25}`. `"🔥".length()` evaluates to 2, so my code stores a length prefix of 2 and writes 2 shorts, `0xd83d` and `0xdd25`. Giving the resulting bytes to a UTF-16 parser will correctly regenerate the original string (per my testing, at least).

I have test cases with emojis, the null character, and Zalgo text to try and prove to myself that this always works.

Please suggest any further tests that might break this code; I'd be interested to know what they are. More tests is always good, in any case.

Calvin-L:
To add to @fhackett's points: In Java, ["`String` represents a string in the UTF-16 format"](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html). That makes UTF-16 a convenient and fast format for Java programs to use. Many `String` methods like `String.length()` and `String.charAt()` directly expose the UTF-16 representation in terms of [16-bit `char` values](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html).

> I have a suspicion that using `ByteBuffer` / `CharBuffer` was causing some de-optimization (after reading around on StackOverflow and seeing related discussions).

Did you happen to save those links? I'm curious what others have found about `*Buffer` performance. The javadocs make it sound like these are designed for good performance, so I'm a bit surprised about the performance problems you saw.

ahelwer:
How does the code handle three-byte codepoints?

fhackett:
@Calvin-L I lost the original thread, but I think this article might point out something similar: https://www.evanjones.ca/software/java-bytebuffers.html

Summary as I understand it: `ByteBuffer` is good for moving bytes, but may be slow to actually fill with bytes from Java, as compared to a byte array. Since our code is mostly filling the array with bytes, wrapping it with `ByteBuffer` would add overhead.

At least, I think. Below is a cross-check with other things I remember.

From a basic "what might that do to the JIT?" point of view, it also seems plausible that a class with lots of non-final methods and pluggability, vs native Java arrays and the final helper methods in the TLC code, would be harder to codegen well if you called those methods in a loop (so the method call overhead matters). Getting a pointer out of a *Buffer and then leaving the methods alone, however, not be subject to these pitfalls, hence possibly the fast bulk I/O story.

fhackett:
I made the change to remove backwards compatibility, and it turns out a test fails if you do that. The test is `tlc2.tool.liveness.CodePlexBug08EWD840FL2FromCheckpointTest`.

Good news: it comes with instructions on how to update the expected checkpoint .zip file.
Bad news: it involves JMX and "request TLC checkpoint to be taken", and my ~20min of guessing and looking around didn't get me very far.

Do you @lemmy  (asking because it sounds debugger-adjacent) have any more pointers on how to do this?
```
1) Run CodePlexBug08EWD840FL2Test with "-gzip"
2) Connect to running test via JMX and request TLC checkpoint to be taken
3) Terminate CodePlexBug08EWD840FL2Test once checkpoint successfully taken
4) Locate the directory with the checkpoint data
5) Replace the content of checkpoint.zip with the content of 4)
6) Update the number below on states found...
```

lemmy:
I'm in favor of deleting the test and removing support for checkpointing; I haven't used checkpointing in ages.

@fhackett 1. Download VisualVM 2. Install the VisualVM-MBeans plugin 3. Launch TLC 4. Connect VisualVM 5. MBeans > tlc2.tool.ModelChecker > Operations > checkpoint

fhackett:
@lemmy I got the plugin installed, but unfortunately I still can't get it to work.

If I try to run the test from Eclipse (stock latest IDE version I installed specifically to work on this), then none of the profiling + MBeans info is available. It's just blank, with all the normal tabs missing. VisualVM can connect to itself and I can see both profiling and the MBeans plugin working, however.

I can get the profiling UI to work if I launch TLC from the command-line as a .jar, but strangely then TLC runs so fast I can barely click anything. (and yes I set `-workers 1`) Maybe I reverse-engineered a path to the wrong TLA+ file...
The file I think the test runs is `EWD840MC2.tla`, for reference.

All that to say, I'm confused. I pushed my changes, but they fail that one test, and clearly I've messed up something in my setup for regenerating the file.

lemmy:
@fhackett Try setting a breakpoint in e.g. `Tool#getNextStates(final INextStateFunctor functor, final TLCState state)`

juliaferraioli:
> It seems reinstating our performance benchmark runs would be beneficial. @juliaferraioli, do you have any updates on the status of the sponsorship request I submitted a few months ago?

Apologies @lemmy. It got lost in my inbox. I'm checking on it and we should hear soon.

fhackett:
@lemmy Thanks for the note - it was a useful starting point to eliminate the timing possibility.

The problem was that I installed Eclipse as a Flatpak package. Somehow this broke VisualVM, but only just enough that it can "see" the JVM instances inside the Flatpak env but it can't actually query them, resulting in my strange experience. I reinstalled Eclipse and it works now.

That said, the quest is not over. Now I can follow the instructions, and one of 2 things happen:
1. It doesn't checkpoint. I tried this a bunch, and I think it only happened once or twice.
2. The checkpoint does not load, and not because of a problem with this PR (probably?).

I also tried hardcoding the checkpointing condition to `true` in `ModelChecker#doPeriodicWork()`, just to make absolutely sure a checkpoint would be generated. In that case, I get a checkpoint every time.

If I have checkpoint files at all, the error message I get from the resume test is this:
```
Starting recovery from checkpoint /home/finn/programming/tlaplus/tlatools/org.lamport.tlatools/test-model/CodePlexBug08/checkpoint/
The exception was a java.io.FileNotFoundException
: /home/finn/programming/tlaplus/tlatools/org.lamport.tlatools/test-model/CodePlexBug08/checkpoint/EWD840MC2-1.chkpt (No such file or directory)
```

It is correct; that checkpoint file does not exist. It was not created.

The actual checkpoint files are (with .dot files removed because one is 1GB):
```
$ ls checkpoint
dgraph_0.chkpt	   EWD840MC2-0.st  EWD840MC2.st.chkpt  queue.chkpt
EWD840MC2-0.chkpt  EWD840MC2_1.fp  nodes_0	       vars.chkpt
EWD840MC2_0.fp	   EWD840MC2.st    ptrs_0
```

So, it seems the `-0` is of interest, but it's looking for `-1` for reasons I can't understand. As far as I know, I am running unmodified checkpoint code, encoding aside, so this is strange. Does checkpointing work for you at all?

lemmy:
If checkpointing has been broken for a while and no one has raised an issue, it strongly suggests that the feature is no longer needed and should be removed.  That being said, during a recent TLC run, I saw that checkpoints were created. However, I did not attempt to restart from a checkpoint, as TLC completed successfully without crashing.

fhackett:
@lemmy I just tried the latest TLC release on a PGo system (so it runs long enough to take a 1min time-based checkpoint). It actually resumed fine.

So, _something_ is broken but it might be to do with this test, or these instructions to update the test, or something. I'll have a closer look. It might be that forcing a checkpoint is broken, but the feature itself is fine.

fhackett:
@lemmy Checkpointing definitely still works (though whether it is needed is an interesting question), but I found other misc stuff that was messing up the test. I fixed it, I think.

What happened:

1. The test was trying to run with a different worker count than the checkpoint. Somehow, that count was 3. I swear I saw that all tests had a worker count of 1, but who knows. I added the appropriate argument to ensure the correct count.
2. I never did get the JMX checkpoint trigger to work. It turns out my workaround of hardcoding the checkpoint condition to `true` was a bit too heavy-handed (maybe checkpointing on the final liveness test is broken, not sure... it is not surprising that I broke something by checkpointing _every possible moment_), so instead I initialized `TLCGlobals.forceChkpt = true`. That makes a very "early" checkpoint, but it works.
3. Because I changed the worker count, some other values in the test changed a bit. It looks like just churn (a file grew by ~1kb, when the test looks for "not enough", and 2 values in the liveness counter-example shifted), but I admit I didn't try to understand the meaning of the test beyond surface-level observations.

I summarized the important parts of the above in an addendum to the test update instructions.

If this commit passes all the tests, I'll flag the PR as ready for review. The feature works, and I think nothing is broken anymore.

tlatools/org.lamport.tlatools/src/tlc2/tool/queue/DiskByteArrayQueue.java
```diff
@@ -738,19 +739,19 @@ public final byte[] toByteArray() {
...
-				this.bytes[idx++] = (byte) c[i];
+			this.writeInt(str.length());
+			ensureCapacity(this.idx + str.length() * 2);
+			for(int i = 0; i < str.length(); i++) {
+				final char ch = str.charAt(i);
+				// big-endian
+				this.bytes[this.idx + i * 2] = (byte)((ch >> 8) & 0xff);
+				this.bytes[this.idx + i * 2 + 1] = (byte)(ch & 0xff);
 			}
+			this.idx += str.length() * 2;
```
Calvin-L, line 750: I know the Encoder/ByteBuffer approach made the tests slower, but I'm curious how this loop compares to a more canonical solution like

```java
			byte[] encodedBytes = str.getBytes(StandardCharsets.UTF_8);
			this.writeInt(encodedBytes.length);
			this.ensureCapacity(this.idx + encodedBytes.length);
			System.arraycopy(encodedBytes, 0, this.bytes, this.idx, encodedBytes.length);
			this.idx += encodedBytes.length;
```

Some part of me is still hopeful we can use UTF-8, since it is backward-compatible with TLC's current ASCII encoding and can be more compact on disk.

Calvin-L: (A _very_ preliminary benchmark I did on my laptop suggests that using `str.getBytes()` over a hand-rolled loop has very little impact on TLC's overall performance---but that result is far from definitive.)

fhackett: Agreed - that would be interesting to see. The optimized code I saw when I looked into the source of `getBytes` suggests the basically the only significant factor might be memory pressure, and it is quite possible that the JVM can optimize such a short-lived local object.

I think I found the benchmarks that @lemmy keeps mentioning: https://github.com/tlaplus/tlaplus/tree/master/tlatools/org.lamport.tlatools/test-benchmark. It looks like they test specifically the subject of this conversation - the disk queue - so it might be relatively straightforward to set up.

I should do some PhD work for a moment, but I can try that on my workstation (or on any other test machines we may or may not have) soon.

Calvin-L: There are also some end-to-end benchmarks under [general/performance/](https://github.com/tlaplus/tlaplus/tree/master/general/performance). Here is the script I use to compare two Git revisions (must be placed in and run from the root of the tlaplus repo):
https://gist.github.com/Calvin-L/227185db3f7f7131259f195978eccb61

fhackett: I used your script to benchmark the original ASCII encoding, UTF-8 with getBytes, and UTF-16.
I set it to run the full 600s per operation and left it for a while (though it was running on my personal machine so maybe not the cleanest benchmark).

Confusingly, using `getBytes` is noticeably more efficient than the UTF-16 solution. Both are worse than the ASCII version by some factor.

Here are the comparisons results.

before = ASCII, after = UTF-16:
```
Spec                 | W   | States/s (old)  | States/s (new)  | Change    
--------------------------------------------------------------------------------
Grid5k               |   2 |         2256543 |         2446747 |       1.08x FASTER
Ghostferry           |   2 |          130599 |          135148 |       1.03x FASTER
PaxosMadeSimple      |   8 |          119442 |          122801 |       1.03x FASTER
SwarmKit             |   4 |           15901 |           16108 |       1.01x FASTER
EWD840               |   4 |         3451637 |         3444705 |       1.00x SLOWER
MongoRepl            |   2 |          162027 |          160870 |       0.99x SLOWER
PaxosCommit          |   4 |          189351 |          187193 |       0.99x SLOWER
SwarmKit             |   2 |            9648 |            9517 |       0.99x SLOWER
PaxosMadeSimple      |   2 |           52861 |           49065 |       0.93x SLOWER
SwarmKit             |   8 |           26582 |           24314 |       0.91x SLOWER
Grid5k               |   4 |         3269984 |         2957318 |       0.90x SLOWER
EWD840               |   2 |         2204050 |         1965863 |       0.89x SLOWER
PaxosMadeSimple      |   4 |           89819 |           79475 |       0.88x SLOWER
PaxosCommit          |   2 |          151036 |          130710 |       0.87x SLOWER
MongoRepl            |   4 |          261426 |          221925 |       0.85x SLOWER
Bloemen              |   4 |          143643 |          115391 |       0.80x SLOWER
Bloemen              |   2 |          145588 |          115771 |       0.80x SLOWER
Ghostferry           |   8 |          231712 |          169007 |       0.73x SLOWER
MongoRepl            |   8 |          361961 |          263078 |       0.73x SLOWER
Ghostferry           |   4 |          217807 |          153329 |       0.70x SLOWER
Bookkeeper           |   4 |           67047 |           45208 |       0.67x SLOWER
Bookkeeper           |   2 |           58110 |           36756 |       0.63x SLOWER
Bloemen              |   8 |          184246 |          108148 |       0.59x SLOWER
PaxosCommit          |   8 |          306250 |          176044 |       0.57x SLOWER
Bookkeeper           |   8 |           77078 |           40276 |       0.52x SLOWER
```

before = ASCII, after = UTF-8 w/ getBytes
```
Spec                 | W   | States/s (old)  | States/s (new)  | Change    
--------------------------------------------------------------------------------
PaxosMadeSimple      |   8 |          119442 |          131574 |       1.10x FASTER
Ghostferry           |   2 |          130599 |          141453 |       1.08x FASTER
MongoRepl            |   2 |          162027 |          166811 |       1.03x FASTER
Grid5k               |   2 |         2256543 |         2319965 |       1.03x FASTER
SwarmKit             |   4 |           15901 |           16204 |       1.02x FASTER
EWD840               |   4 |         3451637 |         3511398 |       1.02x FASTER
SwarmKit             |   8 |           26582 |           26191 |       0.99x SLOWER
PaxosCommit          |   4 |          189351 |          183080 |       0.97x SLOWER
SwarmKit             |   2 |            9648 |            9169 |       0.95x SLOWER
PaxosMadeSimple      |   2 |           52861 |           49874 |       0.94x SLOWER
Bookkeeper           |   2 |           58110 |           54799 |       0.94x SLOWER
EWD840               |   2 |         2204050 |         2046023 |       0.93x SLOWER
PaxosMadeSimple      |   4 |           89819 |           82986 |       0.92x SLOWER
Grid5k               |   4 |         3269984 |         2985655 |       0.91x SLOWER
PaxosCommit          |   2 |          151036 |          137767 |       0.91x SLOWER
Ghostferry           |   8 |          231712 |          199879 |       0.86x SLOWER
MongoRepl            |   4 |          261426 |          216561 |       0.83x SLOWER
PaxosCommit          |   8 |          306250 |          250167 |       0.82x SLOWER
Bloemen              |   4 |          143643 |          112859 |       0.79x SLOWER
MongoRepl            |   8 |          361961 |          273302 |       0.76x SLOWER
Bookkeeper           |   8 |           77078 |           55919 |       0.73x SLOWER
Bookkeeper           |   4 |           67047 |           47630 |       0.71x SLOWER
Bloemen              |   2 |          145588 |          101009 |       0.69x SLOWER
Ghostferry           |   4 |          217807 |          146945 |       0.67x SLOWER
Bloemen              |   8 |          184246 |           94671 |       0.51x SLOWER
```

before = UTF-16, after = UTF-8 w/ getBytes
```
Spec                 | W   | States/s (old)  | States/s (new)  | Change    
--------------------------------------------------------------------------------
Bookkeeper           |   2 |           36756 |           54799 |       1.49x FASTER
PaxosCommit          |   8 |          176044 |          250167 |       1.42x FASTER
Bookkeeper           |   8 |           40276 |           55919 |       1.39x FASTER
Ghostferry           |   8 |          169007 |          199879 |       1.18x FASTER
SwarmKit             |   8 |           24314 |           26191 |       1.08x FASTER
PaxosMadeSimple      |   8 |          122801 |          131574 |       1.07x FASTER
PaxosCommit          |   2 |          130710 |          137767 |       1.05x FASTER
Bookkeeper           |   4 |           45208 |           47630 |       1.05x FASTER
Ghostferry           |   2 |          135148 |          141453 |       1.05x FASTER
PaxosMadeSimple      |   4 |           79475 |           82986 |       1.04x FASTER
EWD840               |   2 |         1965863 |         2046023 |       1.04x FASTER
MongoRepl            |   8 |          263078 |          273302 |       1.04x FASTER
MongoRepl            |   2 |          160870 |          166811 |       1.04x FASTER
EWD840               |   4 |         3444705 |         3511398 |       1.02x FASTER
PaxosMadeSimple      |   2 |           49065 |           49874 |       1.02x FASTER
Grid5k               |   4 |         2957318 |         2985655 |       1.01x FASTER
SwarmKit             |   4 |           16108 |           16204 |       1.01x FASTER
Bloemen              |   4 |          115391 |          112859 |       0.98x SLOWER
PaxosCommit          |   4 |          187193 |          183080 |       0.98x SLOWER
MongoRepl            |   4 |          221925 |          216561 |       0.98x SLOWER
SwarmKit             |   2 |            9517 |            9169 |       0.96x SLOWER
Ghostferry           |   4 |          153329 |          146945 |       0.96x SLOWER
Grid5k               |   2 |         2446747 |         2319965 |       0.95x SLOWER
Bloemen              |   8 |          108148 |           94671 |       0.88x SLOWER
Bloemen              |   2 |          115771 |          101009 |       0.87x SLOWER
```

I feel there's some more to read into which benchmarks are faster vs slower. It took over 24hrs to get the data though, so I'll put this here and come back to it.

fhackett: Here's a graph of the above:
image

I'm realizing now that error bars would be ideal (if I re-ran this 2x times, for 60hrs runtime total which is a lot of time to not really use my workstation for much), but it seems like the distinction between the 3 versions is unclear. I can imagine it's possible that the error bars are larger than the variation I'm seeing.

There may be some true performance regression, at least on a microbenchmark level, compared to the original (see this discussion, anecdotally https://stackoverflow.com/questions/12239993/why-is-the-native-string-getbytes-method-slower-than-the-custom-implemented-getb). That said, the linked discussion is actually wrong: getBytes has hardcoded special cases for UTF-8 and UTF-16 that have none of the drawbacks mentioned (at least, in my copy of the JDK's source code). I was concerned about the temporary array allocation causing problems, but perhaps the GC / JIT  (maybe escape analysis?) finds a way to make that irrelevant. That certainly seems to be the case, based on the resulting performance of TLC.

So, it seems inconclusive but maybe ok?
I'll commit the getBytes version for now, and if we want further study, then we can do that.

fhackett:
@lemmy I noticed that the CI hasn't run on this PR since my last commit, as it's waiting on a manual action of some sort.
I have some idea it was you pressing the button; sorry for bothering if not.

@Calvin-L Let me know if there are any other changes you have in mind (I assume you may be busy, and will get to it eventually).

Calvin-L:
Broadly I approve of the current state of this PR. (I will try to do a close re-read soon. I might have some small comments.) Even if there is a minor performance regression, I would still vote to merge this (to fix the underlying issue) and optimize later.

It seems like the big open question is whether we have to worry about a performance regression with this change. I spent yesterday running the benchmarks on my own beefy workstation, comparing ASCII (03201064b2772e9e6962db1f54e6ffb32430728a) to UTF-8 with getBytes (20998ac7d088de1202c8e1b62b13eb133a56d98e). I saw a lot less variance than you, which is promising---but I'm also not sure how much to trust the results. I ran out of disk space so some configurations didn't run. I will post my partial benchmark results tonight, along with some thoughts.

lemmy:
Slightly related to this PR: Impact of Unicode strings in TLA+ records (raised by Chris in yesterday's community call).

fhackett:
@Calvin-L Thanks for the update, that makes sense. Will wait for your further thoughts.

For context, my "workstation" is a System76 performance laptop with a 13th Gen Intel® Core™ i9-13900HX × 32 CPU. It is quite likely that laptop power management / asymmetric core performance added some noise you wouldn't see on a different system. I did of course take some precautions: leaving the laptop mostly alone, plugged in, and with sleep disabled.

Calvin-L:
Here are the results I got (Linux desktop, 24 (8+16) core Intel i9, 32GiB ram):

```
Spec                 | W   | States/s (old)  | States/s (new)  | Change    
--------------------------------------------------------------------------------
SwarmKit             |   8 |           39458 |           42046 |       1.07x FASTER
Bookkeeper           |   2 |          110291 |          113783 |       1.03x FASTER
SwarmKit             |   4 |           22359 |           23040 |       1.03x FASTER
Bookkeeper           |   8 |          317444 |          326725 |       1.03x FASTER
MongoRepl            |   2 |          219558 |          225698 |       1.03x FASTER
PaxosMadeSimple      |   8 |          211846 |          216078 |       1.02x FASTER
Bookkeeper           |   4 |          200370 |          204252 |       1.02x FASTER
MongoRepl            |   4 |          423768 |          431520 |       1.02x FASTER
PaxosMadeSimple      |   4 |          121735 |          123914 |       1.02x FASTER
SwarmKit             |   2 |           11851 |           12040 |       1.02x FASTER
PaxosCommit          |   8 |          635204 |          642420 |       1.01x FASTER
Ghostferry           |   2 |          169368 |          171226 |       1.01x FASTER
Grid5k               |   2 |         2529493 |         2554468 |       1.01x FASTER
PaxosMadeSimple      |   2 |           65971 |           66348 |       1.01x FASTER
PaxosCommit          |   4 |          371241 |          372862 |       1.00x FASTER
EWD840               |   2 |         2603488 |         2607569 |       1.00x FASTER
PaxosCommit          |   2 |          195533 |          195628 |       1.00x FASTER
MongoRepl            |   8 |          754276 |          753125 |       1.00x SLOWER
Bloemen              |   2 |          316673 |          315218 |       1.00x SLOWER
Bloemen              |   4 |          494553 |          490660 |       0.99x SLOWER
Bloemen              |   8 |          643722 |          638341 |       0.99x SLOWER
Ghostferry           |   4 |          272841 |          237019 |       0.87x SLOWER
Ghostferry           |   8 |          244055 |          187237 |       0.77x SLOWER
```

I did not get results for EWD840 or Grid5k because I ran out of disk space. Low disk space may have impacted the other numbers as well.

The numbers I got look completely normal except Ghostferry with 4+ cores. I am hoping those two runs are just outliers---although I have no evidence to support that theory. In your runs Ghostferry 4 and Ghostferry 8 were also losers:

```
Ghostferry           |   8 |          231712 |          199879 |       0.86x SLOWER
Ghostferry           |   4 |          217807 |          146945 |       0.67x SLOWER
```

But some of your other big losers like Bloemen and Bookkeeper were fine for me.

> error bars would be ideal 

I agree. As a back-of-the-envelope approximation, I'll point out that Grid5k does not use strings or records, and so it should be unaffected by this change. Thus it looks like your results might have error of +/- 10%. I can't play the same trick because I didn't really get results for Grid5k.

fhackett:
@Calvin-L 
> Grid5k does not use strings or records, and so it should be unaffected by this change.

Aren't states (`TLCState`) themselves made using `UniqueString` as well, for the variable identifiers?

That said, there seems to be no rhyme nor reason which examples have regressions vs. not.

I'm wondering if there's some kind of balance, where some benchmarks serialize states more often than others? Or, where serialization dominates in some cases but not others...
I don't know enough about how the state queue actually operates to think precisely, but a distinction of that sort could explain the pseudo-random seeming behavior we're seeing.

tlatools/org.lamport.tlatools/src/util/IDataInputStream.java
```diff
@@ -32,6 +32,7 @@ public interface IDataInputStream {
 
 	int readInt() throws IOException, EOFException;
 
-	String readString(int slen) throws IOException;
+	/** Reads a length-prefixed UTF-16 string **/
```
Calvin-L, line 35: Now UTF-8

tlatools/org.lamport.tlatools/src/util/IDataOutputStream.java
```diff
@@ -31,6 +31,7 @@ public interface IDataOutputStream {
 
 	void writeInt(int varLoc) throws IOException;
 
+	/** Write the string as length-prefixed UTF-16 **/
```
Calvin-L, line 34: Now UTF-8

Calvin-L (approved):

Calvin-L:
> Aren't states (`TLCState`) themselves made using `UniqueString` as well, for the variable identifiers?

States are actually represented as flat arrays of values. TLC decides on an ordering of the variable names at startup, and each `TLCState` object just has the values in that canonical order. The same is true of the on-disk representation.

https://github.com/tlaplus/tlaplus/blob/68b6f50f7e36b19dae03007620df928cb4d1b419/tlatools/org.lamport.tlatools/src/tlc2/tool/TLCStateMut.java#L323

fhackett:
> States are actually represented as flat arrays of values.

Ah, interesting. Thanks for clarifying. I guessed my interpretation based on a flag I saw somewhere to the effect of "is a state variable", attached to a string value type. Now I definitely have no idea what those perf results could mean.

I fixed the 2 stale comments you pointed out (whoops, thanks!), and updated the PR description with the latest findings.

Should I squash my commits, or does that happen automatically on merge?

fhackett:
@lemmy I just pushed a squashed version of everything here (assuming that's the desired end state). Can you run CI again please?

Let me know if there's anything else I should do here.

lemmy:
https://github.com/tlaplus/tlaplus/actions/runs/13021249349 is a run of the commits of this PR on our (new) benchmark infra (I had to push the branch [fhackett-fix-1076](https://github.com/fhackett/tlaplus/tree/fhackett-fix-1076) from @fhackett's fork [into this repo](https://github.com/tlaplus/tlaplus/tree/fhackett-fix-1076) for this to work).

fhackett:
@lemmy Thanks for setting that up!

I know all benchmarks are globally single threaded, so it could just be another one is running (I can't see that part of the dash), but should it still be pending 4+ hours later?

lemmy:
A run usually takes ~9h.  However, something is off with the automation because no runner picked up the job.  Will look into it.

lemmy:
I realized my mistake; I didn’t rebase your branch on top of the current master branch with the latest perf.yml. A new run has been initiated: https://github.com/tlaplus/tlaplus/actions/runs/13024985428/job/36332473408

fhackett:
After a little confusion about the CSV format (I didn't initially notice that each benchmark run appends to the same CSV file, meaning one file is a prefix of the other), I did some stats.

The perf results with the new benchmarking setup look like this
image

There are a few numbers in the data, but I figured that calculating states discovered over time would be a good measure of performance. I combined all recorded runs of the same (spec,workers) combo, and plotted the median value(s). The error values show the min/max values.

Assuming I got all my code right and I am reading the figure correctly, the only measurable performance loss is with Grid5K at 4 workers. Given it's not consistent with different workers, I'm not sure what that means.
I'm also not sure why UTF-8 comes out faster in other cases.

Any thoughts?

lemmy:
Looks to me as if the benchmarks are dominated by other code paths.  Have you considered writing a low-level JMH benchmark like I've done for other parts of TLC: https://github.com/tlaplus/tlaplus/tree/master/tlatools/org.lamport.tlatools/test-benchmark/tlc2

fhackett:
I hadn't considered that actually. I'll look at that "soon" (in that I should make more progress on a paper we're writing first).

lemmy:
@fhackett It looks like this PR may have stalled—are you planning to continue with it?

fhackett:
@lemmy Apologies for the slow reply - I've been stuck keeping up with other projects for some time. I've only just noticed that the PR is 1 year old almost, wow.

I had not forgotten though - let's keep this alive for another month, and I'll make a proper effort to either make progress or abandon.

(aside: I'm working on an unannounced project that does some interesting intrusive things with TLC, so I'm learning a decent amount)

PR #1085, opened 2024-11-22
Add support for new tlapm command-line syntax

glondu:


lemmy:
Follow-up of https://github.com/tlaplus/tlapm/pull/177

toolbox/org.lamport.tla.toolbox.tool.prover/src/org/lamport/tla/toolbox/tool/prover/job/ProverJob.java
```diff
@@ -763,6 +804,34 @@ private String[] constructCommand()
...
+
+        // WSL hack. See constructCommand above.
+        final boolean useWSL = tlapmPath.toOSString().contains("wsl.exe");
+        if (useWSL) {
+            command.add("tlapm");
+        }
+
+        command.add("--config");
+
+        return (String[]) command.toArray(new String[command.size()]);
```
Calvin-L, line 832: I don't think there's any need for an awkward array conversion here. It should be possible to simplify this to `return command;` if you make the method return `List<String>` instead of `String[]`. The only use of this method is as an argument to `new ProcessBuilder`, and [that constructor can take either `String[]` or `List<String>`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ProcessBuilder.html#%3Cinit%3E(java.util.List)).

glondu: Right. I just reproduced the constructCommand pattern already present in the same file. I will push a simplified version.

Calvin-L (approved):

lemmy:
IIRC, things changed over at the tlapm repo. What is the status of this PR?

ahelwer:
This is a good contribution but I would like to see the tests mentioned up above implemented to ensure we haven't broken backwards compatibility. If the original contributor does not want to write them I will get around to writing them myself eventually (in March probably?)

PR #1089, opened 2024-11-28
Java property 'nosymmetry' taken into account

Atafid:


Atafid:
Sorry I tried to open it in my own fork to run the CI tests but made a mistake.

PR #1099, opened 2024-12-13
Debugger: Add support for on demand breakpoint expressions

### LINKED ISSUE #1096: Adding incremental parsing to SANY
Collecting thoughts & discussion about this feature after the meeting this morning.

Currently, the only known method of using SANY is to parse an entire spec and its dependencies starting from source code in a file. However, there are two occasions where we want a more incremental approach:
1. To add to or update a definition in an existing parse tree without having to re-run the entire parse process, which would come in useful in the debugger and also interpreter.
2. To make parsing more responsive by only re-parsing the part of the file that was changed when a user edits a file, commonly seen in a language server scenario.

The second one is [*very difficult*](https://tratt.net/laurie/blog/2024/structured_editing_and_incremental_parsing.html), but the first one seems like it could be relatively straightforward. It is my (untested) belief that the changes in #1015, by removing some setup code that was only placed in the module parsing method, unlocked the ability to begin the parse process at the unit definition or even expression level. Separately, connected with my semantic testing work I've been figuring out the most minimal possible method of interacting with SANY's various parsing stages and got it down [to 100 lines of code or so](https://codeberg.org/tlaplus/simple-checker/src/branch/main/app/src/main/java/us/tlapl/Parser.java). This simplified SANY interface can be used to experiment with parsing & attaching a new definition to an existing spec context. I hope to experiment with this some time in the near future.

ahelwer:
Getting this out there so we can play around with it and build on the idea, see if it's even viable for the interpreter to use this operator definition, etc.

TODO:
- [ ] Conditionally restrict operator action level depending on breakpoint location
- [x] Implement functionality to find all variables local to the breakpoint scope and add them to the operator definition

Closes #1096

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/ModuleNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/ModuleNode.java
@@ -163,12 +163,6 @@ public class ModuleNode extends SymbolNode {
 *    an ExprNode[] array.                                                  *
 ***************************************************************************/
 
-  /**
-   * The {@link Generator} instance that performed the semantic analysis
-   * creating this {@link ModuleNode} instance.
-   */
-  public final Generator semanticChecker;
-
   private final Context      ctxt;
     // The (flat) context with all names known in this module, including
     // builtin ops, and ops declared as CONSTANT or VARIABLE, ops
@@ -331,9 +325,8 @@ public class ModuleNode extends SymbolNode {
     ***********************************************************************/
 
   // Invoked only in Generator
-  public ModuleNode(UniqueString us, Context ct, TreeNode stn, Generator gen) {
+  public ModuleNode(UniqueString us, Context ct, TreeNode stn) {
     super(ModuleKind, stn, us);
-    this.semanticChecker = gen;
     this.ctxt = ct;
   }
 
@@ -544,6 +537,22 @@ public class ModuleNode extends SymbolNode {
     return topLevel;
   }
 
+  /**
+   * Generates a plausible definition name that is not already in use in the
+   * context viewable from this module root.
+   *
+   * @param pattern A base pattern in which a number can be interpolated.
+   * @return A name that is unique within the context of the module.
+   */
+  public String generateUnusedName(String pattern) {
+    Context definedNames = this.getContext();
+    String unusedName = null;
+    do {
+      long suffix = System.currentTimeMillis();
+      unusedName = String.format(pattern, Long.toString(suffix));
+    } while (definedNames.occurSymbol(unusedName));
+    return unusedName;
+  }
 
   /**
  * @return the isInstantiated
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/ModuleNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/ModuleNode.java
@@ -163,6 +163,12 @@ public class ModuleNode extends SymbolNode {
 *    an ExprNode[] array.                                                  *
 ***************************************************************************/
 
+  /**
+   * The {@link Generator} instance that performed the semantic analysis
+   * creating this {@link ModuleNode} instance.
+   */
+  public final Generator semanticChecker;
+
   private final Context      ctxt;
     // The (flat) context with all names known in this module, including
     // builtin ops, and ops declared as CONSTANT or VARIABLE, ops
@@ -325,8 +331,9 @@ public class ModuleNode extends SymbolNode {
     ***********************************************************************/
 
   // Invoked only in Generator
-  public ModuleNode(UniqueString us, Context ct, TreeNode stn) {
+  public ModuleNode(UniqueString us, Context ct, TreeNode stn, Generator gen) {
     super(ModuleKind, stn, us);
+    this.semanticChecker = gen;
     this.ctxt = ct;
   }
 
@@ -1100,11 +1107,12 @@ final void addAssumption(TreeNode stn, ExprNode ass, SymbolTable st,
       if (children != null) {
           return children;
       }
+      OpDefNode[] opDefs = this.getOpDefs();
       children =
-         new SemanticNode[this.opDefs.length + this.topLevel.length];
+         new SemanticNode[opDefs.length + this.topLevel.length];
       int i;
-      for (i = 0; i < this.opDefs.length; i++) {
-          children[i] = this.opDefs[i];
+      for (i = 0; i < opDefs.length; i++) {
+          children[i] = opDefs[i];
       }
       for (int j = 0; j < this.topLevel.length; j++) {
           children[i+j] = this.topLevel[j];
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
@@ -231,14 +231,20 @@ public abstract class SemanticNode
 	}
 
 	/**
-	 * @return The path in the semantic tree up to the given {@link Location}
-	 *         starting from this node. An empty {@link LinkedList} if no reachable
-	 *         {@link SemanticNode} matches the given {@link Location}.
-	 *         <p>
-	 *         {@link LinkedList#getFirst()} is the {@link SemanticNode} matching
-	 *         {@link Location}.
+	 * Given a location, find the path from some node in this subtree at that
+	 * location up to this node.
+	 *
+	 * @param location Find a node in this subtree matching this location.
+	 * @param requireExactLocationMatch If true, only returns the path if a
+	 *          node spans the exact given location; if false allow paths
+	 *          starting from a leaf node that completely encompasses the
+	 *          given location.
+	 * @return The path in the semantic tree starting at the node matching
+	 *         the location and moving up through parents until reaching this
+	 *         node. An empty list is returned if no node in the subtree
+	 *         matches the given location constraints.
 	 */
-	public LinkedList<SemanticNode> pathTo(final Location location) {
+	public LinkedList<SemanticNode> pathTo(final Location location, boolean requireExactLocationMatch) {
 		final ChildrenVisitor<LinkedList<SemanticNode>> visitor = walkChildren(
 				new ChildrenVisitor<LinkedList<SemanticNode>>() {
 					LinkedList<SemanticNode> pathToLoc;
@@ -254,7 +260,10 @@ public abstract class SemanticNode
 					@Override
 					public void preVisit(final SemanticNode node) {
 						if (location.equals(node.getLocation())
-								|| (!node.hasChildren() && node.getLocation().includes(location))) {
+							|| (!requireExactLocationMatch
+								&& !node.hasChildren()
+								&& node.getLocation().includes(location))
+						) {
 							// node will be added to pathToLoc in postVisit!
 							pathToLoc = new LinkedList<>();
 						} else if (node instanceof OpDefNode) {
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
@@ -179,6 +179,9 @@ public abstract class SemanticNode
       return null;
   }
 
+  	public boolean hasChildren() {
+  		return !getListOfChildren().isEmpty();
+  	}
 	/**
 	 * @return Returns an empty list instead of null compared to getChildren.
 	 */
@@ -250,7 +253,8 @@ public abstract class SemanticNode
 
 					@Override
 					public void preVisit(final SemanticNode node) {
-						if (location.equals(node.getLocation()) || (0 == node.getChildren().length && node.getLocation().includes(location))) {
+						if (location.equals(node.getLocation())
+								|| (!node.hasChildren() && node.getLocation().includes(location))) {
 							// node will be added to pathToLoc in postVisit!
 							pathToLoc = new LinkedList<>();
 						} else if (node instanceof OpDefNode) {
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
@@ -231,6 +231,13 @@ public abstract class SemanticNode
 	}
 
 	/**
+	 * @see #pathTo(Location, Boolean)
+	 */
+	public LinkedList<SemanticNode> pathTo(final Location location) {
+		return pathTo(location, true);
+	}
+
+	/**
 	 * Given a location, find the path from some node in this subtree at that
 	 * location up to this node.
 	 *
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
@@ -250,7 +250,7 @@ public abstract class SemanticNode
 
 					@Override
 					public void preVisit(final SemanticNode node) {
-						if (location.equals(node.getLocation())) {
+						if (location.equals(node.getLocation()) || (0 == node.getChildren().length && node.getLocation().includes(location))) {
 							// node will be added to pathToLoc in postVisit!
 							pathToLoc = new LinkedList<>();
 						} else if (node instanceof OpDefNode) {
@@ -276,7 +276,8 @@ public abstract class SemanticNode
 					@Override
 					public boolean preempt(SemanticNode node) {
 						return pathToLoc != null
-								|| !node.getLocation().includes(location);
+								|| !node.getLocation()
+									.includes(location);
 					}
 
 					@Override
--- a/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
@@ -92,29 +92,35 @@ public class TLCBreakpointExpression {
 		SyntaxTreeNode syntaxRoot = parser.ParseTree;
 		if (!syntaxParseSuccess || null == syntaxRoot) {
 			// Parse error is output to ToolIO.out
+			ToolIO.err.println("Syntax error while parsing breakpoint expression \"" + conditionExpr + "\"");
 			return null;
 		}
 
 		Errors semanticLog = new Errors();
-		SemanticNode.setError(semanticLog); // Annoyingly static
+		SemanticNode.setError(semanticLog);
 		Generator semanticChecker = new Generator(processor.getModuleTbl(), semanticLog);
 		ModuleNode bpModule = null;
 		try {
 			bpModule = semanticChecker.generate(syntaxRoot);
 		} catch (AbortException e) {
 			ToolIO.err.print(e.toString());
+			ToolIO.err.println("Semantic error while parsing breakpoint expression \"" + conditionExpr + "\"");
 			return null;
 		}
 		if (null == bpModule || semanticLog.isFailure()) {
 			ToolIO.err.print(semanticLog.toString());
+			ToolIO.err.println("Semantic error while parsing breakpoint expression \"" + conditionExpr + "\"");
 			return null;
 		}
 
 		// Run level-checking. The operator should be restricted to
 		// action-level or below.
 		Errors levelCheckingErrors = new Errors();
+		SemanticNode.setError(levelCheckingErrors);
 		boolean levelCheckingSuccess = bpModule.levelCheck(levelCheckingErrors);
 		if (!levelCheckingSuccess || levelCheckingErrors.isFailure() || !bpModule.levelCorrect) {
+			ToolIO.err.println(levelCheckingErrors.toString());
+			ToolIO.err.println("Level-checking error while parsing breakpoint expression \"" + conditionExpr + "\"");
 			return null;
 		}
 		
--- a/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
@@ -1,20 +1,54 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Linux Foundation. All rights reserved.
+ *
+ * The MIT License (MIT)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+ * of the Software, and to permit persons to whom the Software is furnished to do
+ * so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN
+ * AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * Contributors:
+ *   Andrew Helwer - code interfacing with SANY
+ *   Markus Alexander Kuppe - code interfacing with TLC interpreter
+ ******************************************************************************/
 package tlc2.debug;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import tla2sany.parser.SyntaxTreeNode;
 import tla2sany.parser.TLAplusParser;
 import tla2sany.semantic.AbortException;
 import tla2sany.semantic.Context;
 import tla2sany.semantic.Errors;
+import tla2sany.semantic.FormalParamNode;
 import tla2sany.semantic.Generator;
+import tla2sany.semantic.LetInNode;
 import tla2sany.semantic.LevelConstants;
 import tla2sany.semantic.ModuleNode;
+import tla2sany.semantic.OpApplNode;
+import tla2sany.semantic.OpArgNode;
 import tla2sany.semantic.OpDefNode;
 import tla2sany.semantic.SemanticNode;
+import tla2sany.semantic.SymbolNode;
+import tla2sany.st.Location;
 import tlc2.tool.impl.SpecProcessor;
 import util.ToolIO;
 
@@ -26,13 +60,19 @@ public class TLCBreakpointExpression {
 	/**
 	 * Given a spec and an unparsed expression, build an operator that can be
 	 * evaluated within the context of that spec.
-	 * 
+	 *
+	 * @param processor     Processes expression within the model context.
 	 * @param semanticRoot  The root of the spec's semantic parse tree.
-	 * @param conditionExpr The unparsed expression.
+	 * @param location      The breakpoint location.
+	 * @param conditionExpr The unparsed breakpoint expression.
 	 * @return A breakpoint expression, or null if parsing failed.
 	 */
-	public static OpDefNode process(final SpecProcessor processor, final ModuleNode semanticRoot,
-			final String conditionExpr) {
+	public static OpDefNode process(
+			final SpecProcessor processor,
+			final ModuleNode semanticRoot,
+			final Location location,
+			final String conditionExpr
+	) {
 		if (null == conditionExpr || conditionExpr.isBlank()) {
 			return null;
 		}
@@ -42,10 +82,13 @@ public class TLCBreakpointExpression {
 		final String bpModName = generateUnusedName(semanticRoot, "__BreakpointModule__%s");
 		ToolIO.out.println("BPExpr: wrapping with module \"" + bpModName + "\"");
 		final String bpOpName = generateUnusedName(semanticRoot, "__BreakpointExpr__%s");
-		ToolIO.out.println("BPExpr: wrapping with op \"" + bpOpName + "\"");
+		final Set<String> paramNames = getScopedIdentifiers(semanticRoot, location);
+		final String params = paramNames.size() > 0 ? "(" + String.join(", ", paramNames) + ")" : "";
+		final String bpOpDef = bpOpName + params;
+		ToolIO.out.println("BPExpr: wrapping with op \"" + bpOpDef + "\"");
 
 		final String wrapper = "---- MODULE %s ----\nEXTENDS %s\n%s == %s\n====";
-		String wrappedConditionExpr = String.format(wrapper, bpModName, rootModName, bpOpName, conditionExpr);
+		String wrappedConditionExpr = String.format(wrapper, bpModName, rootModName, bpOpDef, conditionExpr);
 		byte[] wrappedConditionExprBytes = wrappedConditionExpr.getBytes(StandardCharsets.UTF_8);
 		InputStream sourceCode = new ByteArrayInputStream(wrappedConditionExprBytes);
 		TLAplusParser parser = new TLAplusParser(sourceCode, StandardCharsets.UTF_8.name());
@@ -65,7 +108,6 @@ public class TLCBreakpointExpression {
 			bpModule = semanticChecker.generate(syntaxRoot);
 		} catch (AbortException e) {
 			ToolIO.err.print(e.toString());
-			ToolIO.err.print(semanticRoot.semanticChecker.errors.toString());
 			return null;
 		}
 		if (null == bpModule || semanticLog.isFailure()) {
@@ -89,10 +131,10 @@ public class TLCBreakpointExpression {
 			ToolIO.err.println("ERROR: unable to find breakpoint expression op " + bpOpName);
 		}
 
-		if (!(LevelConstants.ConstantLevel == bpOp.level || LevelConstants.VariableLevel == bpOp.level
-				|| LevelConstants.ActionLevel == bpOp.level)) {
+		if (!(LevelConstants.ConstantLevel == bpOp.getLevel() || LevelConstants.VariableLevel == bpOp.getLevel()
+				|| LevelConstants.ActionLevel == bpOp.getLevel())) {
 			ToolIO.err.println("ERROR: Debug expressions must be action-level or below; actual level: "
-					+ Integer.toString(bpOp.level));
+					+ SemanticNode.levelToString(bpOp.getLevel()));
 			return null;
 		}
 
@@ -102,13 +144,19 @@ public class TLCBreakpointExpression {
 		return bpOp;
 	}
 	
+	/**
+	 * Generates a plausible definition name that is not already in use.
+	 *
+	 * @param semanticRoot The module in which to check for name collisions.
+	 * @param pattern A base pattern in which a number can be interpolated.
+	 * @return A name that is unique within the context of the module.
+	 */
 	private static String generateUnusedName(ModuleNode semanticRoot, String pattern) {
 		Context definedNames = semanticRoot.getContext();
 		String unusedName = null;
-		long suffix = 0L;
 		do {
+			long suffix = System.currentTimeMillis();
 			unusedName = String.format(pattern, Long.toString(suffix));
-			suffix++;
 		} while (definedNames.occurSymbol(unusedName));
 		return unusedName;
 	}
@@ -145,14 +193,77 @@ public class TLCBreakpointExpression {
 	 *     F(i, j) (* want to add breakpoint here *)
 	 * VARIABLE i
 	 *
-	 * Since when we add our auto-generated operator to the module it is
+	 * Since when we extend the module so as to define our generated op it is
 	 * implicitly parsed at the end of the file, "__BreakpointExpr__0(i, j)"
 	 * will result in a parse error due to multiple "i" definitions.
 	 * Currently these types of specs are simply not supported. Supporting
-	 * them would probably require changes to the semantic analysis code
-	 * enabling parsing "ghost" definitions at arbitrary points in the tree.
+	 * them would probably require changes to the semantic analysis and
+	 * constants-resolver code enabling parsing "ghost" definitions at
+	 * arbitrary points in the tree.
+	 *
+	 * The way this method works is it first finds the parse tree node where
+	 * the breakpoint was introduced, then traces its parentage up to the
+	 * module root while recording identifiers produced by:
+	 *  - LET/IN blocks
+	 *  - Quantification operators
+	 *  - Operator parameters
+	 *
+	 * There are probably others but we will add to this code as the need is
+	 * discovered. A better solution would be to record this info in the
+	 * semantic nodes during the semantic checking process but that will have
+	 * to wait until we have good test coverage of it. It should be noted
+	 * this information is already sort of recorded in the (private)
+	 * OpApplNode.allParams field, but only if the given expression actually
+	 * references an outside parameter. By the time the semantic analysis
+	 * process has actually reached that stage it's too late to append the
+	 * references to the operator definition node.
+	 *
+	 * @param semanticRoot The root node of the semantic parse tree.
+	 * @param location     The breakpoint location.
+	 * @return A list of local identifiers accessible at the given location.
 	 */
-	static List<String> getScopedIdentifiers() {
-		return null;
+	static Set<String> getScopedIdentifiers(
+			ModuleNode semanticRoot,
+			Location location
+	) {
+		Set<String> identifiers = new HashSet<String>();
+		// pathTo starts at breakpoint location then goes up to module root
+		List<SemanticNode> path = semanticRoot.pathTo(location);
+		for (SemanticNode current : path) {
+			// Extract i from LET i == 5 IN ...
+			if (current instanceof LetInNode) {
+				LetInNode node = (LetInNode)current;
+				for (OpDefNode def : node.getLets()) {
+					identifiers.add(def.getName().toString());
+				}
+			// Extract op, i, j, k from op(i, j, k) == ...
+			// Note: will not extract "op" if is top-level definition
+			} else if (current instanceof OpDefNode) {
+				OpDefNode node = (OpDefNode)current;
+				if (null == semanticRoot.getOpDef(node.getName())) {
+					identifiers.add(node.getName().toString());
+				}
+				for (FormalParamNode param : node.getParams()) {
+					identifiers.add(param.getName().toString());
+				}
+			// Extract i, j from \A i, j \in Nat : ...
+			} else if (current instanceof OpApplNode) {
+				OpApplNode node = (OpApplNode)current;
+				for (FormalParamNode param : node.getQuantSymbolLists()) {
+					identifiers.add(param.getName().toString());
+				}
+			// Extract i, j from LAMBDA i, j : ...
+			} else if (current instanceof OpArgNode) {
+				OpArgNode node = (OpArgNode)current;
+				SymbolNode opSymbol = node.getOp();
+				if (opSymbol instanceof OpDefNode) {
+					OpDefNode op = (OpDefNode)opSymbol;
+					for (FormalParamNode param : op.getParams()) {
+						identifiers.add(param.getName().toString());
+					}
+				}
+			}
+		}
+		return identifiers;
 	}
 }
--- a/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023 Linux Foundation. All rights reserved.
+ * Copyright (c) 2024 Linux Foundation. All rights reserved.
  *
  * The MIT License (MIT)
  *
@@ -36,7 +36,6 @@ import java.util.Set;
 import tla2sany.parser.SyntaxTreeNode;
 import tla2sany.parser.TLAplusParser;
 import tla2sany.semantic.AbortException;
-import tla2sany.semantic.Context;
 import tla2sany.semantic.Errors;
 import tla2sany.semantic.FormalParamNode;
 import tla2sany.semantic.Generator;
@@ -76,16 +75,13 @@ public class TLCBreakpointExpression {
 		if (null == conditionExpr || conditionExpr.isBlank()) {
 			return null;
 		}
-		ToolIO.out.println("BPExpr: processing expression \"" + conditionExpr + "\"");
 
 		final String rootModName = semanticRoot.getName().toString();
-		final String bpModName = generateUnusedName(semanticRoot, "__BreakpointModule__%s");
-		ToolIO.out.println("BPExpr: wrapping with module \"" + bpModName + "\"");
-		final String bpOpName = generateUnusedName(semanticRoot, "__BreakpointExpr__%s");
+		final String bpModName = semanticRoot.generateUnusedName("__BreakpointModule__%s");
+		final String bpOpName = semanticRoot.generateUnusedName("__BreakpointExpr__%s");
 		final Set<String> paramNames = getScopedIdentifiers(semanticRoot, location);
 		final String params = paramNames.size() > 0 ? "(" + String.join(", ", paramNames) + ")" : "";
 		final String bpOpDef = bpOpName + params;
-		ToolIO.out.println("BPExpr: wrapping with op \"" + bpOpDef + "\"");
 
 		final String wrapper = "---- MODULE %s ----\nEXTENDS %s\n%s == %s\n====";
 		String wrappedConditionExpr = String.format(wrapper, bpModName, rootModName, bpOpDef, conditionExpr);
@@ -98,11 +94,10 @@ public class TLCBreakpointExpression {
 			// Parse error is output to ToolIO.out
 			return null;
 		}
-		ToolIO.out.println("BPExpr: success [syntactic analysis]");
 
 		Errors semanticLog = new Errors();
 		SemanticNode.setError(semanticLog); // Annoyingly static
-		Generator semanticChecker = new Generator(semanticRoot.semanticChecker.moduleTable, semanticLog);
+		Generator semanticChecker = new Generator(processor.getModuleTbl(), semanticLog);
 		ModuleNode bpModule = null;
 		try {
 			bpModule = semanticChecker.generate(syntaxRoot);
@@ -114,21 +109,19 @@ public class TLCBreakpointExpression {
 			ToolIO.err.print(semanticLog.toString());
 			return null;
 		}
-		ToolIO.out.println("BPExpr: success [semantic analysis]");
 
 		// Run level-checking. The operator should be restricted to
 		// action-level or below.
 		Errors levelCheckingErrors = new Errors();
 		boolean levelCheckingSuccess = bpModule.levelCheck(levelCheckingErrors);
 		if (!levelCheckingSuccess || levelCheckingErrors.isFailure() || !bpModule.levelCorrect) {
-			ToolIO.err.println(levelCheckingErrors.toString());
 			return null;
 		}
-		ToolIO.out.println("BPExpr: success [level analysis]");
 		
 		OpDefNode bpOp = bpModule.getOpDef(bpOpName);
 		if (null == bpOp) {
 			ToolIO.err.println("ERROR: unable to find breakpoint expression op " + bpOpName);
+			return null;
 		}
 
 		if (!(LevelConstants.ConstantLevel == bpOp.getLevel() || LevelConstants.VariableLevel == bpOp.getLevel()
@@ -140,25 +133,8 @@ public class TLCBreakpointExpression {
 
 		processor.processConstantsDynamicExtendee(bpModule);
 		
-		ToolIO.out.println("BPExpr: integration complete");
+		ToolIO.out.println("Processed breakpoint expression \"" + conditionExpr + "\"");
 		return bpOp;
-	}
-	
-	/**
-	 * Generates a plausible definition name that is not already in use.
-	 *
-	 * @param semanticRoot The module in which to check for name collisions.
-	 * @param pattern A base pattern in which a number can be interpolated.
-	 * @return A name that is unique within the context of the module.
-	 */
-	private static String generateUnusedName(ModuleNode semanticRoot, String pattern) {
-		Context definedNames = semanticRoot.getContext();
-		String unusedName = null;
-		do {
-			long suffix = System.currentTimeMillis();
-			unusedName = String.format(pattern, Long.toString(suffix));
-		} while (definedNames.occurSymbol(unusedName));
-		return unusedName;
 	}
 	
 	/**
--- a/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
@@ -228,7 +228,7 @@ public class TLCBreakpointExpression {
 	) {
 		Set<String> identifiers = new HashSet<String>();
 		// pathTo starts at breakpoint location then goes up to module root
-		List<SemanticNode> path = semanticRoot.pathTo(location);
+		List<SemanticNode> path = semanticRoot.pathTo(location, false);
 		for (SemanticNode current : path) {
 			// Extract i from LET i == 5 IN ...
 			if (current instanceof LetInNode) {
--- a/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
@@ -1,0 +1,158 @@
+package tlc2.debug;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+import tla2sany.parser.SyntaxTreeNode;
+import tla2sany.parser.TLAplusParser;
+import tla2sany.semantic.AbortException;
+import tla2sany.semantic.Context;
+import tla2sany.semantic.Errors;
+import tla2sany.semantic.Generator;
+import tla2sany.semantic.LevelConstants;
+import tla2sany.semantic.ModuleNode;
+import tla2sany.semantic.OpDefNode;
+import tla2sany.semantic.SemanticNode;
+import tlc2.tool.impl.SpecProcessor;
+import util.ToolIO;
+
+/**
+ * An expression for conditionally triggering a breakpoint.
+ */
+public class TLCBreakpointExpression {
+
+	/**
+	 * Given a spec and an unparsed expression, build an operator that can be
+	 * evaluated within the context of that spec.
+	 * 
+	 * @param semanticRoot  The root of the spec's semantic parse tree.
+	 * @param conditionExpr The unparsed expression.
+	 * @return A breakpoint expression, or null if parsing failed.
+	 */
+	public static OpDefNode process(final SpecProcessor processor, final ModuleNode semanticRoot,
+			final String conditionExpr) {
+		if (null == conditionExpr || conditionExpr.isBlank()) {
+			return null;
+		}
+		ToolIO.out.println("BPExpr: processing expression \"" + conditionExpr + "\"");
+
+		final String rootModName = semanticRoot.getName().toString();
+		final String bpModName = generateUnusedName(semanticRoot, "__BreakpointModule__%s");
+		ToolIO.out.println("BPExpr: wrapping with module \"" + bpModName + "\"");
+		final String bpOpName = generateUnusedName(semanticRoot, "__BreakpointExpr__%s");
+		ToolIO.out.println("BPExpr: wrapping with op \"" + bpOpName + "\"");
+
+		final String wrapper = "---- MODULE %s ----\nEXTENDS %s\n%s == %s\n====";
+		String wrappedConditionExpr = String.format(wrapper, bpModName, rootModName, bpOpName, conditionExpr);
+		byte[] wrappedConditionExprBytes = wrappedConditionExpr.getBytes(StandardCharsets.UTF_8);
+		InputStream sourceCode = new ByteArrayInputStream(wrappedConditionExprBytes);
+		TLAplusParser parser = new TLAplusParser(sourceCode, StandardCharsets.UTF_8.name());
+		boolean syntaxParseSuccess = parser.parse();
+		SyntaxTreeNode syntaxRoot = parser.ParseTree;
+		if (!syntaxParseSuccess || null == syntaxRoot) {
+			// Parse error is output to ToolIO.out
+			return null;
+		}
+		ToolIO.out.println("BPExpr: success [syntactic analysis]");
+
+		Errors semanticLog = new Errors();
+		SemanticNode.setError(semanticLog); // Annoyingly static
+		Generator semanticChecker = new Generator(semanticRoot.semanticChecker.moduleTable, semanticLog);
+		ModuleNode bpModule = null;
+		try {
+			bpModule = semanticChecker.generate(syntaxRoot);
+		} catch (AbortException e) {
+			ToolIO.err.print(e.toString());
+			ToolIO.err.print(semanticRoot.semanticChecker.errors.toString());
+			return null;
+		}
+		if (null == bpModule || semanticLog.isFailure()) {
+			ToolIO.err.print(semanticLog.toString());
+			return null;
+		}
+		ToolIO.out.println("BPExpr: success [semantic analysis]");
+
+		// Run level-checking. The operator should be restricted to
+		// action-level or below.
+		Errors levelCheckingErrors = new Errors();
+		boolean levelCheckingSuccess = bpModule.levelCheck(levelCheckingErrors);
+		if (!levelCheckingSuccess || levelCheckingErrors.isFailure() || !bpModule.levelCorrect) {
+			ToolIO.err.println(levelCheckingErrors.toString());
+			return null;
+		}
+		ToolIO.out.println("BPExpr: success [level analysis]");
+		
+		OpDefNode bpOp = bpModule.getOpDef(bpOpName);
+		if (null == bpOp) {
+			ToolIO.err.println("ERROR: unable to find breakpoint expression op " + bpOpName);
+		}
+
+		if (!(LevelConstants.ConstantLevel == bpOp.level || LevelConstants.VariableLevel == bpOp.level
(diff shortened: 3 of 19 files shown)


tlatools/org.lamport.tlatools/src/tla2sany/semantic/ModuleNode.java
```diff
@@ -163,6 +163,12 @@ public class ModuleNode extends SymbolNode {
 *    an ExprNode[] array.                                                  *
 ***************************************************************************/
 
+  /**
+   * The {@link Generator} instance that performed the semantic analysis
+   * creating this {@link ModuleNode} instance.
+   */
+  public final Generator SemanticChecker;
```
lemmy, line 170: Nit: `semanticChecker`

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
```diff
@@ -55,8 +68,95 @@ public TLCSourceBreakpoint(final String module, final SourceBreakpoint s) {
...
+		do {
+			bpOpName = String.format(candidatePattern, Long.toString(suffix));
+			suffix++;
+		} while (semanticRoot.getOpDef(bpOpName) != null);
+
+		// Then, send expression through syntax parser; wrap expression
+		// in module to make syntax parser happy, then throw away the
+		// encapsulating module. Future syntax parser updates might allow
+		// parsing the operator definition directly.
+		final String wrapper = "---- MODULE Ignore ----\n%s == %s\n====";
```
lemmy, line 100: Perhaps, e.g. `Ignore21384ylshfutfaoug2` because `Ignore` could also be a user spec.

ahelwer: The Ignore name will never make it to semantic checker so it won't be checked for conflicts; the syntax parser doesn't check for name collisions.

ahelwer:
Also thought of a tricky problem. Implicitly, the new operator is parsed at the very bottom of the file; however, consider this spec:
```tla
---- MODULE Test ----
op(x) == (* want to debug here *) TRUE
VARIABLE x
====
```
We would generate an operator with definition `__BreakpointExpr__0(x) == ...` but this would then cause a semantic failure since `x` is already defined in the global scope by the end of the file. Something to think about. It might be that in addition to generating a unique `__BreakpointExpr__n` name for the operator, we also need to generate unique names for the parameters, ending up with something like `__BreakpointExpr__0(__bpParam__x) == ...` then when resolving the breakpoint we extract "`x`" from `__bp_Param__x` and plug that name into the stack frame.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
```diff
@@ -55,8 +68,95 @@ public TLCSourceBreakpoint(final String module, final SourceBreakpoint s) {
...
+			ToolIO.err.print(semanticRoot.SemanticChecker.errors.toString());
+			return null;
+		}
+		semanticRoot.clearOpDefCache();
+		
+		// Run level-checking. The operator should be restricted to
+		// action-level or below.
+		Errors levelCheckingErrors = new Errors();
+		SemanticNode.setError(levelCheckingErrors); // Annoyingly static
+		if (!semanticOp.levelCheck(1) || levelCheckingErrors.isFailure()) {
```
lemmy, line 152: Consider adding a zero-arity `LevelNode#levelCheck` that invokes `LevelNode#levelCheck` with `LevelNode#levelChecked + 1`

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
lemmy: Please add a corresponding note to the comment.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
```diff
@@ -55,8 +69,105 @@ public TLCSourceBreakpoint(final String module, final SourceBreakpoint s) {
...
+		
+		// Run level-checking. The operator should be restricted to
+		// action-level or below.
+		Errors levelCheckingErrors = new Errors();
+		SemanticNode.setError(levelCheckingErrors); // Annoyingly static
+		if (!semanticOp.levelCheck() || levelCheckingErrors.isFailure() || !semanticOp.levelCorrect) {
+			ToolIO.err.println(levelCheckingErrors.toString());
+			return null;
+		}
+		if (LevelConstants.ConstantLevel == semanticOp.level
```
lemmy, line 157: `tla2sany.semantic.LevelNode.getLevel()`

More importantly, can you make the context of the breakpoint determine the maximum level of the debug expression? For instance, the maximum level of a breakpoint expression for a constant formula is constant, and so on.

Related:
```java
tla2sany.semantic.SemanticNode.pathTo(Location)
```

tlatools/org.lamport.tlatools/src/tla2sany/semantic/ModuleNode.java
```diff
@@ -419,6 +426,20 @@ public final OpDefNode[] getOpDefs() {
...
+   * Function {@link ModuleNode#getOpDefs} creates a cache of all operator
+   * definitions on first call. While this was fine when we could expect SANY
+   * to never have its parse tree modified after original parsing, now we
+   * occasionally want to insert a definition or expression & evaluate it in
+   * the context of an existing spec without having to restart the entire
+   * parse process. Once we do that we need to invalidate the cache so the
+   * definition can be surfaced when the cache is rebuilt; that is what this
+   * function does.
+   */
+  public void clearOpDefCache() {
```
lemmy, line 440: Are you certain that the underlying `ModuleNode#ctxt` cannot be mutate after `getOpDefs` is invoked the first time by legacy code?

ahelwer: I'm not sure; this is something to explore with this prototype. The role of the context class is still a bit mysterious to me and I don't know how it connects to other things in the parser. It is home to a piece of static global state I want to get rid of though, so understanding it is hopefully in my future.

ahelwer:
@lemmy all right I'm at the end of my expertise here, I have a test setup and have debugged it all the way to where it constructs the context to evaluate the breakpoint expression. Unfortunately I get an empty context no matter what, it doesn't even contain the global variable and constant values. For now I'm just testing expressions that depend only on global identifiers to see whether this works at all before investing time on doing the whole local identifier analysis thing.

lemmy:
> @lemmy all right I'm at the end of my expertise here, I have a test setup and have debugged it all the way to where it constructs the context to evaluate the breakpoint expression. Unfortunately I get an empty context no matter what, it doesn't even contain the global variable and constant values. For now I'm just testing expressions that depend only on global identifiers to see whether this works at all before investing time on doing the whole local identifier analysis thing.

Are you saying that `tlc2.debug.TLCStackFrame#ctxt` is always empty?  Did you try with *inline* breakpoints?

Related: 4a2586e2b24afdf633cf3c28c4900449f9793656

ahelwer:
Well in my test I have this module:
```tla
---- MODULE ExpressionBreakpointTest ----
EXTENDS Naturals
VARIABLE i
CONSTANT j
op(k) ==
  \A l \in 0 .. k :
    LET m == i IN
      (* breakpoint 1 here *) TRUE
(* breakpoint 2 here *) Inv == op(i)
Init == i = 0
Next == IF i < j THEN i' = i + 1 ELSE UNCHANGED i
====
---- CONFIG ExpressionBreakpointTest ----
CONSTANT j = 10
INVARIANT Inv
INIT Init
NEXT Next
====
```
and yeah when I run the debugger and it hits breakpoint 1 the context I get from `getContext()` is empty.

ahelwer:
Okay so the Context seems to only include local identifier values, not global identifier values. How do I access global identifier values?

ahelwer:
The evaluator access state-level variables through the state; the missing part now is the value of constants. Are constant values inlined in TLA+ expressions? If so will need to perform that on the user-supplied breakpoint expression.

lemmy:
> The evaluator access state-level variables through the state; the missing part now is the value of constants. Are constant values inlined in TLA+ expressions? If so will need to perform that on the user-supplied breakpoint expression.

There was no need to explicitly handle variables and constants with the “pre-startup” or “hardcoded” breakpoint expressions because `tlc2.tool.impl.SpecProcessor` automatically handled them. However, `tlc2.tool.impl.SymbolNodeValueLookupProvider` should be able to handle most, if not all, of the job. Alternatively, `tlc2.tool.impl.SpecProcessor.getConstantDefs()` could be relevant.

ahelwer:
There's a missing link somewhere unfortunately. Even just trying to evaluate the breakpoint expression `4 > 3` results in the following error:
>tlc2.tool.EvalException: Attempted to apply the operator overridden by the Java method public static tlc2.value.impl.BoolValue tlc2.module.Integers.GT(tlc2.value.impl.Value,tlc2.value.impl.Value), but it produced the following error:
Cannot invoke "tlc2.value.impl.Value.toString()" because "x" is null

So it seems just running it through the semantic analysis engine isn't enough, there is some additional expression processing that needs to happen before it can be evaluated.

lemmy:
You will have to process the `Ignore` module with `SpecProcessor#processConstants`.

https://github.com/tlaplus/tlaplus/blob/68b6f50f7e36b19dae03007620df928cb4d1b419/tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java#L1769-L1834

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
```diff
@@ -55,8 +58,9 @@ public TLCSourceBreakpoint(final String module, final SourceBreakpoint s) {
 			}
 		}
 		hits = h;
+		this.condition = TLCBreakpointExpression.process(semanticRoot, s.getCondition());
```
lemmy, line 61: This needs to check if s.getCondition is an existing definition such as `TypeOK`, ... that doesn't need processing.

lemmy:
> You will have to process the `Ignore` module (or the `semanticOp`) with `SpecProcessor#processConstants`.

Processing the breakpoint expression within the scope of the `semanticRoot` is likely the most suitable approach. However, the `SpecProcessor` is unable to incrementally process the `semanticRoot`, and enhancing it would be highly error-prone (a debugger's design guideline is not to modify the existing interpreter). Therefore, I propose attempting to make the `Ignore` module an *inner* module of the `semanticRoot`.

lemmy:
More debugging indicates that the inner module approach may not be sufficient (the inner module doesn’t appear to be encapsulated within its own `ModuleNode`). If confirmed, `Ignore` could also extend `semanticRoot`.

ahelwer:
>Processing the breakpoint expression within the scope of the semanticRoot is likely the most suitable approach. However, the SpecProcessor is unable to incrementally process the semanticRoot, and enhancing it would be highly error-prone

We could punt on this work until after I'm finished with the semantic test corpus work, which would give us more confidence trying to modify it to accept incremental processes of this type.

ahelwer:
>More debugging indicates that the inner module approach may not be sufficient (the inner module doesn’t appear to be encapsulated within its own ModuleNode). If confirmed, Ignore could also extend semanticRoot.

That could be another possible approach. It should be possible to find & iterate through the `ExternalModuleTable` to ensure generation of a module name that does not conflict with any others.

ahelwer:
>More debugging indicates that the inner module approach may not be sufficient (the inner module doesn’t appear to be encapsulated within its own ModuleNode). If confirmed, Ignore could also extend semanticRoot.

I've now implemented this approach. However, I don't know how to easily construct a `SpecProcessor` instance with the state available while defining a new breakpoint; do you know what would need to be pulled in for that?

lemmy:
Why don't you reuse the existing `tool.getSpecProcessor` instance?

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
```diff
@@ -0,0 +1,161 @@
...
+		return bpOp;
+	}
+	
+	private static String generateUnusedName(ModuleNode semanticRoot, String pattern) {
+		Context definedNames = semanticRoot.getContext();
+		String unusedName = null;
+		long suffix = 0L;
+		do {
+			unusedName = String.format(pattern, Long.toString(suffix));
+			suffix++;
```
lemmy, line 114: Minor: The likelihood of a collision is significantly reduced if the suffix is `System.currentTimeMillis()`, which aligns with the approach employed by the Toolbox.

lemmy:
The TLA+ Debugger also supports "Watch" expressions.  AFAICT, Watch and Breakpoint expressions will be able to share most of the implementation.

https://www.youtube.com/watch?v=r7kGbjcSaKU

lemmy:
The branch `mku-pr1099` is based on this pull request and modifies the `SpecProcessor` to process breakpoint expressions correctly. Specifically, `SpecProcessor` now takes another snapshot of the `SpecProcessor#defns` before module overrides, constants definitions, ... are processed. This modification allows context-free breakpoint expressions, such as those found in the `EchoDebuggerTest` module, to successfully evaluate.

ahelwer:
Impressive! it can now process expressions like:
- `4 > 3` (built-in constants)
- `j > 3` (user-defined constants)
- `(i + 3) > j` (user-defined constants and variables)

So looks like it's time to identify local-scope identifiers and add them to the operator definition.

ahelwer:
It's proving a bit difficult to calculate what local variables can be accessed from any given expression node. Some of the semantic tree nodes have a `SymbolTable` instance associated with them but it isn't a copy, so as the semantic checker process continued the table got modified from under it. So unless some magic is being done in the spec processor that can surface this information I think I'll have to modify the semantic checking process to record some additional info in each semantic node.

lemmy:
Assuming `p1` and `p2` are local identifiers within the scope of the breakpoint expression, would it make things easier if the user to be required to write something like `LET p1 = TRUE, p2 = TRUE IN variable = p1 + p2`?  In other words, let the user identify the local identifiers for you.

ahelwer:
For now perhaps I will do a best-effort crawl back up the path from breakpoint node to module node, recording identifiers introduced by:
1. Formal parameter nodes of operators like `f(a,b,c) == `
2. Identifiers introduced in `LET/IN` blocks
3. Identifiers introduced in quantifiers

This will probably miss some cases (lambdas? idk) but will at least be a start.

ahelwer:
Local identifier resolution is now implemented and seems to basically work, which means we can transition this out of a draft PR!

tlatools/org.lamport.tlatools/src/tla2sany/semantic/SymbolTable.java
```diff
@@ -2,7 +2,9 @@
 // Portions Copyright (c) 2003 Microsoft Corporation.  All rights reserved.
 package tla2sany.semantic;
 
-import tla2sany.utilities.Stack;
+import java.util.ArrayDeque;
```
lemmy, line 5: Why is this change necessary? The collections in `java.util` and those in `tla2sany.util` are unlikely to be fully compatible.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
```diff
@@ -250,7 +250,7 @@ public LinkedList<SemanticNode> get() {
 
 					@Override
 					public void preVisit(final SemanticNode node) {
-						if (location.equals(node.getLocation())) {
+						if (location.equals(node.getLocation()) || (0 == node.getChildren().length && node.getLocation().includes(location))) {
```
lemmy, line 253: `SemanticNode#getChildren` can return null > possible NPE.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
```diff
@@ -0,0 +1,222 @@
...
+		final String bpOpName = generateUnusedName(semanticRoot, "__BreakpointExpr__%s");
+		final List<String> paramNames = getScopedIdentifiers(semanticRoot, location);
+		final String params = paramNames.size() > 0 ? "(" + String.join(", ", paramNames) + ")" : "";
+		final String bpOpDef = bpOpName + params;
+		ToolIO.out.println("BPExpr: wrapping with op \"" + bpOpDef + "\"");
+
+		final String wrapper = "---- MODULE %s ----\nEXTENDS %s\n%s == %s\n====";
+		String wrappedConditionExpr = String.format(wrapper, bpModName, rootModName, bpOpDef, conditionExpr);
+		byte[] wrappedConditionExprBytes = wrappedConditionExpr.getBytes(StandardCharsets.UTF_8);
+		InputStream sourceCode = new ByteArrayInputStream(wrappedConditionExprBytes);
```
lemmy, line 64: Should this stream be closed explicitly? Perhaps, be wrapped in a “try-with-resources” block?

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
```diff
@@ -27,20 +27,26 @@
...
-	public TLCSourceBreakpoint(final String module, final SourceBreakpoint s) {
+	public TLCSourceBreakpoint(final SpecProcessor processor, final String module, final SourceBreakpoint s,
+			final ModuleNode semanticRoot) {
 		setColumn(s.getColumn());
 		setLine(s.getLine());
 		// Create a location that's not a point.
 		final int column = getColumn() != null ? getColumn() : 1;
 		//TODO: If the location spans lines, getLine() + 1 should be the endline (second line parameter of Location).
-		location = new Location(module, getLine() + 1, column, getLine(), column + 1);
+		location = new Location(module, getLine(), column, getLine(), column + 1);
```
lemmy, line 49: This change is likely the cause of the test failures. I believe there was a mismatch between the SANY and the DAP locations.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
```diff
@@ -0,0 +1,222 @@
+package tlc2.debug;
```
lemmy, line 1: Copyright header

tlatools/org.lamport.tlatools/src/tla2sany/semantic/SymbolTable.java
lemmy: If this is absolutely necessary and cannot be avoided, isolate the code into a dedicated commit.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
```diff
@@ -1480,6 +1480,11 @@ protected String getNodeRef() {
       default: throw new IllegalArgumentException("unsupported kind: " + getKind() + " in xml export");
     }
   }
+  
+  public String toString() {
```
lemmy, line 1484: This appears to be a debugging tool that might no longer be relevant.

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java
```diff
@@ -123,7 +123,8 @@ public class SpecProcessor implements ValueConstants, ToolGlobals {
     private ModuleNode rootModule; // The root module.
     private Set<OpDefNode> processedDefs;
     private SpecObj specObj;
-    private Defns snapshot;
+    private Defns preConstantSnapshot;
```
lemmy, line 126: The changes made to the `SpecProcessor` will also need to be isolated into a separate commit. This is my TODO.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
lemmy: Not just hypothetical; just happened when toying with the PR.

lemmy: Note the comment about elems of getChildren potentially being `null` in `tla2sany.semantic.SemanticNode#getListOfChildren`

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
```diff
@@ -0,0 +1,222 @@
...
+			return null;
+		}
+		ToolIO.out.println("BPExpr: success [level analysis]");
+		
+		OpDefNode bpOp = bpModule.getOpDef(bpOpName);
+		if (null == bpOp) {
+			ToolIO.err.println("ERROR: unable to find breakpoint expression op " + bpOpName);
+		}
+
+		if (!(LevelConstants.ConstantLevel == bpOp.level || LevelConstants.VariableLevel == bpOp.level
```
lemmy, line 105: This level check is sufficient for evaluating *Watch* expressions (not breakpoints). A Watch expression is evaluated for the currently active stack frame, regardless of whether that frame matches the expression’s level. If the evaluation fails, the debugger displays an appropriate message indicating that the Watch expression failed to evaluate.

lemmy:
> The TLA+ Debugger also supports "Watch" expressions. AFAICT, Watch and Breakpoint expressions will be able to share most of the implementation.
> 
> https://www.youtube.com/watch?v=r7kGbjcSaKU

Ad-hoc/on-the-fly/dynamic Watch expressions have been implemented in branch `mku-pr1099`, which has cherry-picked the commits from this branch.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
lemmy: https://github.com/tlaplus/tlaplus/commit/81ed3ffbbcb71ef70c3a1f6721c9a113a77f09c1

lemmy:
> Local identifier resolution is now implemented and seems to basically work, which means we can transition this out of a draft PR!

Is the following case (`bciFastPath` let expression) supported?  Hovering (right) resolves `bciFastPath`.

Screenshot 2024-12-19 at 6 48 07 PM

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
ahelwer: If the user puts in an existing definition like `TypeOK` I think the processing code will ultimately result in what the user intends; it could be short-circuited by looking up existing definitions in the module but this sort of special casing would also add another code path to test.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/SymbolTable.java
ahelwer: I will cherry-pick this commit out into another branch for submission in the future. Primarily I made it because the state of our custom-written `Stack` class is hard to understand when viewed in the debugger. It would be nice to get rid of it eventually.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
ahelwer: Per Java docs closing a `ByteArrayInputStream` has no effect: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/ByteArrayInputStream.html#close()

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
ahelwer: I recall I ran into trouble getting a match using the `SemanticNode.pathTo` method with the `+ 1` added; will investigate further.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
ahelwer: Does that mean I shouldn't implement the
>Conditionally restrict operator action level depending on breakpoint location

feature?

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
lemmy: Related, breakpoint validation/verification (hollow breakpoint vs. Red breakpoint) no longer seems to works.  Bisecting should be the fastest to find the offending commit

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
lemmy: Definitely not for Watch expressions, but I could imagine that novice users would appreciate it if the debugger would notify them if a breakpoint expression's level exceeds the scope's level. That said, does DAP even provide a mechanism to notify the user beyond visual breakpoint validation/verification (hollow vs. red breakpoint)?

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
lemmy: How does it ultimately result in what the user intends?  Does it create a module containing a single definition, `TypeOK_139488708`, which delegates to `TypeOK` in the original specification, similar to how it would handle true ad-hoc/on-the-fly/dynamic breakpoint expressions?

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
ahelwer: Like bisect within these changes or in the repo in general?

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
ahelwer: Not that I was able to find, other than hollow breakpoints + debug log.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
ahelwer: It would create something like
```tla
---- MODULE __BreakpointModule__12345678
EXTENDS OriginalModuleName
__BreakpointDef__12345678(i, j, k) == TypeOK
====
```
and then when the breakpoint is evaluated it should just evaluated `TypeOK`.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
ahelwer: It seems like the `+ 1` works with my changes to how `pathTo` works (accept if node is a leaf node and the location is contained within it).

ahelwer:
I added a bunch of tests for the `getScopedIdentifier` method. It supports the `bciFastPath` case, and I also added support for `LAMDA` expressions (no idea whether the debugger itself supports breaking inside a `LAMBDA` but whatever).

ahelwer:
There's a failure in `EWD998ChanDebuggerTest` and I'm not really sure how to debug it:

```
org.junit.ComparisonFailure: expected:<[OpAppl]Node> but was:<[At]Node>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at tlc2.debug.EWD998ChanDebuggerTest.testSpec(EWD998ChanDebuggerTest.java:374)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at util.IsolatedTestCaseRunner.run(IsolatedTestCaseRunner.java:71)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:93)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:40)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:529)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:757)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:452)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:210)

java.lang.AssertionError: expected:<0> but was:<-1>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at tlc2.tool.liveness.ModelCheckerTestCase.assertExitStatus(ModelCheckerTestCase.java:267)
	at tlc2.tool.liveness.ModelCheckerTestCase.tearDown(ModelCheckerTestCase.java:263)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.RunAfters.invokeMethod(RunAfters.java:46)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:33)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at util.IsolatedTestCaseRunner.run(IsolatedTestCaseRunner.java:71)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:93)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:40)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:529)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:757)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:452)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:210)
```

lemmy:
> There's a failure in `EWD998ChanDebuggerTest` and I'm not really sure how to debug it:
> 
> ```
> org.junit.ComparisonFailure: expected:<[OpAppl]Node> but was:<[At]Node>
> 	at org.junit.Assert.assertEquals(Assert.java:117)
> 	at org.junit.Assert.assertEquals(Assert.java:146)
> 	at tlc2.debug.EWD998ChanDebuggerTest.testSpec(EWD998ChanDebuggerTest.java:374)
> 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
> 	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
> 	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
> 	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
> 	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
> 	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
> 	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
> 	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
> 	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
> 	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
> 	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
> 	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
> 	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
> 	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
> 	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
> 	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
> 	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
> 	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
> 	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
> 	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
> 	at util.IsolatedTestCaseRunner.run(IsolatedTestCaseRunner.java:71)
> 	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:93)
> 	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:40)
> 	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:529)
> 	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:757)
> 	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:452)
> 	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:210)
> 
> java.lang.AssertionError: expected:<0> but was:<-1>
> 	at org.junit.Assert.fail(Assert.java:89)
> 	at org.junit.Assert.failNotEquals(Assert.java:835)
> 	at org.junit.Assert.assertEquals(Assert.java:647)
> 	at org.junit.Assert.assertEquals(Assert.java:633)
> 	at tlc2.tool.liveness.ModelCheckerTestCase.assertExitStatus(ModelCheckerTestCase.java:267)
> 	at tlc2.tool.liveness.ModelCheckerTestCase.tearDown(ModelCheckerTestCase.java:263)
> 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
> 	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
> 	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
> 	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
> 	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
> 	at org.junit.internal.runners.statements.RunAfters.invokeMethod(RunAfters.java:46)
> 	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:33)
> 	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
> 	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
> 	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
> 	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
> 	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
> 	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
> 	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
> 	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
> 	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
> 	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
> 	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
> 	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
> 	at util.IsolatedTestCaseRunner.run(IsolatedTestCaseRunner.java:71)
> 	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:93)
> 	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:40)
> 	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:529)
> 	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:757)
> 	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:452)
> 	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:210)
> ```

It is due to the following change:

https://github.com/tlaplus/tlaplus/pull/1099/files#diff-7ecdbd526fe811798f77baebd73fd5cf826805f74199fbd832a3037cb505bda7R256-R257

ahelwer:
Okay I fixed it, UTs should pass now. The only remaining thing is to fix up my `ExpressionBreakpointTest` to actually assert some things.

lemmy:
I force-pushed [two commits](https://github.com/tlaplus/tlaplus/compare/master...mku-pr1099) to branch [mku-pr1099](https://github.com/tlaplus/tlaplus/compare/master...mku-pr1099) that (logically) replace the `SpecProcessor`-related changes in commit https://github.com/tlaplus/tlaplus/pull/1099/commits/3287902cfbf52774d2d194d25542cd69f9c532d5

lemmy:
Below is the prototype to enable support for dynamic Watch expressions:

```diff
commit 3de9dc64aafcc4f70439046dbd7ac82aa90137de
Author: Markus Alexander Kuppe <github.com@lemmster.de>
Date:   Thu Dec 19 18:36:40 2024 -0800

    Initial support for ad-hoc/on-the-fly/dynamic Watch expressions.

    [Feature][TLC][Debuggeer]

    Signed-off-by: Markus Alexander Kuppe <github.com@lemmster.de>

diff --git a/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java b/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
index 10a99000e..4e64b9499 100644
--- a/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
@@ -38,12 +38,19 @@ public class TLCBreakpointExpression {
 	 * @param conditionExpr The unparsed breakpoint expression.
 	 * @return A breakpoint expression, or null if parsing failed.
 	 */
-	public static OpDefNode process(
-			final SpecProcessor processor,
-			final ModuleNode semanticRoot,
-			final Location location,
-			final String conditionExpr
-	) {
+	public static OpDefNode process(final SpecProcessor processor, final ModuleNode semanticRoot,
+			final Location location, final String conditionExpr) {
+		final List<String> paramNames = getScopedIdentifiers(semanticRoot, location);
+		return process(processor, semanticRoot, paramNames, conditionExpr);
+	}
+
+	public static OpDefNode process(final SpecProcessor processor, final ModuleNode semanticRoot,
+			final String conditionExpr) {
+		return process(processor, semanticRoot, new ArrayList<>(), conditionExpr);
+	}
+
+	public static OpDefNode process(final SpecProcessor processor, final ModuleNode semanticRoot,
+			final List<String> paramNames, final String conditionExpr) {
 		if (null == conditionExpr || conditionExpr.isBlank()) {
 			return null;
 		}
@@ -53,7 +60,6 @@ public class TLCBreakpointExpression {
 		final String bpModName = generateUnusedName(semanticRoot, "__BreakpointModule__%s");
 		ToolIO.out.println("BPExpr: wrapping with module \"" + bpModName + "\"");
 		final String bpOpName = generateUnusedName(semanticRoot, "__BreakpointExpr__%s");
-		final List<String> paramNames = getScopedIdentifiers(semanticRoot, location);
 		final String params = paramNames.size() > 0 ? "(" + String.join(", ", paramNames) + ")" : "";
 		final String bpOpDef = bpOpName + params;
 		ToolIO.out.println("BPExpr: wrapping with op \"" + bpOpDef + "\"");
diff --git a/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCStackFrame.java b/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCStackFrame.java
index d48ea4a04..09e71cad9 100644
--- a/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCStackFrame.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCStackFrame.java
@@ -545,8 +545,17 @@ public class TLCStackFrame extends StackFrame {
 		if (name == null) {
 			return new EvaluateResponse();
 		}
+
+		// 1) Try to resolve the name to an existing expression.
 		final ModuleNode module = tool.getSpecProcessor().getRootModule();
-		return getWatch(module.getOpDef(name));
+
+		// 2) Expression doesn't exist. Check if name is an ad-hoc expression.
+		OpDefNode odn = module.getOpDef(name);
+		if (odn == null) {
+			odn = TLCBreakpointExpression.process(tool.getSpecProcessor(), module, name);
+		}
+
+		return getWatch(odn);
 	}

 	public EvaluateResponse getWatch(final OpDefNode odn) {
```

ahelwer:
Do we want to merge this PR and then work on watch expressions on top of it as an additional feature?

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
```diff
@@ -0,0 +1,269 @@
...
+			final String conditionExpr
+	) {
+		if (null == conditionExpr || conditionExpr.isBlank()) {
+			return null;
+		}
+		ToolIO.out.println("BPExpr: processing expression \"" + conditionExpr + "\"");
+
+		final String rootModName = semanticRoot.getName().toString();
+		final String bpModName = generateUnusedName(semanticRoot, "__BreakpointModule__%s");
+		ToolIO.out.println("BPExpr: wrapping with module \"" + bpModName + "\"");
```
lemmy, line 83: The logging here and below seems a bit excessive.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
```diff
@@ -0,0 +1,269 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Linux Foundation. All rights reserved.
```
lemmy, line 2: Class was created in 2024.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
```diff
@@ -0,0 +1,269 @@
...
+		SyntaxTreeNode syntaxRoot = parser.ParseTree;
+		if (!syntaxParseSuccess || null == syntaxRoot) {
+			// Parse error is output to ToolIO.out
+			return null;
+		}
+		ToolIO.out.println("BPExpr: success [syntactic analysis]");
+
+		Errors semanticLog = new Errors();
+		SemanticNode.setError(semanticLog); // Annoyingly static
+		Generator semanticChecker = new Generator(semanticRoot.semanticChecker.moduleTable, semanticLog);
```
lemmy, line 105: There is no need to change `ModuleNode` (see above) because this can use `SpecProcessor#getModuleTbl`.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
```diff
@@ -0,0 +1,269 @@
...
+	 * OpApplNode.allParams field, but only if the given expression actually
+	 * references an outside parameter. By the time the semantic analysis
+	 * process has actually reached that stage it's too late to append the
+	 * references to the operator definition node.
+	 *
+	 * @param semanticRoot The root node of the semantic parse tree.
+	 * @param location     The breakpoint location.
+	 * @return A list of local identifiers accessible at the given location.
+	 */
+	static Set<String> getScopedIdentifiers(
```
lemmy, line 225: Should/Could this be a method of `ModuleNode`?

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
```diff
@@ -0,0 +1,269 @@
...
+	}
+	
+	/**
+	 * Generates a plausible definition name that is not already in use.
+	 *
+	 * @param semanticRoot The module in which to check for name collisions.
+	 * @param pattern A base pattern in which a number can be interpolated.
+	 * @return A name that is unique within the context of the module.
+	 */
+	private static String generateUnusedName(ModuleNode semanticRoot, String pattern) {
```
lemmy, line 154: Again, method of `ModuleNode`?

tlatools/org.lamport.tlatools/test/util/ParserInterface.java
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Fully parses and checks the given source code input.
+   *
+   * @param The TLA+ source to parse, as a string.
+   * @return A level-checked semantic parse tree, or null if failure.
+   */
+  public static ModuleNode parse(String input) {
+    SyntaxTreeNode syntaxRoot = processSyntax(input);
+    if (null == syntaxRoot) return null;
```
lemmy, line 31: Please add curly braces.

tlatools/org.lamport.tlatools/test/util/ParserInterface.java
```diff
@@ -0,0 +1,108 @@
...
+import tla2sany.semantic.Generator;
+import tla2sany.semantic.ModuleNode;
+import tla2sany.semantic.SemanticNode;
+
+/**
+ * A simple interface for parsing self-contained TLA+ modules, for use in
+ * tests. This class can be viewed as a prototype of a future API to be
+ * integrated into SANY itself.
+ */
+public class ParserInterface {
```
lemmy, line 21: Terminology: "Interface" yet it is a Java class.

tlatools/org.lamport.tlatools/test/tlc2/debug/GetScopedIdentifiersTests.java
```diff
@@ -0,0 +1,95 @@
...
+    // expressions at arbitrary points in the semantic tree, rendering all
+    // of this functionality redundant.
+    new TestCase("op(_+_) ≜", "", "+"),
+  };
+
+  /**
+   * Run all the test cases and check the identifiers are found as expected.
+   */
+  @Test
+  public void test() {
```
lemmy, line 85: Let's stick to ordinary junit tests and one @Test method per expression so that we know which one fails (if any).

tlatools/org.lamport.tlatools/test/util/ParserInterface.java
```diff
@@ -0,0 +1,108 @@
+package util;
```
lemmy, line 1: This class seems largely redundant to TLCBreakpointExpression?  Can this class be moved out of the test package and be used by TLCBreakpointExpression?

tlatools/org.lamport.tlatools/test/tlc2/debug/GetScopedIdentifiersTests.java
ahelwer: This is a deficiency in JUnit 4 (which we use) that was fixed in JUnit 5 with the addition of [dynamic tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-dynamic-tests), so each of the test cases here could be generated as a separate test at runtime. Testing has broadly moved away from hardcoded tests toward tests-as-data or even auto-generated test cases. I find it much easier to add another line in the array compared to adding another test method.

tlatools/org.lamport.tlatools/test/util/ParserInterface.java
ahelwer: I was thinking of calling it API. Do you want to use ParserAPI or ParserApi casing?

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
ahelwer: I think it's a bit half-baked to move into ModuleNode. I would rather fix this by refactoring each node to keep a snapshot of the `tla2sany.semantic.SymbolTable` stack around. Not something that can be done without the semantic test corpus.

tlatools/org.lamport.tlatools/test/util/ParserInterface.java
ahelwer: I agree it is redundant; my sort of long-term goal is to develop this API class according to the needs of the test code, and then eventually introduce it in a `tla2sany.api` java package once it has been sufficiently refined. Ideally it would then not change very much. However a few things remain before I think it's fully baked:
1. Remove `SemanticNode.error` and `Context.initialContext` global static state
2. Figure out a good module resolution process
3. Figure out a good incremental parse API, allowing users to add modules & definitions and hopefully ghost expressions to be evaluated at an arbitrary point in the tree.
4. Figure out how the API would integrate with a systematic error reporting system as sketched in https://github.com/tlaplus/rfcs/issues/15

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
ahelwer: Seems like a good candidate yeah

tlatools/org.lamport.tlatools/test/tlc2/debug/GetScopedIdentifiersTests.java
Calvin-L: JUnit 4 supports [parameterized tests](https://github.com/junit-team/junit4/wiki/Parameterized-tests), which is the best of both worlds:
 - You can have a simple array of test inputs
 - Each input is treated as a separate test in JUnit's reports

lemmy: +1 for parameterized tests, which have already been used in e.g. https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/test/tlc2/value/impl/SubsetEnumeratorTest.java

tlatools/org.lamport.tlatools/test/tlc2/debug/ExpressionBreakpointTest.java
```diff
@@ -0,0 +1,66 @@
...
+	public ExpressionBreakpointTest() {
+		super(RM, FOLDER, new String[] { "-config", RM + ".tla" }, EC.ExitStatus.SUCCESS);
+	}
+	
+	private static int getContextVal(TLCStackFrame frame, String var) {
+		Object result = frame.getContext().lookup(s -> s.getName().equals(var));
+		if (result instanceof IntValue) {
+			return ((IntValue)result).val;
+		} else if (result instanceof LazyValue) {
+			// Evade cache guard by guessing value from small range.
```
lemmy, line 27: Why is this necessary?

ahelwer: I don't know how to access the value of lazy things. It seems like it is guarded fairly well.

lemmy:
```java
return ((IntValue) lazy.getValue(frame.tool, frame.getS(), frame.getT(), EvalControl.Clear)).val;
```

tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
```diff
@@ -228,14 +231,20 @@ public T get() {
...
+	 *          node spans the exact given location; if false allow paths
+	 *          starting from a leaf node that completely encompasses the
+	 *          given location.
+	 * @return The path in the semantic tree starting at the node matching
+	 *         the location and moving up through parents until reaching this
+	 *         node. An empty list is returned if no node in the subtree
+	 *         matches the given location constraints.
 	 */
-	public LinkedList<SemanticNode> pathTo(final Location location) {
+	public LinkedList<SemanticNode> pathTo(final Location location, boolean requireExactLocationMatch) {
```
lemmy, line 247: The 1-arity `pathTo` could delegate to the new, 2-arity pathTo.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCStackFrame.java
```diff
@@ -499,7 +499,7 @@ public EvaluateResponse get(final EvaluateArguments ea) {
 			final String moduleName = Paths.get(u.getPath()).getFileName().toString().replaceAll(".tla$", "");
 
 			final Location location = Location.parseCoordinates(moduleName, u.getFragment());
-			final LinkedList<SemanticNode> path = tool.getModule(location.source()).pathTo(location);
+			final LinkedList<SemanticNode> path = tool.getModule(location.source()).pathTo(location, true);
```
lemmy, line 502: https://github.com/tlaplus/tlaplus/pull/1099/files#r1899790891

tlatools/org.lamport.tlatools/test/tla2sany/semantic/TestLevelChecking.java
```diff
@@ -1,14 +1,10 @@
 package tla2sany.semantic;
```
lemmy, line 1: Copyright header

tlatools/org.lamport.tlatools/test/tlc2/debug/ExpressionBreakpointTest.java
```diff
@@ -0,0 +1,66 @@
+package tlc2.debug;
```
lemmy, line 1: Copyright header

tlatools/org.lamport.tlatools/test/tlc2/debug/GetScopedIdentifiersTests.java
```diff
@@ -0,0 +1,104 @@
+package tlc2.debug;
```
lemmy, line 1: Copyright header

tlatools/org.lamport.tlatools/test/util/ParserAPI.java
```diff
@@ -0,0 +1,110 @@
+package util;
```
lemmy, line 1: Copyright header

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
```diff
@@ -27,14 +27,20 @@
...
+import tla2sany.semantic.OpDefNode;
 import tla2sany.st.Location;
+import tlc2.tool.impl.SpecProcessor;
 
 public class TLCSourceBreakpoint extends SourceBreakpoint {
 
 	private final int hits;
 	private final Location location;
+	// Use Getter
+	public final OpDefNode condition;
```
lemmy, line 40: `condition` can be private if the implementation of the method `TLCStateStackFrame#matchesExpression` is moved into `TLCSourceBreakpoint` (`TLCStateStackFrame.matchesExpression` then delegates to `TLCSourceBreakpoint#matchesExpression).  Moreover, the impl no longer has to check `bp.getCondition() != null && !bp.getCondition().isEmpty()`.

tlatools/org.lamport.tlatools/test/tlc2/debug/ExpressionBreakpointTest.java
lemmy: Even shorter:

```java
return ((IntValue) frame.unlazy(lazy)).val;
```

tlatools/org.lamport.tlatools/test/tlc2/debug/ExpressionBreakpointTest.java
```diff
@@ -0,0 +1,66 @@
...
+	public void testSpec() throws Exception {
+		// Set breakpoint
+		SourceBreakpoint bp = createBreakpointInfo(8, 5, 1, "(i + l + k) > j");
+		SetBreakpointsArguments sba = createBreakpointArgument(RM, bp);
+		debugger.setBreakpoints(sba);
+
+		// Break at init
+		debugger.continue_();
+		
+		// Break at bp
```
lemmy, line 54: This block can be simplified into:

```java
// Break at bp
final TLCStateStackFrame current = (TLCStateStackFrame) debugger.continue_()[0];
assertEquals(2, ((IntValue) current.getT().getVals().get(UniqueString.of("i"))).val);
final Map<String, String> allVariables = new HashMap<>();
allVariables.put("k", "5");
allVariables.put("l", "4");
assertTLCStateFrame(current, 8, 8, RM, allVariables);
```

`getStateVal` and `getContextVal` become obsolete.

tlatools/org.lamport.tlatools/test/tlc2/debug/GetScopedIdentifiersTests.java
ahelwer: Do we really need the MIT header in every file?

lemmy: IANAL, but this is standard practice, and most IDEs typically fold the license header automatically.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
```diff
@@ -0,0 +1,245 @@
...
+		final Set<String> paramNames = getScopedIdentifiers(semanticRoot, location);
+		final String params = paramNames.size() > 0 ? "(" + String.join(", ", paramNames) + ")" : "";
+		final String bpOpDef = bpOpName + params;
+
+		final String wrapper = "---- MODULE %s ----\nEXTENDS %s\n%s == %s\n====";
+		String wrappedConditionExpr = String.format(wrapper, bpModName, rootModName, bpOpDef, conditionExpr);
+		byte[] wrappedConditionExprBytes = wrappedConditionExpr.getBytes(StandardCharsets.UTF_8);
+		InputStream sourceCode = new ByteArrayInputStream(wrappedConditionExprBytes);
+		TLAplusParser parser = new TLAplusParser(sourceCode, StandardCharsets.UTF_8.name());
+		boolean syntaxParseSuccess = parser.parse();
```
lemmy, line 91: A `conditionExpr` such as "Does not exist" cause `parse()` to print:

```
***Parse Error***
Was expecting "==== or more Module body"
Encountered "not" at line 3, column 39 and token "Does" 
```

lemmy:
> Do we want to merge this PR and then work on watch expressions on top of it as an additional feature?

The size of the related commit (https://github.com/tlaplus/tlaplus/commit/8baf0d86ac3d14f26c11c3b3261a6b8f2e70eb54) doesn't warrant a dedicated PR.  Please just cherry-pick on top of your branch.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
lemmy: See `EWD998ChanDebuggerTest` after cherry-picking https://github.com/tlaplus/tlaplus/commit/8baf0d86ac3d14f26c11c3b3261a6b8f2e70eb54

lemmy: To alleviate the issue, the error message could be modified to include the module name, thereby prompting users to recognize that the problem lies within the debug module.

tlatools/org.lamport.tlatools/test/util/ParserInterface.java
lemmy: Let’s consider this out of scope for this PR.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
lemmy: Out of scope for this PR.

lemmy (approved):
LGTM once my unresolved comments have been addressed.

tlatools/org.lamport.tlatools/test/tlc2/debug/ExpressionBreakpointTest.java
ahelwer: I kind of like the test as it is. It's a bit more clear that a property is being tested instead of hardcoding specific expected values.

lemmy: For consistency, please stick to the style that is used by the other `tlc2.debug` tests.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCBreakpointExpression.java
ahelwer: Added some additional error logging

tlatools/org.lamport.tlatools/test/tlc2/debug/ExpressionBreakpointTest.java
ahelwer: I will change it, but I would like to lodge feedback that I find these type of tests confusing since it is never clear if they're just recording & playing back behavior or actually testing a property that matters. A good example is the test failure that was caused by changing the `pathTo` function, which manifested in a very weird way.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCSourceBreakpoint.java
ahelwer: I looked into doing this but there is a fair bit of `TLCStateStackFrame`-specific stuff in the `matchesExpression` function:
- The context
- The tool
- `getS()`
- `getT()`

tlatools/org.lamport.tlatools/test/util/ParserAPI.java
```diff
@@ -0,0 +1,133 @@
...
+    SemanticNode.setError(log);
+    // The null parameter here is the {@link ExternalModuleTable}, which will
+    // need to be resolved & populated if this API is to one day support TLA+
+    // modules that EXTEND or INSTANCE other TLA+ modules.
+    Generator gen = new Generator(null, log);
+    try {
+      ModuleNode semanticRoot = gen.generate(parseTree);
+      return log.isSuccess() ? semanticRoot : null;
+    } catch (AbortException e) {
+      Assert.fail(e.toString() + log.toString());
```
lemmy, line 116: Why does this catch the specific `AbortException` to replace it with the generic `TLCRuntimeException`?

tlatools/org.lamport.tlatools/test/util/ParserAPI.java
```diff
@@ -0,0 +1,133 @@
...
+      return null;
```
lemmy, line 117: Dead code.

ahelwer: Yes but the Java compiler is not smart enough to know that and raises a compilation error that the method needs to return a value in all paths.

tlatools/org.lamport.tlatools/test/util/ParserAPI.java
ahelwer: Temporary. The error handling of this API is not yet figured out. Probably it will eventually be much more elaborate as https://github.com/tlaplus/rfcs/issues/15 is worked on.

lemmy: Please use TODO markers for incomplete work.

tlatools/org.lamport.tlatools/test/tlc2/debug/GetScopedIdentifiersTests.java
```diff
@@ -0,0 +1,134 @@
...
+   * terminated by expression TRUE located at the start of the next line,
+   * along with a variable-length list of identifiers accessible at the
+   * point of the TRUE expression. A terminator to be appended after TRUE is
+   * also accepted - for example ⟩ or ) or ] or }.
+   */
+  @Parameters(name = "{index}: {0}")
+  public static TestCase[] testCases() {
+    return new TestCase[] {
+      new TestCase("op ≜", ""),
+      new TestCase("op(i, j, k) ≜", "", "j", "i", "k"),
```
lemmy, line 95: This doesn't have to be changed, but in general, I want to use ASCII in our test suite to maintain compatibility with older versions of tla2tools that were released before the Unicode feature was introduced. This ensures we can run new tests on those older versions should this become necessary.

lemmy:
Thanks, logically merged into `master`

lemmy:
`tlc2.debug.TLCDebuggerExpression.process(SpecProcessor, ModuleNode, String)` doesn't handle expressions of the form `LET M == INSTANCE SomeModule IN M!SomeOp` if `SomeModule` appears nowhere else.  The fact that the `ExternalModuleTable` contains no entry for `SomeModule` makes  `tla2sany.semantic.Generator.processModuleDefinition(TreeNode, Vector, Vector, ModuleNode)` add a parse error.

ahelwer:
I can see why that would be desirable to write. I will think about how to handle it.

ahelwer:
@lemmy the `LET/IN` expression approach has been prototyped with tests in PR #1228.

PR #1101, opened 2024-12-17
SANY: Remove most static error logging usage

ahelwer:
Reduce usage of static SemanticNode.errors field
Some additional places of use remain but this is a good incremental effort. Reduced usage count from 32 to 7.

This was all fairly mechanical (adding `Errors` parameters to methods) except for two places:
1. When constructing the SANY `Explorer` instance which is used to print out the SANY parse tree from the command line, provide `spec.getSemanticErrors()` which is the same `Errors` logging instance as the `SemanticNode.errors` field whose usage is being replaced
2.  In `tlatools/org.lamport.tlatools/src/tlc2/tool/Specs.java`, which is the only non-SANY part of the codebase calling into a `tla2sany.semantic` class constructor requiring an `Errors` instance (`SubstInNode`), I provide a new blank `Errors` instance; this means if an error is logged (which would only seem to happen if one of the other constructor parameters is null) it will be ignored, but that error is already being ignored because nothing checks the `SemanticNode.errors` field that far into the model-checking process.

[Refactoring][SANY]
Ref #891

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/explorer/Explorer.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/explorer/Explorer.java
@@ -16,6 +16,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.StringTokenizer;
 
+import tla2sany.semantic.Errors;
 import tla2sany.semantic.ExternalModuleTable;
 import tla2sany.semantic.FormalParamNode;
 import tla2sany.semantic.Generator;
@@ -39,8 +40,6 @@ import util.UniqueString;
 
 public class Explorer {
 
-	public Generator generator;
-
 	// Next three vars used in reading commands from keyboard
 	private final InputStreamReader inStream = new InputStreamReader(System.in);
 	private final int inCapacity = 100;
@@ -59,11 +58,13 @@ public class Explorer {
 	private ExploreNode obj;
 
 	private ExternalModuleTable mt;
+	private final Errors errors;
 
 	// Constructor
-	public Explorer(ExternalModuleTable mtarg) {
+	public Explorer(ExternalModuleTable mtarg, Errors errors) {
 
 		mt = mtarg;
+		this.errors = errors;
 
 	}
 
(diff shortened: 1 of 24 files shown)


lemmy:
> 2\. @lemmy since you're more familiar with the TLC side of the codebase do you know if that would ever happen or is it guaranteed to be non-null?

I don't know and, unfortunately, I won't have time to investigate anytime soon.

ahelwer:
All right, well I think a safe thing to do then is just pass in a new blank Errors instance for the error to be logged into. It will be ignored if it happens but it's already ignored (nothing looks at the semantic log that far into the model checking process) so whatever.

tlatools/org.lamport.tlatools/src/tla2sany/explorer/Explorer.java
```diff
@@ -59,11 +58,13 @@ public class Explorer {
 	private ExploreNode obj;
 
 	private ExternalModuleTable mt;
+	private Errors errors;
```
Calvin-L, line 61: Nit: this field can be `final`.

(So can many other fields in this class, but let's keep things focused on `Errors`.)

tlatools/org.lamport.tlatools/src/tla2sany/semantic/ModuleNode.java
```diff
@@ -1025,8 +1025,9 @@ public final int getLevel() {
    * 3. It extends and instantiates only constant modules.
    *
    * NOTE: Can only be called after calling levelCheck
+ * @param errors TODO
```
Calvin-L, line 1028: Please finish writing this comment.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/SubstInNode.java
```diff
@@ -279,8 +279,9 @@ final void addExplicitSubstitute(Context instanceCtxt, UniqueString lhs,
    * implicitly subject to the substitution X <- X.  If that is not
    * possible, because X is not defined in the instantiating module,
    * then we have an error.
+ * @param errors TODO
```
Calvin-L, line 282: Please finish writing this comment.

Calvin-L (approved):
Thanks for following up on those small nits. This change looks good to me. 👍

PR #1088, opened 2024-11-27
Add new `-dumptrace` format "tlcaction"

lemmy:
Add `-dumpTrace tlcaction file.tla` to dump the sequence of actions and their arguments/parameters leading to the counterexample into a file called `file.tla`.

#### The dumped trace
```tla
---- MODULE TLCActionTrace ----
EXTENDS Simple

_length == 9

_next(_idx) ==
\/ 1 = _idx /\ A(23)
\/ 2 = _idx /\ B(77, 42)
...
\/ 9 = _idx /\ A(87)

_actionConstraint ==
_next(LET T == INSTANCE TLC IN T!TLCGet("level"))
====
```
#### User provided
```tla
------- MODULE MinimizeSimple -------
EXTENDS Simple, TLCActionTrace

VARIABLE _idx

MinimizeInit ==
    /\ Init \* Simple!Init
    /\ _idx = {}

MinimizeNext ==
    \E i \in 1.._length \ _idx:
            /\ _idx' = _idx \cup {i}
            /\ _next(i)

MinimizeView ==
    x \* Simple!x

=====
---- CONFIG MinimizeSimple -----
INIT MinimizeInit
NEXT MinimizeNext
VIEW MinimizeView

INVARIANT Inv
=====
```
#### Original Spec
```tla
------- MODULE Simple -------
EXTENDS Naturals

VARIABLES x

Init == x = 0

A(i) == x' = x + i

B(i, n) == x' = x - i

Next == \E i \in 0..100: A(i) \/ B(i, 42)

Inv == x < 2000

====
```

Related to https://github.com/tlaplus/tlaplus/issues/400 and https://github.com/tlaplus/tlaplus/issues/1084

PR #1102, opened 2024-12-18
Make deserializing previously unseen strings work

fhackett:
Fixes the problem demonstrated in https://github.com/tlaplus/CommunityModules/pull/107.

Currently, if TLC tries to load a string value from disk that its current instance has not already "seen", it will crash.
This is because loading strings from disk does not update the string intern table. The current code only reads from it, and that leads to the impossible situation of TLC needing to know a string already before it can load it.

This behavior was not originally a problem, because the binary serialization mechanism was written for the state queue and checkpointing. Those correctly load a separately-serialized string table that has all the right entries, making loading an "unknown" string impossible.

When this feature was extended to support loading arbitrary values in the community modules, however, "unknown" strings become not only possible, but quite likely in that scenario.

My proposed fix is to ignore the `Map` passed in as a read-only string table proxy, and directly add/load the string to the intern table. Since this code path was written for the community modules and does not interfere with regular state queue / checkpointing (I think?), I can't think of any situation where the `Map` could be anything other than a read-only proxy to the existing intern table anyhow. At that point, my change is to just make sure new strings get added.

A more aggressive change would be to delete the parameter and rename the "when not loading a checkpoint" methods to something like `readExtern`. However, the community modules and TLC are released separately, and the community modules call the exact method I'm talking about. I wouldn't want to cause Java method errors for people, so the "smoothest" fix seems to be to just ignore the `Map` param.

@lemmy as original author of the code I'm changing, let me know if I misunderstood something, or if there's something else I should take into account.


tlatools/org.lamport.tlatools/src/util/UniqueString.java
```diff
@@ -339,7 +339,7 @@ public static UniqueString read(IDataInputStream dis, final Map<String, UniqueSt
         dis.readInt(); // skip, because invalid for the given internTbl
         final int slen = dis.readInt();
         final String str = dis.readString(slen);
-        return tbl.get(str);
```
lemmy, line 342: It's probably a good idea to leave a note as to why the Map gets ignored.

fhackett: Good idea. I actually found a way to (1) minimize tbl usage, and (2) also comment that minimal usage. See latest commit.

tlatools/org.lamport.tlatools/test/tlc2/value/ValueInputOutputStreamTest.java
```diff
@@ -115,4 +122,83 @@ public void testWriteNat() throws IOException {
 		assertEquals(0, in.readNat());
 		in.close();
 	}
+
```
lemmy, line 125: I propose adding a TLA+-level test where the specification contains several definitions that make up the set of strings `S`. The spec's correctness property should involve string equality.  The input `readExternal` should deserialize strings that are present in `S` and those that are not.  

Background: Two `UniqueString`s instance are equal regardless of their `toString` representation. Instead, the position of the String in TLC's InternTable determines if two `UniqueString` instances are equal. The position in the `InternTable` is assigned non-deterministically, making it possible for two `UniqueString` instances to be equal (unequal) even if their string reps are equal (unequal).

tlatools/org.lamport.tlatools/src/tlc2/value/ValueInputStream.java
```diff
@@ -91,7 +91,27 @@ public final IValue read() throws IOException {
 		}
 	}
 	
+	/**
```
lemmy, line 94: Perhaps, this (and https://github.com/tlaplus/tlaplus/pull/1088#discussion_r1876749598) suggests that the `IOUtils` module should be migrated to TLC,  like the `Json` module was eventually [migrated](https://github.com/tlaplus/CommunityModules/pull/84).

fhackett: I see, interesting. I agree. This feature is strongly coupled, and having it in 2 codebases doesn't make sense.

I'm not sure what this migration process would be, however.

It seems maybe:
1. bundle `IOUtils.tla` at some point in time (now, later?)
2. at some future point, make that feature built-in?

tlatools/org.lamport.tlatools/test/tlc2/value/ValueInputOutputStreamTest.java
fhackett: Yes, that seems worth it. Will add.

fhackett: (well, as much as I can without perhaps having access to IOUtils? I think the community modules are not available during test right now... but that doesn't mean I can't do something less direct)

tlatools/org.lamport.tlatools/src/tlc2/value/ValueInputStream.java
lemmy: The Json module and its Java module override was just copied over to TLC (modifications have been applied to both repositories).

fhackett: Oh ok, and then the assumption is that people will download a new TLC and a new CommunityModules at the same time... (right?)

I could do that in this PR then. That would make testing nicer, certainly (see other conversation).

tlatools/org.lamport.tlatools/test/tlc2/value/ValueInputOutputStreamTest.java
lemmy: https://github.com/tlaplus/tlaplus/pull/1103

tlatools/org.lamport.tlatools/src/tlc2/value/ValueInputStream.java
lemmy: [Yes, the other alternative is to make the CM a test-time dependency of TLC. https://github.com/tlaplus/CommunityModules/commit/e43de9c902a1d84744509592d9380c4bd9335751 makes sure that `IOUtils` can be used from ](https://github.com/tlaplus/tlaplus/pull/1103)

lemmy:
@fhackett Please check if https://github.com/tlaplus/tlaplus/pull/1103 addresses your `IOUtils` dependency problem.

fhackett:
@lemmy sorry for the slow response - got sidetracked for about a day.

I checked, and I'm not sure if I did it right.

I checked out your PR branch, and there I added a test like this (100% reverse engineered from other tests, maybe the asserts are wrong, but it doesn't even reach those anyhow):

```java
package tlc2.value;

import static org.junit.Assert.assertFalse;
import static org.junit.Assert.assertTrue;

import org.junit.Test;

import tlc2.output.EC;
import tlc2.tool.liveness.ModelCheckerTestCase;

public class StringDeserializeTLCTest extends ModelCheckerTestCase {
	public StringDeserializeTLCTest() {
		super("StringDeserialize");
	}
	
	@Test
	public void test() {
		assertTrue(recorder.recorded(EC.TLC_FINISHED));
		assertFalse(recorder.recorded(EC.GENERAL));
	}
}
```

I also put the corresponding TLA+ in `test-model/`:
```tla
--------------------------- MODULE StringDeserialize ---------------------------
EXTENDS TLC, IOUtils

ASSUME PrintT("hello world!")

```

I then run the test from Eclipse, and the outcome is:
```
TLC2 Version 2.20 of Day Month 20??
Running breadth-first search Model-Checking with fp 0 and seed 1 with 1 worker on 32 cores with 16040MB heap and 64MB offheap memory [pid: 2993420] (Linux 6.9.3-76060903-generic amd64, Eclipse Adoptium 23.0.1 x86_64, MSBDiskFPSet, DiskStateQueue).
Parsing file /home/finn/programming/tlaplus/tlatools/org.lamport.tlatools/test-model/StringDeserialize.tla
Parsing file /home/finn/programming/tlaplus/tlatools/org.lamport.tlatools/class/tla2sany/StandardModules/TLC.tla

Fatal errors while parsing TLA+ spec in file StringDeserialize

tla2sany.semantic.AbortException
*** Abort messages: 1

Unknown location

Cannot find source file for module IOUtils imported in module StringDeserialize.


Starting... (2024-12-19 22:52:40)
Error: Parsing or semantic analysis failed.
Finished in 00s at (2024-12-19 22:52:40)
```

... so no it seems? But I have no idea why, or how half of the things your PR touches work. At least, it seems like what I posted here would be a good test case for your PR, if I understood things properly.

fhackett:
@lemmy I rebased on your changes, added a TLA+ test (how you suggested, I think?), and squashed.

I think this is mergeable/reviewable as-is so I'm un-drafting it, but there is still one question. It depends if you think it's best to move IOUtils into TLC in the same PR, or leave that as its own thing (I'll make a note to come back and do it - I'm a little curious how the TLC builtins work).

lemmy:
> @lemmy I rebased on your changes, added a TLA+ test (how you suggested, I think?), and squashed.
> 
> I think this is mergeable/reviewable as-is so I'm un-drafting it, but there is still one question. It depends if you think it's best to move IOUtils into TLC in the same PR, or leave that as its own thing (I'll make a note to come back and do it - I'm a little curious how the TLC builtins work).

https://github.com/tlaplus/tlaplus/pull/1103 eliminates the need to move `IOUtils` into TLC.

Out of Apalache, TLAPS, and TLC, only TLC has a meaningful interpretation of `IOUtils`. However, I can see community tools like tla-web, pluspy, and others support `IOUtils`. For this reason, I lean towards retaining `IOUtils` within the CommunityModules.

tlatools/org.lamport.tlatools/test/tlc2/value/StringDeserializeTLCTest.java
```diff
@@ -0,0 +1,21 @@
+package tlc2.value;
```
lemmy, line 1: Please add a copyright header.

lemmy (approved):
LGTM

lemmy:
@Calvin-L Your feedback; it's always appreciated. I plan to merge this PR next week, but please let us know if you'd like to review it first.

tlatools/org.lamport.tlatools/test/tlc2/value/StringDeserializeTLCTest.java
fhackett: I attempted this. I reproduced the MIT license part from another test file I've touched, but removed the Microsoft part (I am not affiliated right now) and added my name as contributor, like it seems to be the procedure.

tlatools/org.lamport.tlatools/src/tlc2/value/ValueInputStream.java
```diff
@@ -91,7 +91,27 @@ public final IValue read() throws IOException {
...
+	 *   change community modules to call readExternal.
+	 * - Once all practically deployed community modules that use this method are phased out,
+	 *   delete it.
+	 */
+	@Deprecated
 	public final IValue read(final Map<String, UniqueString> tbl) throws IOException {
+		return readExternal();
+	}
+	
+	public final IValue readExternal() throws IOException {
```
Calvin-L, line 114: I would add a docstring to this method about the name and purpose of the method; something like
```java
/**
 * Safely read a value from an external source that may have a different intern table for UniqueStrings.
 * By contrast, {@link #read()} assumes that the value was written by this process using the same intern table.
 */
```

(Otherwise this change looks great to me! 👍 )

fhackett: Good point. I liked your phrasing, so I used it verbatim.

Calvin-L (approved):

lemmy:
@fhackett Some build checks failed.

fhackett:
@lemmy Curiously, it is precisely the TLA+ test I added. I can't see much in the CI log, but I strongly suspect it is because of the path format for the `.vos` file I tell it to load.

Is there any good approach to getting these paths right (e.g., making it a correct absolute path?) in TLA+ tests?

Otherwise, I'll try something to do with programmatically setting environment variables, or something like that.

lemmy:
`tlc2.tool.DumpAsDotTest` might serve as an example how to load a resource in a test.

fhackett:
@lemmy the example you linked is partially relevant, but ultimately I can't do anything similar for 2 different reasons:
1.  the whole point of the test is to do the lookup from TLA+ using IOUtils, so I can't use any of the known techniques for handling resources in Java etc. The trouble is getting IOUtils to use the right path using only a string from TLA+.
2.  IOUtils does not respect `util.ToolIO.userDir` (which can be set with the testing infrastructure, so it looked hopeful). It works with the actual current working directory the JVM happens to have (which is probably where `ant` was launched, or the dir with `customBuild.xml` when using Eclipse... they are clearly different whatever they are). This type of behavior is often fine actually during general use, because the current working directory is often "right here where I'm working"... but not this time.

I was hoping the current working directory was stably "the project dir", relative to which I could write a path, but it seems CI and local diverge.
Now of course, the docs of IODeserialize specifically indicate that the path should be absolute, so it's technically working as designed, and I was abusing that it allowed relative paths and resolved them using JVM behavior.

Changing (2) might be a future improvement, in that using `new File(util.ToolIO.getUserDir(), filename)` would make the IO functions more reliably look "next to the TLA+", which I imagine is what a user might expect... if I understand the behavior of userDir correctly, at least. With that, I could just write the `.vos` filename with no prefix and it would work.

I also looked into my idea about using environment variables, and while TLA+ code can read environment variables, it seems the testing infrastructure can't set them (because it uses an in-process TLC, right?), so I'm less confident that's a good idea.

I could of course add a hack that puts the path as `tlatools/org.lamport.tlatools/test-model/StringDeserialize.vos` instead of `test-model/StringDeserialize.vos`, but then it will break the local workspace tests, where the instructions say to open `tlatools/org.lamport.tlatools` in Eclipse, putting the working directory 2 dirs deeper than CI (well, what I assume is CI's working dir).

There might be something possible by faking `TLCSet` with a corrected path in the test, and then using `TLCGet` to access the value, but I feel like this message is getting long and I've ruled out / expressed doubt in quite a few alternatives by now.

Any thoughts, recommendations on direction, and/or did I completely miss something?
(I'm beginning to think altering lookup behavior in the community modules Java might be the "cleanest" solution myself)

lemmy:
`TLCSet` already [supports *setting* Java system properties](https://github.com/tlaplus/tlaplus/blob/142b1fb5a48be413b84a7f5daa7940d0b66581e1/tlatools/org.lamport.tlatools/src/tlc2/module/TLCGetSet.java#L488-L491).  We could also add support for `TLCGet` to support them.  For example, `TLCGet("-DpathToTestInput")` could then be used to retrieve the path to the test input.

fhackett:
@lemmy that sounds like a good idea. I'll try that relatively soon. I will be traveling / on break, however, so it will take me a "couple of" days to get back to this.

lemmy:
https://github.com/tlaplus/tlaplus/pull/1105

fhackett:
@lemmy I found a minute to fix up the PR. Thanks for implementing the extra feature - the tests should hopefully pass now (they still do on my machine, and they use BASE_PATH to lookup the `.vos` file now so working directories shouldn't matter anymore).

Also, happy new year :tada: .

PR #1106, opened 2025-01-01
SANY: Finally remove global static semantic error log

ahelwer:
Finally remove the necessity of calling `SemanticNode.setError()` before doing any parsing. One incantation down, many more to go.

The only place this changes behavior is in the SANY parse tree debug explorer, where instead of the `ModuleNode.toString()` method accessing the static semantic error log to print out some stuff, we move that access into the `Explorer` class itself. This slightly changes output; for example currently, if parsing the following module in SANY with the `-d` flag to open the debug prompt:
```tla
---- MODULE Test ----
VARIABLE x
op ≜ ◇[1]_x
====
```
then running `d Test` or `304` (the module UID) in the debug shell will print out:
```
*ModuleNode: Test    uid: 304  kind: ModuleKind  constant module: false  errors: 1
| Context Entry: x  305 
| | *OpDeclNode: x    uid: 305  kind: VariableDeclKind  arity: 0  orgDefInModule: Test
| |   originallyDefinedInModule: Test
...
```
but after these changes it will print out:
```
Constant module: false
Module errors: 1
*ModuleNode: Test    uid: 304  kind: ModuleKind
| Context Entry: x  305 
| | *OpDeclNode: x    uid: 305  kind: VariableDeclKind  arity: 0  orgDefInModule: Test
| |   originallyDefinedInModule: Test
...
```
To my knowledge no tool actually depends on the precise text output of the SANY debug shell so this should be fine (actually the specific debug shell command used here isn't even documented in the SANY CLI help text).

Ref #891 #1101 #1100

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/explorer/Explorer.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/explorer/Explorer.java
@@ -19,7 +19,6 @@ import java.util.StringTokenizer;
 import tla2sany.semantic.Errors;
 import tla2sany.semantic.ExternalModuleTable;
 import tla2sany.semantic.FormalParamNode;
-import tla2sany.semantic.Generator;
 import tla2sany.semantic.OpDefOrDeclNode;
 import tla2sany.semantic.SemanticNode;
 import tla2sany.semantic.SymbolNode;
@@ -102,7 +101,7 @@ public class Explorer {
 		// See if the object requested is already in the table
 		if ((obj = (ExploreNode) semNodesTable.get(icmd)) != null) {
 			// Print tree to depth of icmd2
-			System.out.println(((ExploreNode) obj).toString(depth));
+			System.out.println(((ExploreNode) obj).toString(depth, errors));
 			System.out.print("\n" + ((ExploreNode) obj).levelDataToString());
 		} else {
 			// object requested is not in semNodesTable
@@ -170,8 +169,7 @@ public class Explorer {
 			} else if (sym instanceof FormalParamNode) {
 				System.out.print("Module: " + ((FormalParamNode) sym).getModuleNode().getName());
 			}
-			System.out.println(((ExploreNode) (symbolVect.elementAt(i))).toString(100));
-			System.out.println();
+			System.out.println(((ExploreNode) (symbolVect.elementAt(i))).toString(100, errors) + "\n");
 		}
 
 	}
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
@@ -380,7 +380,7 @@ public abstract class SemanticNode
    * of SemanticNode for implementing ExploreNode interface; the depth
    * parameter is a bound on the depth of the tree that is converted to String.
    */
-  public String toString(int depth) {
+  public String toString(int depth, Errors errors) {
     if (depth <= 0) return "";
     return ("  uid: " + myUID +
 	    "  kind: " + (kind == -1 ? "<none>" : kinds[kind])
@@ -505,7 +505,7 @@ public abstract class SemanticNode
      */
 
   protected Element getSemanticElement(Document doc, tla2sany.xml.SymbolContext context) {
-      throw new UnsupportedOperationException("xml export is not yet supported for: " + getClass() + " with toString: " + toString(100));
+      throw new UnsupportedOperationException("xml export is not yet supported for: " + getClass());
     }
 
     /**
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
@@ -41,8 +41,6 @@ public abstract class SemanticNode
 
   private static final AtomicInteger uid = new AtomicInteger();  // the next unique ID for any semantic node
 
-  protected static Errors errors;
-
   public    final int      myUID;    // the unique ID of THIS semantic node
   public    TreeNode stn;      // the concrete syntax tree node associated with THIS semantic node
   private   Object[] tools;    // each tool has a location in this array where
@@ -56,8 +54,6 @@ public abstract class SemanticNode
     this.stn = stn;
     this.tools = EmptyArr;
   }
-
-  public static void setError(Errors errs) { errors = errs; }
 
   public static String levelToString(int level) {
     switch (level) {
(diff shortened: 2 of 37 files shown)


tlatools/org.lamport.tlatools/src/tla2sany/explorer/Explorer.java
```diff
@@ -96,12 +97,27 @@ private boolean getLine() {
...
+
 	// Integer command
 	private void printNode(int depth) {
 
 		// See if the object requested is already in the table
 		if ((obj = (ExploreNode) semNodesTable.get(icmd)) != null) {
 			// Print tree to depth of icmd2
+			if (obj instanceof ModuleNode) {
+				printModuleNodeDetails((ModuleNode)obj);
+			}
```
Calvin-L, line 120: > instead of the `ModuleNode.toString()` method accessing the static semantic error log to print out some stuff, we move that access into the `Explorer` class itself

An alternative would be to add `Errors` as a parameter to the `toString(depth)` method:

```java
public String toString(int depth, Errors errors) { ... }
```

Although that would result in more lines changed, I think it's cleaner in other ways:
 - SANY's output wouldn't change
 - No awkward `instanceof` checks

Virtually all of the calls to `toString(depth)` are recursive calls, so there are actually very few "real" changes to introduce that parameter.

Calvin-L: (If you do decide to do it that way, which I endorse, there are exactly two awkward call sites that look like `throw new UnsupportedOperationException("xml export is not yet supported for: " + getClass() + " with toString: " + toString(100));` -- these are dead code, so removing the `toString(100)` call should be fine.)

tlatools/org.lamport.tlatools/src/tla2sany/explorer/Explorer.java
```diff
@@ -96,12 +97,27 @@ private boolean getLine() {
...
 	}
 
+	/**
+	 * Print diagnostics about the given module that require access to the
+	 * semantic error log.
+	 *
+	 * @param node The module to print info about.
+	 */
+	private void printModuleNodeDetails(ModuleNode node) {
+		System.out.println("Constant module: " + node.isConstant(this.errors));
```
lemmy, line 107: This uses `System.out`, whereas the code base generally uses `ToolIO.out`.  IIRC, the latter works with the Toolbox.

tlatools/org.lamport.tlatools/src/tla2sany/explorer/Explorer.java
ahelwer: I started with that approach but there are *a lot* of `toString(int)` overloads would have to be changed, in all the various node classes; also `toString(int)` is only ever used by the `Explorer` class and the `Errors` is only ever used by `ModuleNode`.

lemmy: Eclipse `(Refactor > Change Method Signature)` (and probably other IDEs as well) can automate such a refactoring.

tlatools/org.lamport.tlatools/src/tla2sany/explorer/Explorer.java
```diff
@@ -170,7 +170,7 @@ private void lookUpAndPrintDef(String symbName) {
 			} else if (sym instanceof FormalParamNode) {
 				System.out.print("Module: " + ((FormalParamNode) sym).getModuleNode().getName());
 			}
-			System.out.println(((ExploreNode) (symbolVect.elementAt(i))).toString(100));
+			System.out.println(((ExploreNode) (symbolVect.elementAt(i))).toString(100, errors));
```
lemmy, line 173: Consider logically merging the subsequent `System.out.println();` statement into this line.

tlatools/org.lamport.tlatools/test/util/ParserAPI.java
```diff
@@ -130,7 +129,6 @@ public static ModuleNode processSemantics(SyntaxTreeNode parseTree, Errors log)
    * @return Whether levels in parse tree are correct.
    */
   public static boolean checkLevel(ModuleNode semanticRoot, Errors log) {
```
lemmy, line 131: Is it now possible to refactor this into a method of ModuleNode?

lemmy: By the way, (in a separate PR) the function `processSemantics` could be moved into `ModuleNode` as e.g. `ModuleNode#create`, `processSyntax` could be moved into `SyntaxTreeNode`, and `parse` could be moved into `SANY`.

ahelwer: It is a trivial enough wrapper here that `ParserAPI.checkLevel` could be done away with, but if we are designing the `ParserAPI` class to eventually be the primary API through which developers use SANY, it makes sense to expose `checkLevel` as one of a small number of functions in this class for discoverability purposes - otherwise, consumers would have to know about the `ModuleNode.checkLevel` method. It's just a discoverability issue for such an important function.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
```diff
@@ -57,7 +55,16 @@ public SemanticNode(int kind, TreeNode stn) {
...
+  /**
+   * Set the global static error log for semantic & level checking.
+   *
+   * @param errs The error log instance to use.
+   * @deprecated This method is no longer necessary; all uses of the global
+   * static error log have been removed in favor of passing in the error log
+   * instance at the start of the semantic & level checking processes.
+   */
+  @Deprecated
+  public static void setError(Errors errs) { }
```
lemmy, line 67: Why are we keeping `setError`?  Apalache, the only known consumer of SANY, [doesn't seem to call it](https://github.com/search?q=repo%3Aapalache-mc%2Fapalache%20setError&type=code).

tlatools/org.lamport.tlatools/test/util/ParserAPI.java
lemmy: Personally, I don’t find a utility class discoverable, but it’s pointless to discuss discoverability since we lack an objective measure.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/SemanticNode.java
ahelwer: My experimental interpreter repo does call it, but sure we can just remove it and I'll remove the call from there whenever the snapshot maven package is updated.

tlatools/org.lamport.tlatools/test/util/ParserAPI.java
ahelwer: I'm envisioning this class would eventually graduate to living in `src/tla2sany/API.java` or similar, which might help.

Issue #545, opened 2020-11-30
TLC checks wrong TLA file when TLA file of same name is in current working directory


ahelwer:
**Version Info**

TLC version: 11/24/2020 nightly build (also 1.7.0 release)
JRE version:  AdoptOpenJDK build 15.0.1+9
OS version: Windows 10

**What happened? What did you expect to happen?**

I had the following directory structure:

```
.
+-- Test.tla
+-- Test.cfg
+-- test
|   +-- Test.tla
|   +-- Test.cfg
```
I ran TLC from the top-level directory containing `Test.tla` with the following command:
```
java tlc2.TLC -config .\test\Test.cfg .\test\Test.tla
```
I expected it to model check `test\Test.tla`, but instead it model checked the top-level `Test.tla` file.

**Steps to Reproduce**

 1. Download the TLA/CFG files with above directory structure:  [cwdtlafailure.zip](https://github.com/tlaplus/tlaplus/files/5618848/cwdtlafailure.zip)
 2. Execute the following commands:
   - In `test` directory: `java tlc2.TLC Test.tla`
   - In top-level directory: `java tlc2.TLC -config .\test\Test.cfg .\test\Test.tla`

Note the specs are identical except the top-level `Test.tla` uses a variable named `y` instead of `x`.

lemmy:
This won't be a regression because relative paths is a recent feature addition.

ahelwer:
Addition: this error also occurs if the `Test.tla` file is present in any of TLC's search paths as set in `TLC.setResolver` or via environment variable.

Calvin-L:
I'm not able to reproduce this on current master. Bisect reports that it was fixed in d3accfcee8932f3775c680f81aed9b84eead42dc.

PR #1116, opened 2025-01-10
Refactor `SimpleFilenameToStream` for correctness and clarity

### LINKED ISSUE #1114: TLC's module resolution relies on undefined/undocumented behavior from the Java standard library
## Description
While [investigating SANY's performance](https://github.com/tlaplus/rfcs/issues/16#issuecomment-2563975827) I experimented with new ways of building the TLA+ tools: [Graal native images](https://www.graalvm.org/22.1/reference-manual/native-image/index.html) and [jlink](https://docs.oracle.com/en/java/javase/11/tools/jlink.html). (The two have little to do with each other, but they are both ways to reduce the overhead of JVM startup.)

Both approaches require substantial patches to `SimpleFilenameToStream`, or else the compiled tools cannot find the standard library modules. `SimpleFilenameToStream` is responsible for finding TLA+ modules on disk. There are two latent bugs in `SimpleFilenameToStream`, both relating to [`ClassLoader.getResource()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ClassLoader.html#getResource(java.lang.String)):

1. The returned URL is not guaranteed to be a `jar:` or `file:` URL. The current code assumes that it is one of those two. In Graal for instance, the returned URL starts with `resource:`.
2. There is no guarantee that resolving a directory like `tla2sany/StandardModules/` produces anything at all. The docs simply say that `getResource()` "finds the resource with the given name. A resource is some data (images, audio, text, etc)" --- and indeed, in a jlink'd executable, `getResources("dir/")` always returns `null`, even when `getResources("dir/file")` works.

## Steps to Reproduce

**Graal**

First patch `TLAFlightRecorder.java`, since Graal does not support JFR:

```diff
diff --git a/tlatools/org.lamport.tlatools/src/util/TLAFlightRecorder.java b/tlatools/org.lamport.tlatools/src/util/TLAFlightRecorder.java
index a2ca81232..07ca3fd53 100644
--- a/tlatools/org.lamport.tlatools/src/util/TLAFlightRecorder.java
+++ b/tlatools/org.lamport.tlatools/src/util/TLAFlightRecorder.java
@@ -27,7 +27,8 @@ package util;
 
 // Label is only used by Mission control but not relevant for the recording itself.
 public class TLAFlightRecorder {
-       
+
+       /*
     @jdk.jfr.Label("Progress")
     @jdk.jfr.Category({"TLC", "Progress"})
     @jdk.jfr.StackTrace(false)// No need to capture the stack of the reporting (main) thread when it emits the event.
@@ -89,4 +90,19 @@ public class TLAFlightRecorder {
                }
                return message;
        }
+        */
+
+       public static void progress(
+                       boolean isFinal,
+                       final int diameter,
+                       final long states,
+                       final long distinctStates,
+                       final long unseen,
+                       final long statesPerMinute,
+                       final long distinctStatesPerMinute) {
+       }
+
+       public static String message(final String message) {
+               return message;
+       }
 }
\ No newline at end of file
```

Build the tools, then:

```
$ native-image \
    -cp dist/tla2tools.jar:lib/CommunityModules.jar \
    -H:+UnlockExperimentalVMOptions \
    -H:-CheckToolchain \
    -H:IncludeResources=.* \
    -Os \
    tla2sany.SANY \
    sany-native

$ cd ../../general/performance/Bloemen

$ ../../../tlatools/org.lamport.tlatools/sany-native -DTLA-Library=.. MC.tla 

****** SANY2 Version 2.2 created 08 July 2020

resource:/tla2sany/StandardModules/
java.lang.IllegalArgumentException: URI scheme is not "file"
	at java.base@23.0.1/java.io.File.<init>(File.java:423)
	at util.SimpleFilenameToStream.getInstallationBasePath(SimpleFilenameToStream.java:89)
	at util.SimpleFilenameToStream.<init>(SimpleFilenameToStream.java:53)
	at util.ToolIO.setDefaultResolver(ToolIO.java:217)
	at util.ToolIO.getDefaultResolver(ToolIO.java:204)
	at tla2sany.modanalyzer.SpecObj.<init>(SpecObj.java:129)
	at tla2sany.drivers.SANY.SANYmain(SANY.java:415)
	at tla2sany.SANY.main(SANY.java:12)
	at java.base@23.0.1/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH)
Parsing file /Users/cloncaric/src/tlaplus/general/performance/Bloemen/MC.tla
Parsing file /Users/cloncaric/src/tlaplus/general/performance/Bloemen/BloemenSCC.tla

Fatal errors while parsing TLA+ spec in file MC.tla

tla2sany.semantic.AbortException
*** Abort messages: 1

Unknown location

Cannot find source file for module TLC imported in module MC.
```

**jlink**

This one is tricky. Make a fake module-info.java file:

```
open module org.lamport.tlatools {
}
```

```
$ javac module-info.java

$ jar cf module-info.jar module-info.java

$ ... build combined.jar out of module-info.jar, dist/tla2tools.jar, and lib/CommunityModules.jar ...

$ jlink \
        --module-path ./combined.jar \
        --add-modules org.lamport.tlatools \
        --launcher sany=org.lamport.tlatools/tla2sany.SANY \
        --compress=2 \
        --no-header-files \
        --no-man-pages \
        --output ./sany-jlink

$ cd ../../general/performance/Bloemen

$ ../../../tlatools/org.lamport.tlatools/sany-jlink/bin/sany MC.tla                 

****** SANY2 Version 2.2 created 08 July 2020

Exception in thread "main" java.lang.NullPointerException
	at org.lamport.tlatools/util.SimpleFilenameToStream.getInstallationBasePath(SimpleFilenameToStream.java:87)
	at org.lamport.tlatools/util.SimpleFilenameToStream.<init>(SimpleFilenameToStream.java:53)
	at org.lamport.tlatools/util.ToolIO.setDefaultResolver(ToolIO.java:217)
	at org.lamport.tlatools/util.ToolIO.getDefaultResolver(ToolIO.java:204)
	at org.lamport.tlatools/tla2sany.modanalyzer.SpecObj.<init>(SpecObj.java:137)
	at org.lamport.tlatools/tla2sany.drivers.SANY.SANYmain(SANY.java:477)
	at org.lamport.tlatools/tla2sany.SANY.main(SANY.java:12)

```

## Expected Behavior
The tools should work the same whether they are built with javac+jar, Graal, or jlink.

## Steps Taken to Fix
There is no real workaround without patching the code. I have a set of patches already; I will be cleaning them up and opening a PR or two over the next week or so.

## Your Environment
 - TLC version: [commit 40b92d0e](https://github.com/tlaplus/tlaplus/commit/40b92d0e9ad89edc8ed1400b692f2cf88e151691)
 - JDK version: OpenJDK 11
 - Operating System: MacOS 14

Calvin-L:
Fixes #1114.

This change has multiple purposes.  The primary purpose is to allow `SimpleFilenameToStream` to resolve the standard library files in other environments, such as when the tools are compiled as a Graal native image or as a jlink'd executable.

Previously `SimpleFilenameToStream` relied on resolving the resource folder `tla2sany/StandardModules` to a URL and then examining the URL's scheme, which is not defined behavior:

 - In Graal the scheme is `resource:` and not `jar:`
 - In jlink the directory always resolves to `null`, even if entries within it are accessible.

To fix the problem, the search logic in `SimpleFilenameToStream` has been broken out into different "Resolver" classes, each of which does very little, but which can be composed into a sequence of locations to search.  (I had originally thought to make a set of composable implementations of `FilenameToStream`, but the contract for that interface has too many warts.  It was much easier to replace the internals of `SimpleFilenameToStream` than to rework the contracts of existing methods, especially since the Toolbox uses `FilenameToStream`.)

The fix is in `ClasspathResolver.java`, where resources are resolved directly using `classLoader.getResource(...)` and there is no "pre-resolve" step on the prefix directory.

As a secondary purpose, I believe these changes also improve the clarity of TLC's module resolution process---or at least, they eliminate a lot of bizarre behaviors:

 - Clearer typing (everything is now URLs instead of Strings-that-are- either-URLs-or-paths).
 - Fewer resource lookups (the old code ran `getResource` _and_ `getResourceAsStream` after examining the prefix string).
 - Reduced redundant calls to `File.exists()` (the `read()` method will always return an existing file).
 - Straightforward for-each loops (no special-casing on index 0, no special case on libaryPaths.length-1, no reliance on the last lookup path equaling the standard module folder, etc).

We have good test coverage of `SimpleFilenameToStream`, and the existing tests pass, along with some new ones that pass both before and after this change.

There is a bonus fix here as well: `isStandardModule` now works even when TLC is running out of a jar, and its implementation has been simplified to reduce duplication with `TLAFile.isLibraryModule()`.


tlatools/org.lamport.tlatools/src/util/Resolver.java
```diff
@@ -0,0 +1,34 @@
...
+ * Resolves abstract filenames (e.g. <code>"Integers.tla"</code>) to concrete locations
+ * (e.g. <code>"jar:file:/path/to/tla2tools.jar!/tla2sany/StandardModules/Integers.tla"</code>).
+ *
+ * <p>This interface is simple enough to be composable: multiple resolvers can be chained together
+ * using {@link SequentialResolver}.
+ */
+interface Resolver {
+
+    /**
+     * Locate the given file.
```
ahelwer, line 19: Add a note emphasizing that this function expects `ModuleName.tla`, not `ModuleName`.

tlatools/org.lamport.tlatools/src/util/SequentialResolver.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * A resolver that tries several other resolvers in sequence.  The first found result is returned.
+ */
+class SequentialResolver implements Resolver {
+
+    private final List<Resolver> resolvers;
+
+    public SequentialResolver(List<Resolver> resolvers) {
```
ahelwer, line 18: Might be a nicer API to use `Resolver... resolvers`?

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
```diff
@@ -65,66 +70,43 @@ public SimpleFilenameToStream(final String libraryPath) {
...
-    	e.printStackTrace();
-    } catch (IllegalArgumentException e) {
-    	System.err.println(path);
-    	// may never happen
-    	e.printStackTrace();
-    }
-    return path;
-   }
+  private static Resolver getInstallationBasePath() {
+    return new ClasspathResolver(cl, STANDARD_MODULES);
```
ahelwer, line 83: If this is the only consumer of `STANDARD_MODULES`, consider renaming it to `STANDARD_MODULES_INTERNAL_PATH` or similar.

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
```diff
@@ -136,112 +118,94 @@ private String[] getLibraryPaths(final String installationBasePath, String[] lib
...
+          res.add(new FilesystemResolver(Path.of(s)));
       }
-      res[paths.length] = installationBasePath + FileUtil.separator;
     }
-    return res;
+
+    // Finally, search the bare classpath (e.g. for the community modules)
+    res.add(new ClasspathResolver(cl, ""));
+
+    return new SequentialResolver(res);
```
ahelwer, line 137: Since this line will fundamentally be what we look at in the future when figuring out the resolver search order, I think using the `Resolver...` API suggested above would make it very clear.

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
```diff
@@ -136,112 +118,94 @@ private String[] getLibraryPaths(final String installationBasePath, String[] lib
...
   }
 
   /**
-   *  From name, prepend path elements to try to find the actual
-   *  file intended, and use the resulting fully-qualified name to initialize
-   *  the File.  The first fully-qualified name that refers to a file that actually
-   *  exists is the File returned; but if none exists, the last one tried is returned
+   *  From name, try to find the actual file intended.  If the file is not a normal
```
ahelwer, line 141: Also add note whether the function expects a `.tla` suffix or not.

tlatools/org.lamport.tlatools/src/util/SequentialResolver.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public Optional<URL> resolve(String name) throws IOException {
+        for (Resolver resolver : resolvers) {
+            Optional<URL> resolved = resolver.resolve(name);
+            if (resolved.isPresent()) {
+                return resolved;
+            }
+        }
+        return Optional.empty();
```
ahelwer, line 30: Could this be replaced with something like:
```
resolvers.stream()
.map(resolver -> resolver.resolve())
.flatMap(Optional::stream)
.findFirst();
```

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
```diff
@@ -65,66 +70,43 @@ public SimpleFilenameToStream(final String libraryPath) {
...
    * added functionality for supplying additional libraries.
    * All usage of this except the one added by TL is supplying libraries  = null
    */
-  private String[] getLibraryPaths(final String installationBasePath, String[] libraries) {
-    String[] res;
-    String path = null;
+  private static Resolver getLibraryPaths(String[] libraries) {
+    List<Resolver> res = new ArrayList<>();
+
+    // First, search the current directory
```
ahelwer: Does the current behavior search the CWD or the directory that the root spec is in? I recall I opened an issue about this leading to some confusing SANY/TLC behavior a long time ago... #545

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
```diff
@@ -136,112 +118,94 @@ private String[] getLibraryPaths(final String installationBasePath, String[] lib
...
+    Optional<TLAFile> libraryModule = locateUsingResolver(standardLibraryResolver, name, true);
+    if (libraryModule.isPresent()) {
+        return libraryModule.get();
+    }
 
-    // The TLA standard modules are presumed to be in a directory tla2sany/StandardModules
-    // i.e. a toplevel subdirectory of the main installation directory for SANY
-    // We append the StandardModules director to the library path, so the user does
-    // not have to do anything special and the standard modules are automatically accessible
+    // Small bit of awkwardness here: this method requires us to return a nonexistent file if the requested one
```
ahelwer, line 164: Add note if this is for backwards compatibility reasons or what. Can this be fixed at the callsite?

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
```diff
@@ -309,24 +273,15 @@ public File resolve(String name) {
 	}
 
 	/**
-	 * Returns true iff moduleName is the name of a standard module.  It
-	 * assumes the empirically determined fact that libraryPaths is set to
-	 * have the path of the standard modules as the last element of the array.
+	 * Returns true iff moduleName is the name of a standard module.
```
ahelwer, line 282: Add documentation regarding whether `.tla` suffix is expected.

tlatools/org.lamport.tlatools/test/util/SimpleFilenameToStreamTest.java
```diff
@@ -186,6 +182,7 @@ public void testTLALibrarySystemProperty() throws IOException  {
 			System.setProperty(SimpleFilenameToStream.TLA_LIBRARY, d1.toString());
 			File f1 = resolver.resolve("Integers", true);
 			assertTrue("Resolver should still find Integers when library path is empty", f1.exists());
+			assertTrue("Integers is a standard module, even with TLA_LIBRARY set", resolver.isStandardModule("Integers"));
 
 			Files.writeString(d1.resolve("MyModule.tla"), "", StandardCharsets.US_ASCII);
```
ahelwer, line 208: Some additional test cases based off weird behavior I've seen in SANY and TLC:
- The example of #545
- I've noticed that the current resolver won't find modules inside .jar files unless the .jar files go first in the classpath (which functions as SANY's search path currently); is that still the case or do these changes incidentally fix that?

ahelwer (approved):
Overall design looks good, and a big improvement in how understandable the code is.

tlatools/org.lamport.tlatools/src/model/InJarFilenameToStream.java
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2023, Oracle and/or its affiliates.
+// Copyright (c) 2023, 2025, Oracle and/or its affiliates.
```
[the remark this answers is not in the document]
Calvin-L: The instructions from my company are that this should include two years: the first year we touched it and the most recent year we touched it. I'm not sure why, but it's an easy enough rule to follow.

tlatools/org.lamport.tlatools/test/util/SimpleFilenameToStreamTest.java
Calvin-L: My goal is not to change any of TLC's weird module resolution rules, just to fix misuses of the Java resource API and do some cleanup. I might add a test case ensuring that this change _preserves_ any weird behavior from #545. 😆 

For your second point: maybe. See if #977 (merged in June last year) fixes the problem. If it doesn't, then this PR might incidentally.

tlatools/org.lamport.tlatools/src/util/SequentialResolver.java
Calvin-L: That's a convenient approach if the list of resolvers is known at compile-time, but `SimpleFilenameToStream` actually builds a variable-length list depending on some checks and run-time arguments. Such a constructor wouldn't simplify anything today, so I probably won't do this.

(A trio of personal preferences are also seeping in here: One, I generally prefer _a single_ constructor with a clear contract rather than many "convenience" constructors that easily get out-of-sync. Two, I generally prefer to push data conversion tasks onto the caller; the list is stored as a list, so the caller has the job of constructing the list. Three, since performance is not a consideration I find `List` nicer than `[]`; the `...` syntax gives you an array of arguments, putting you in warty array-land. The client can always just do `new SequentialResolver(List.of(r1, r2, r3))`.)

tlatools/org.lamport.tlatools/src/util/Resolver.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Optional;
+
+/**
+ * Resolves abstract filenames (e.g. <code>"Integers.tla"</code>) to concrete locations
+ * (e.g. <code>"jar:file:/path/to/tla2tools.jar!/tla2sany/StandardModules/Integers.tla"</code>).
+ *
+ * <p>This interface is simple enough to be composable: multiple resolvers can be chained together
+ * using {@link SequentialResolver}.
+ */
+interface Resolver {
```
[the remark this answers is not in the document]
Calvin-L: I agree, the name is not great. How about `ResourceLocator`? Resolver seems like a vague word anyway.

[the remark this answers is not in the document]
Calvin-L: Or `FilenameToURL`?

tlatools/org.lamport.tlatools/src/util/ClasspathResolver.java
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2025, Oracle and/or its affiliates.
```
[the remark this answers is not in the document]
Calvin-L: I agree about the location. But, for now these are just an internal (package-private) implementation detail for `SimpleFilenameToStream`, so they should be colocated with that class. These will be easy to move if/when we move the various `FilenameToStream` classes.

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
Calvin-L: I once tried to change the contract here (file-that-might-or-might-not-exist -> `Optional<File>`). There are many callsites to change, and some of them are in the Toolbox.

I will add a note that this is for historical reasons. 👍 For now, at least the contract is well-documented in `FilenameToStream`.

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
Calvin-L: I made the minimal possible change to this comment to keep it valid. A clearer contract is in the top-level interface `FilenameToStream`:
```
     * Returns true iff <code>moduleName</code> resolves to a standard module.  The argument is treated the same as a
     * call to {@link #resolve(String, boolean)} with <code>isModule=true</code>.
```

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
Calvin-L: I won't add a `...` constructor (see notes on the other comment).

BUT, I am very sympathetic to "this particular block needs to be as clear as possible". What I will do is promise a follow-up PR to improve the clarity of this section. The final code might look something like
```java
return new SequentialResolver(List.of(
    searchUserDirOrCWD(),
    searchFilesystemPaths(libraries),
    searchBareClasspath()));
```

(This block of code is in need of other changes as well which are best addressed in a separate PR. For instance, it constructs a `":"`-separated String, only to immediately destruct that String to get the constituent paths.)

tlatools/org.lamport.tlatools/src/util/SequentialResolver.java
Calvin-L: Ideally yes, but since `resolve` throws `IOException` it cannot be used cleanly with Java's functional stream API. The for-loop is less elegant for sure, but it does not require exception handling (and hey, it's probably a little more efficient too).

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
Calvin-L: Thanks a ton for pointing this out. See my reply on [your other comment about this](https://github.com/tlaplus/tlaplus/pull/1116#discussion_r1911428941).

tlatools/org.lamport.tlatools/test/util/SimpleFilenameToStreamTest.java
Calvin-L: It seems that #545 has been fixed. I've added a regression test to this PR so we do not break that functionality in the future.

lemmy: > * I've noticed that the current resolver won't find modules inside .jar files unless the .jar files go first in the classpath (which functions as SANY's search path currently); is that still the case or do these changes incidentally fix that?

This is the intended behavior because it would be quite unusual to load the module override from the jar file but not the corresponding specification.

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
Calvin-L: This is a fine suggestion, but I think I'm going to defer it. The diff in this PR is big enough already. 😓

tlatools/org.lamport.tlatools/src/util/Resolver.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Locate the given file.  The name is treated as a filename, not a TLA+ module name;
+     * if you are looking for the Integers module, pass <code>"Integers.tla"</code>.
+     *
+     * @param name the file to locate, e.g. <code>"Integers.tla"</code>
+     * @return a URL to the file, or {@link Optional#empty() empty} if it could not be found
+     * @throws IOException if an error occurs while searching
+     */
+    Optional<URL> resolve(String name) throws IOException;
```
lemmy, line 26: rename `name` to `filename`?

lemmy: The majority of the codebase uses `null` instead of `Optional` (except for two places outside of `test/` that escaped my review years ago). If you anticipate the `Resolver` interface to be used beyond `SimpleFilenameToStream`, I recommend aligning with the rest of the codebase.

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
```diff
@@ -65,66 +70,53 @@ public SimpleFilenameToStream(final String libraryPath) {
...
+    } else {
+      String cwd = System.getProperty("user.dir");
+      if (cwd != null) {
+        res.add(new FilesystemResolver(Path.of(cwd)));
+      }
+    }
+
+    // Next, search the given `libraries` (or the TLA_LIBRARY system property if `libraries` is null).
+    String path; // actually a list of paths separated by FileUtil.pathSeparator
     if (libraries == null) path = System.getProperty(TLA_LIBRARY);
```
lemmy, line 120: Perhaps you could also add `{…}` while you’re at it.

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
```diff
@@ -136,112 +128,96 @@ private String[] getLibraryPaths(final String installationBasePath, String[] lib
...
-    // i.e. a toplevel subdirectory of the main installation directory for SANY
-    // We append the StandardModules director to the library path, so the user does
-    // not have to do anything special and the standard modules are automatically accessible
+    // Small bit of awkwardness here: to preserve the historical FilenameToStream API, this method must return a
+    // nonexistent file if the requested one could not be located.  The file we return will be used in error messages
+    // about the missing file, so we should return one that looks plausible, but we can't return one that MIGHT exist,
+    // because the result will be read if it does!  The approach below correctly synthesizes a non-existent
+    // believable-looking path, but relies on the fact that ToolIO.getUserDir() is in the search paths used by
+    // userFileResolver---this file does not exist because it has certainly already been searched.  In the future, we
+    // might improve this pattern by changing the FilenameToStream contract to say that it throws an exception or
```
lemmy, line 174: This brings back unpleasant memories because the Toolbox suffered and suffers from this awkwardness.  To make matters worse, the Toolbox additionally accesses files through Eclipse's FS abstraction.  Instead of refactoring `FilenameToStream`, it would likely be better to deprecate it and create a new API in tla2tools.

lemmy (approved):
My biggest concern is that this refactoring breaks the `InJarFilenameToStream`, which has zero test coverage.  That said, I don't think that that functionality is widely used.

tlatools/org.lamport.tlatools/src/util/Resolver.java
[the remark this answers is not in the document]
ahelwer: Any of these would work! FilenameResolver maybe? I think that captures that it is resolving a filename. It's a bit similar to the name of the implementing class FilesystemResolver though.

ahelwer:
Is there a way you could expose a simple module name -> `InputStream` method in these classes, reusing most of the same module resolution logic? This would open the door to slowly transitioning the tools onto `InputStream`. If that's too much work it can be moved into another PR perhaps.

Calvin-L:
> Is there a way you could expose a simple module name -> `InputStream` method in these classes

It's not too much work; the new classes return `URL`, and [`URL` is already `InputStream`-adjacent](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URL.html#openStream()).

But it feels like a separate task. There are some nontrivial API considerations that I haven't fully thought through. 

((Here are the thoughts I have right now:

- `InputStream` isn't enough. We also need
  - Human-readable info about where the file came from (for informational output and error messages)
  - Information about whether the file is one of the standard modules (used by the Toolbox in ways I don't yet grok, and possibly useful to IDEs etc)
  - Maybe a way to resolve "nearby" objects---e.g. after finding a .tla file, you might want to find the .cfg file in the same directory, or the .class file in the same jar that provides operator overrides. See `TLAFile`'s `resolver` field.
- `InputStream` can be awkward because it holds a physical resource that must be [closed](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Closeable.html); it might be better to spit out something closer to plain-old-data (maybe augmented with an `open()` method for obtaining an `InputStream`, like `URL` already has?)
- `NamedInputStream` and `TLAFile` are two attempts to encapsulate the necessary information; perhaps we could combine them.
- Distributed TLC has special needs as well. (I know [it's on the chopping block](https://github.com/tlaplus/tlaplus/issues/718), but there's a part of me that wants to try and preserve it.)

))

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
Calvin-L: I'll do that in a follow-up PR, since this method deserves some additional attention anyway (see [related comment](https://github.com/tlaplus/tlaplus/pull/1116#discussion_r1911629890)).

tlatools/org.lamport.tlatools/src/util/SimpleFilenameToStream.java
Calvin-L: Agreed. I don't know exactly what that should look like, but I can at least amend this comment to include that idea.

ahelwer:
Much of this can be trusted to the caller; the `InputStream` instance only needs to live long enough to be sent into `TLAplusParser` to get the syntax parsing done, then it can be closed. I don't think wrapping it into a `NamedInputStream` with some associated data is too important; all that info is available to the caller who called the resolver to get the InputStream. So it would look like:
```java
TreeNode root = null;
try (InputStream input = resolver.resolve(filename) {
  TLAplusParser parser = new TLAplusParser(input, StandardCharsets.UTF_8.name());
  parser.parse();
  root = parser.ParseTree;
} finally {
  input.close();
}
// check whether filename matches module name etc. here
```

lemmy:
> Lifetime management can be trusted to the caller

Anything can be trusted to the caller, but this often leads to duplication and inconsistencies.

> Actually this idea seems good, keeping the file contents around as a string. It could even give us the ability to eventually output nice rust-style error messages printing out the user's code with an arrow pointing to their error.

Are you suggesting a string encapsulated within a dedicated type?  We should avoid passing raw strings around, as this has already caused issues elsewhere in the codebase.  For example, there have been numerous error-prone conversions between `File` and `String` (e.g. chopping off the extension, appending it again, ...).

TLA+ files are relatively small, and it’s a reasonable assumption that their size will never exceed a few megabytes.  This means we can safely keep their contents in memory.   In other words, there can be a type that caches the file content in memory after the InputStream has been closed.  That type probably shouldn't extend `java.util.File`.

ahelwer:
Yes, definitely we should encapsulate the string inside a type. I am imagining resolve will return a class instance containing:
 - The expected module name (given as input to resolve)
 - The module contents as a string
 - An enum indicating the module source (filesystem, standard modules, jar on classpath, user-supplied string, etc.)
 - A path (or URL?) that will be valid for some subset of those sources (can't wait till we get those sweet sweet variant types)

Calvin-L:
> My biggest concern is that this refactoring breaks the `InJarFilenameToStream`, which has zero test coverage.

This is a good point, and I don't have a great answer. I will do a little manual testing before merging this PR.

What I can say is that
 - The logic in `InJarFilenameToStream` has not changed, nor have any of the helpers it uses
 - ...except the calls to `super.resolve()`, which direct to `SimpleFilenameToStream`.

So as long as the changes to `SimpleFilenameToStream` are correct, then I believe we can trust `InJarFilenameToStream` will also continue to work.

Calvin-L:
I confirmed manually that `InJarFilenameToStream` still seems to work:
```
[... first build tla2tools.jar on this branch ...]

~ $ cd /tmp
/tmp $ mkdir test
/tmp $ cd test
/tmp/test $ jar xf /path/to/tlatools/org.lamport.tlatools/dist/tla2tools.jar

[... create ./model/MC.{tla,cfg} ...]

/tmp/test $ jar --manifest=META-INF/MANIFEST.MF -c -f ../tla2tools.jar .
/tmp/test $ java -jar ../tla2tools.jar

[... normal output ...]
```

Calvin-L:
If there are no further objections, I will merge this later today or over the weekend.

As a reminder, here are the remaining TODOs from comments:
 - Improve clarity, formatting, and performance of `getLibraryPaths` -- I will do this shortly in a follow-up PR
 - Move the new classes to `tla2sany`, provide a well-designed public facade for them, and make `SimpleFilenameToStream` into a simple adapter (see lengthy discussion in comments) -- I have no immediate plans to do this, but it would be nice

lemmy:
This PR seems to have caused some examples to fail: https://github.com/tlaplus/Examples/actions/runs/12839494732/job/35806841891

ahelwer:
Specifically on Windows, which is why it wasn't caught by the PR workflow. Here's the error:
```
Error: Exception in thread "main" java.nio.file.InvalidPathException: Illegal char <:> at index 2: /D:/a/Examples/Examples/specifications/tcp/MCtcp.tla
	at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:232)
	at java.base/java.nio.file.Path.of(Path.java:147)
	at util.SimpleFilenameToStream.locateAndCopyToFilesystemIfNecessary(SimpleFilenameToStream.java:205)
```

Calvin-L:
😅 

I will take a look in a moment. If it looks hard to fix, I will revert.

Issue #1112, opened 2025-01-09
Warning: DiskFPSet.mergeNewEntries: xxx is already on disk with hangup following                                                                                 


arssher:
<!---
Hi!
Have you tried searching for your issue on the following forums?
If you have any questions, please ask them there.

Forums:
 - TLA⁺ Google Groups forum: https://groups.google.com/g/tlaplus/
 - GitHub Discussions forum: (TBA?)

Thanks!
-->

## Description
<!---
Provide a more detailed introduction to the issue itself, and why you consider it to be a bug.

If you need to share a specification, either:
 - Paste it in your description between the <details> </details> tags if it's too long;
 - Send a link to a Gist, GitHub reposity, Pastebin, etc.;
-->

I'm running TLC on a big model. After `states` reaches 142GB which takes about 10 days it appears to consistently fall into weird position; the last logs are
```
Progress(31) at 2025-01-06 16:15:53: 294,462,576,924 states generated (20,387,673 s/min), 6,445,025,382 distinct states found (388,579 ds/min), 1,059,501,205 states left on
 queue.                                                                                                                                                                     
Progress(31) at 2025-01-06 16:16:53: 294,482,750,054 states generated (20,173,130 s/min), 6,445,481,051 distinct states found (455,669 ds/min), 1,059,590,781 states left on
 queue.                                                                                                                                                                     
Warning: DiskFPSet.mergeNewEntries: 9223372034574378188 is already on disk.                                                                                                 
                                                                                                                                                                            
(Use the -nowarning option to disable this warning.)                                                                                                                        
Warning: DiskFPSet.mergeNewEntries: 9223372034574378188 is already on disk.                                                                                                 
                                                                                                                                                                            
Progress(31) at 2025-01-06 16:17:53: 294,484,476,399 states generated (1,726,345 s/min), 6,445,523,635 distinct states found (42,584 ds/min), 1,059,601,836 states left on q
ueue.                                                                                                                                                                       
Progress(31) at 2025-01-06 16:18:53: 294,484,476,399 states generated (0 s/min), 6,445,523,635 distinct states found (0 ds/min), 1,059,601,836 states left on queue.        
Progress(31) at 2025-01-06 16:19:53: 294,484,476,399 states generated (0 s/min), 6,445,523,635 distinct states found (0 ds/min), 1,059,601,836 states left on queue.        
[...]

PR #1122, opened 2025-01-16
Warning: DiskFPSet.mergeNewEntries: xxx is already on disk with hangup following.

lemmy:
Warning: DiskFPSet.mergeNewEntries: xxx is already on disk with hangup following. 

Use double precision when calculating `tblScalingFactor` to avoid rounding issues in `Indexer#getIdx`. These rounding issues can lead to critical errors, such as the largest fingerprints being incorrectly mapped to lower indices instead of the highest index.  This problem occurs intermittently for specific ranges of positions within the interval `(0, Integer.MAX_VALUE)`.

Part of Github issue #1112
https://github.com/tlaplus/tlaplus/issues/1112

[Bug][TLC]

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/fp/OffHeapDiskFPSet.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/fp/OffHeapDiskFPSet.java
@@ -610,7 +610,7 @@ public final class OffHeapDiskFPSet extends NonCheckpointableDiskFPSet implement
 		public Indexer(final long positions, final int fpBits, final long maxFingerprint) {
 			this.positions = positions;
 			// (position-1L) because array is zero indexed.
-			this.tblScalingFactor = (positions - 1L) / ((maxFingerprint - minFingerprint) * 1f);
+			this.tblScalingFactor = (positions - 1L) / ((maxFingerprint - minFingerprint) * 1d);
 		}
 		
 		protected long getIdx(final long fp) {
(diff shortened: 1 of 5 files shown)


tlatools/org.lamport.tlatools/src/tlc2/tool/fp/OffHeapDiskFPSet.java
```diff
@@ -604,13 +604,14 @@ public static class Indexer {
...
 			this(positions, fpBits, 0xFFFFFFFFFFFFFFFFL >>> fpBits);
+			Assert.check(positions < Integer.MAX_VALUE, EC.SYSTEM_FINGERPRINT_OVERFLOW_ERROR);
 			assert fpBits > 0;
 		}
 
 		public Indexer(final long positions, final int fpBits, final long maxFingerprint) {
 			this.positions = positions;
 			// (position-1L) because array is zero indexed.
-			this.tblScalingFactor = (positions - 1L) / ((maxFingerprint - minFingerprint) * 1f);
+			this.tblScalingFactor = (positions - 1L) / ((maxFingerprint - minFingerprint) * 1d);
```
lemmy, line 614: This is the actual fix.

tlatools/org.lamport.tlatools/test/tlc2/tool/fp/LongArraysTest.java
```diff
@@ -622,8 +622,8 @@ public void test7() {
 		final List<Long> expected = new ArrayList<Long>();
 		expected.add(1L);
 		expected.add(0L);
-		expected.add(4L);
 		expected.add(0L);
+		expected.add(4L);
```
lemmy, line 626: Double precision caused the index of element 4 to shift; this is fine.

tlatools/org.lamport.tlatools/src/tlc2/tool/fp/OffHeapDiskFPSet.java
Calvin-L: Not all longs are representable as doubles. Is there any chance that this still isn't enough precision?

Calvin-L (approved):
Seems like a straightforward improvement. 👍 (I left a comment about precision, but I don't think that should block this change.)

PR #1125, opened 2025-01-20
Prototype new programmatic API for SANY

ahelwer:
This is a PR to solicit feedback on the design of a new API for SANY. This API is envisioned to eventually live in the root `src/tla2sany` directory, or in a `src/tla2sany/api` directory. Currently it is only exercised by test code.

To start, read the comments in the `ParserInterface.java` file. Designing these methods took a fair amount of time, trial, and error. The methods that have to be called in a certain order have a nice type-driven design that I am fairly happy with. I heavily favored nested record classes to avoid the scourge of partial initialization. The goal is to provide an interface free of side-effects (such as copying files to temp directories) that can be easily consumed by applications, not just TLC (although TLC can certainly switch to this API if desired).

Feedback I am looking for:
1. Bikeshedding: do the methods and classes have decent names? I think `ParserInterface` and `ParserAPI` definitely need work, and I'm not super happy about `SyntaxClosure` (might just make it a generic `Tree<ModuleSyntaxTree>` datatype).
2. What APIs are missing? I think I have covered all existing use-cases, although emerging use-cases like the desire to add incremental parsing functionality to SANY (see #1096) are not yet covered.
3. The big question: how should error handling be done? I especially want the error reporting design to support unique error codes as outlined in https://github.com/tlaplus/rfcs/issues/15. Should each method throw one of a large number of checked exceptions? You can take a look [here](https://codeberg.org/tlaplus/simple-checker/src/commit/5641793467fafcf86b261be3db6e6d39a21a658f/app/src/main/java/us/tlapl/Parser.java) at a prior prototype to see what the checked exception design looks like.

Feedback I am *not* looking for:
1. "This functionality is already implemented at (location)" - that is true of several things here; the goal is to prototype what the API will look like for the end user before doing the work of fully integrating it with existing functionality, especially for the resolver.


tlatools/org.lamport.tlatools/test/tla2sany/api/SyntaxClosure.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * The syntax tree of the current module.
+   */
+  public final ModuleSyntaxTree Syntax;
+
+  /**
+   * Links to the subtrees of this module's dependencies.
+   */
+  public final List<SyntaxClosure> Dependencies;
```
Calvin-L, line 41: There's a very interesting library design choice to be made here.

I don't have a name for the style you've chosen here, but let's call it "pointer-style": each `SyntaxClosure` object has pointers to other `SyntaxClosure` objects. The dependency graph is formed of heap-allocated objects and pointers.

Although "pointer-style" designs seem quite natural and are extremely efficient, I think they have some sharp edges. For example: object addresses matter. There is a hidden assumption that the heap address of the `SyntaxClosure` object uniquely identifies the corresponding module. That is, `x` and `new SyntaxClosure(x.Syntax, x.Dependencies)` are _not_ the same, even though they contain exactly the same data, which might be surprising.

I don't have a name for the alternative, but let's call it "identifier-style". An identifier-style dependency tree might look like
```java
class DependencyGraph {
    private final ModuleID root;
    private final Map<ModuleID, ModuleSyntaxTree> modules;
    private final Map<ModuleID, Set<ModuleName>> dependencies;
}
```
Each object has an explicit identifier (a `ModuleID`) that is plain-old-data. Modules are always identified by their unique name, not by the address of their `SyntaxClosure`. In fact, everything is plain-old-data, even the `DependencyGraph` itself. Usage is a tad more awkward because of the indirection through `ModuleID`, but there are fewer surprises and less pointer manipulation.

I'm not sure which design is better, but I would lean toward identifier-style for a few reasons:
 1. It's easier to handle diamonds in the graph (er, at least without resorting to exotic things like [`IdentityHashMap`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/IdentityHashMap.html)). In fact, I think `processSemantics` already gets the diamond case wrong; it looks like if a module appears twice in the graph it will be processed twice.
 2. It mirrors existing patterns (`ExternalModuleTable` is an identifier-style graph).
 3. It prevents obscure sharp edges (everything can have value-based-equality, not address-based-equality).
 4. It's impossible to do some operations in a pointer-style graph. For instance: reverse-lookups. To handle those you would need `Set<SyntaxClosure> importers`---but adding such a set creates a bookkeeping nightmare and horrible cyclic heap data structures. Identity-style graphs don't make reverse-lookups efficient, but they at least make them _possible_ with no additional bookkeeping.

tlatools/org.lamport.tlatools/test/tla2sany/api/SyntaxClosure.java
```diff
@@ -0,0 +1,53 @@
...
+ ******************************************************************************/
+package tla2sany.api;
+
+import java.util.List;
+
+/**
+ * A directed graph of module syntax trees, where each module has a link to
+ * the syntax trees of the modules it imports.
+ */
+public class SyntaxClosure {
```
Calvin-L, line 31: Perhaps `DependencyGraph`?

tlatools/org.lamport.tlatools/test/tla2sany/api/ParserInterface.java
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * TODO: figure out this method's ideal signature.
+   */
+  public ExternalModuleTable parse(Path modulePath, Errors log);
+
+  /**
+   * TODO: figure out this method's ideal signature.
+   */
+  public ExternalModuleTable parse(String input, Errors log);
```
Calvin-L, line 68: If possible, at least for the first draft, I would prefer to eliminate all convenience overrides. Let's offer _one_ `parse` method, _one_ `processSyntax` method, etc.

Also, since we already have a clean `IResolver` interface, let's see if we can make this API filesystem-agnostic (i.e. `Path`-free).

I know that's a bit of a strict restriction, especially since this class is meant to cater to a few different use-cases, but I do think it will result in a higher-quality API. There are already some subtle differences between the different overrides that make me nervous: for instance, one of the `processSyntax` overrides includes an additional check for the expected module name. Callers might actually want that check even if their input is coming from a non-`Path` location.

tlatools/org.lamport.tlatools/test/tla2sany/api/ParserInterface.java
```diff
@@ -0,0 +1,134 @@
...
+ * it was made an interface to enable SANY consumers to isolate their code in
+ * testing by injecting their own implementation if desired. Similarly, the
+ * only implementation of {@link IResolver} is {@link DefaultResolver}. Users
+ * can implement their own resolver to customize module resolution behavior
+ * to - for example - draw from their own set of in-memory modules.
+ *
+ * TODO: design an API for incremental parsing - updating a complete parse
+ *       tree to add additional definitions or ghost expressions or modules.
+ */
+public interface ParserInterface {
```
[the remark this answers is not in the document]
ahelwer: I think that's a good name. The motivation for having there be a separation for me is twofold:
1) Discoverability & understandability - the reason I'm making this API when all these things can already be done by just calling the right functions is because those functions are hard to find; it took me a long time to learn the right incantations. Ideally we want to make this interface as clean as possible without any clutter and put it in a prominent place.
2) Testing support - consumers should be able to inject their own implementation of the parser if they want to mock it for whatever reason. I personally appreciate when dependencies expose interface types for this reason.

tlatools/org.lamport.tlatools/test/tla2sany/api/IResolver.java
```diff
@@ -0,0 +1,42 @@
...
+ ******************************************************************************/
+package tla2sany.api;
+
+/**
+ * This important interface is how the SANY API finds the source code of
+ * modules imported by other modules. Currently its only implementation is
+ * {@link DefaultResolver}, but it exists as an interface so API consumers
+ * can override the functionality & inject their own logic during testing.
+ */
+public interface IResolver {
```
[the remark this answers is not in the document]
ahelwer: Hah, my C# past betrays me. Makes sense.

tlatools/org.lamport.tlatools/test/tla2sany/api/ParserInterface.java
ahelwer: The only reason I would push back the simplifying idea is that both of these methods (the `Path`-based and `String`-based) are used by test code, and both are useful. Although it would be acceptable to only use the `String`-based one, thinking about other languages I do tend to like doing all the file handling myself then using whatever in-memory API exists.

Regarding the filename/modulename check, I believe this is part of the language standard so we would want to enforce that in this API somewhere. If they want to enforce it themselves on a `String` then the `ModuleSyntaxTree` class does expose a `getModuleName()` method. Having a `Path` API also identifies a parent directory to search in.

I am interested in making this filesystem agnostic, since I don't actually know how! Is it to use URIs or URLs?

tlatools/org.lamport.tlatools/test/tla2sany/api/SyntaxClosure.java
ahelwer: That's a good name.

tlatools/org.lamport.tlatools/test/tla2sany/api/SyntaxClosure.java
ahelwer: I think those are all good thoughts and you are correct we mishandle diamond dependencies here. Identifier-style seems better.

tlatools/org.lamport.tlatools/test/tla2sany/api/ParserInterface.java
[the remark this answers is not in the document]
Calvin-L: Ok, I think you've sold me on the separation. Also, I think the interface makes it easier to skim the public API without getting distracted by implementation details.

tlatools/org.lamport.tlatools/test/tla2sany/api/ParserInterface.java
Calvin-L: I think the path (haha) to a filesystem-agnostic API might be to take a `String name` and `Resolver resolver` wherever you would take a `Path` argument. (URIs and URLs are another alternative, but they have limitations: for instance, it isn't possible to construct a URI that points at an in-memory string, but it is possible to construct a `Resolver` that wraps it.)

ahelwer: That's a good idea, I quite like that design.

ahelwer: It's a bit clunky but at least the final API is nice. Take a look at `StringResolver` and `RootModuleResolver`.

ahelwer:
After incorporating @Calvin-L's feedback I've put this in its own PR (without the semantic test corpus) since it was getting too big. I think this is a good incremental improvement that is worth merging after whatever additional feedback is given. It is currently only used by test code.

lemmy (approved):

tlatools/org.lamport.tlatools/test/tla2sany/api/DefaultResolver.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.List;
+
+import tla2sany.api.ModuleSourceCode.ModuleOrigin;
+
+/**
+ * This resolver encodes SANY's default behavior for finding modules. This
+ * is currently only a prototype for testing purposes and needs to be
+ * integrated with the {@link util.ResourceLocator} - which is what SANY
+ * currently actually uses - by modifying those (& associated) classes to
+ * return {@link ModuleSourceCode} instances.
```
Calvin-L, line 39: I was about to leave a comment about this, but I see you captured the TODO here. :) To keep things moving I will merge this PR momentarily.

Calvin-L:
...I may have been premature. It seems there is a compile failure against current master. @ahelwer are you able to diagnose and fix it?

PR #1129, opened 2025-01-25
Parameterize operator precedence tests with JUnit

ahelwer:
Nicely separate test cases using JUnit runner logic. Previously this was run in one test in a gigantic for-loop but now we split it into different cases using JUnit 4's parameterized class functionality. This does increase the runtime a bit (now runs on the order of 1-2 seconds) but should be worth it for debugging failures. We can now see how many test cases are generated - around 13,000, and these tests don't include the Unicode operator variants!

Previously there were three separate tests that ran in the same class but due to how JUnit class parameterization works I needed to separate these out into their own classes.

[Refactor][Test][SANY]

### DIFF
--- a/tlatools/org.lamport.tlatools/test/tla2sany/parser/OperatorAssociativityTests.java
+++ b/tlatools/org.lamport.tlatools/test/tla2sany/parser/OperatorAssociativityTests.java
@@ -1,0 +1,95 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Linux Foundation. All rights reserved.
+ *
+ * The MIT License (MIT)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+ * of the Software, and to permit persons to whom the Software is furnished to do
+ * so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN
+ * AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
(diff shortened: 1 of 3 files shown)


tlatools/org.lamport.tlatools/test/tla2sany/parser/OperatorPrecedenceTests.java
```diff
@@ -1,463 +1,460 @@
...
-				UniqueString symbolUnique = UniqueString.uniqueStringOf(symbol);
-				Assert.assertTrue(Operators.existsOperator(symbolUnique));
-				Assert.assertEquals(expected, Operators.resolveSynonym(symbolUnique).toString());
-			}
-		}
-	}
+    /**
+     * Operator symbol variants.
+     */
+    public final String[] Symbols;
```
lemmy, line 68: lowerCamelCase?

tlatools/org.lamport.tlatools/test/tla2sany/parser/OperatorAssociativityTests.java
```diff
@@ -0,0 +1,95 @@
...
+ * correctly; if operators are marked as (left-)associative then parsing
+ * should succeed, but if not it should fail.
+ */
+@RunWith(Parameterized.class)
+public class OperatorAssociativityTests {
+
+  /**
+   * Generates all test cases.
+   */
+  @Parameters(name = "{index}: A {1} B {2} C")
```
lemmy, line 50: Should `{0}` appear here too?

tlatools/org.lamport.tlatools/test/tla2sany/parser/OperatorPrecedenceTests.java
```diff
@@ -1,463 +1,460 @@
...
+      Assert.assertEquals(higherPrecOpSymbol, op1Symbol);
+    }
+  }
+
+  /**
+   * Generates all test cases.
+   */
+  @Parameters(name = "{index}: {1}, {3}")
+  public static List<Object[]> getTests() {
+    final List<Object[]> testCases = new ArrayList<Object[]>();
```
lemmy, line 399: To avoid excessive allocations at startup, consider initializing the ArrayList with a large number of elements, such as 10k (or the number of tests), instead of growing it dynamically.

tlatools/org.lamport.tlatools/test/tla2sany/parser/OperatorPrecedenceTests.java
```diff
@@ -1,463 +1,460 @@
...
+    for (final Operator op1 : OPERATORS) {
+      for (Operator op2 : OPERATORS) {
+        if (FixKind.POSTFIX == op1.Fix && FixKind.PREFIX == op2.Fix) {
+          // Can't construct a test case from this
+          continue;
+        }
+        if (FixKind.PREFIX == op2.Fix && op2.Symbols[0] == "-") {
+          // Skip this case because it's bugged; see
```
lemmy, line 407: `assumeThat` in `testOperatorCombination` should cause JUnit to provide an indicator in its output that the test was skipped.

tlatools/org.lamport.tlatools/test/tla2sany/parser/OperatorAssociativityTests.java
ahelwer: No, since the 1st and 2nd parameters (the actual symbols being used) uniquely identify the test case; the 0th parameter is just info about that operator.

ahelwer:
@lemmy does this look good now?

PR #1133, opened 2025-02-01
Create semantic test corpus for TLA+

ahelwer:
These tests assert various properties about SANY's semantic parse tree. A special Semantics module is developed that includes in-language assertions like `RefersTo` and `IsLevel`. The test code searches the parse tree for these operator calls, analyzes their parameters, and determines whether they are correct at parse time.


tlatools/org.lamport.tlatools/test/tla2sany/semantic/SemanticCorpusTests.java
```diff
@@ -0,0 +1,279 @@
...
+   * The directory containing all the semantic corpus test files.
+   */
+  private static final String CORPUS_DIR = "test/tla2sany/semantic/corpus";
+
+  /**
+   * Finds all semantic corpus test files.
+   */
+  @Parameters(name = "{index}: {0}")
+  public static List<SemanticTestCase> getTestFiles() throws IOException {
+    Path corpusDir = Paths.get(CommonTestCase.BASE_DIR).resolve(CORPUS_DIR);
```
lemmy, line 138: If the tests are executed on the tla2tools.jar file, loading the .tla files with `Files` will fail. Since the .tla files are on the classpath, it’s better to use `classloader.getResourceAsStream(…)` instead.

lemmy (approved):

tlatools/org.lamport.tlatools/test/tla2sany/semantic/SemanticCorpusTests.java
ahelwer: I think this sort of change might have to be delayed until the API prototyped in #1125 is integrated with existing resource loading code.

ahelwer:
@lemmy it would be useful to have this merged so it can be built upon in conjunction with #1140, since while chasing code coverage of semantic errors I naturally see many "close but not quite an error" cases that should be added to this corpus.

tlatools/org.lamport.tlatools/test/tla2sany/semantic/SemanticCorpusTests.java
lemmy: Please add a code-level TODO.

lemmy:
> @lemmy it would be useful to have this merged so it can be built upon in conjunction with #1140, since while chasing code coverage of semantic errors I naturally see many "close but not quite an error" cases that should be added to this corpus.

@Calvin-L What about your review?

PR #1135, opened 2025-02-03
Refactor Errors log class to keep error deserialized internally

ahelwer:
The motivation for these changes are to facilitate development & testing of https://github.com/tlaplus/rfcs/issues/15. In general, it is also nice for users to have programmatic access to error messages (locations and such) instead of them being encoded in a string.

These changes do three things and are broken into a series of logical commits:
1. (Before any changes) Write regression tests against the `Error` class API to ensure these changes do not break anything
2. Refactor the class to use a standard Java `List<>` instead of our custom `Vector` class
3. Change the `Errors` class to store `ErrorDetails` objects instead of string-serialized messages; this class is serialized to a string (in the same format as before) upon request.

The intention here is that this makes it very easy to modify `ErrorDetails` by adding an enumerated error code parameter, so as to incrementally introduce standardized semantic- & level-checking error codes into the codebase.

There was one small bug found by the initial tests: while the `addWarning` and `addError` APIs handled the case of a null `Location` parameter being passed in, the `addAbort` API did not; this was fixed in place.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/Errors.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/Errors.java
@@ -28,31 +28,66 @@ import tla2sany.st.Location;
 
 public class Errors {
 
-  private List<String> warnings = new ArrayList<String>();
-  private List<String> errors   = new ArrayList<String>();
-  private List<String> aborts   = new ArrayList<String>();
+  public static class ErrorDetails {
+
+    public final Location location;
+
+    public final String message;
+
+    public ErrorDetails(Location location, String message) {
+      this.location = location;
+      this.message = message;
+    }
+
+    @Override
+    public String toString() {
+      return this.location.toString() + "\n\n" + this.message;
+    }
+
+    @Override
+    public boolean equals(Object o) {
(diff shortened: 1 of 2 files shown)


Calvin-L (approved):
This seems like a good and straightforward improvement. If @lemmy has no feedback, I will merge it soon. 👍

tlatools/org.lamport.tlatools/src/tla2sany/semantic/Errors.java
```diff
@@ -93,20 +95,14 @@ public final void addError(Location loc, String str) {
...
   public final void addAbort(Location loc, String str, boolean abort) throws AbortException {
-    String errMsg = loc.toString() + "\n\n" + str;
-    int i;
-    for (i = aborts.size()-1; i >= 0; i--) {
-      if (errMsg.equals(aborts.elementAt(i))) break;
-    }
-    if (i < 0) {
-      aborts.addElement(errMsg);
-      numAborts++;
+    if (loc == null) loc = Location.nullLoc;
```
lemmy, line 98: Let’s please never use if/while/... statements without curly braces.

lemmy (approved):

PR #1137, opened 2025-02-04
Warning: DiskFPSet.mergeNewEntries: xxx is already on disk with hangup following.

lemmy:
https://github.com/tlaplus/tlaplus/issues/1112

(supersedes https://github.com/tlaplus/tlaplus/pull/1126)


tlatools/org.lamport.tlatools/src/tlc2/tool/fp/OffHeapDiskFPSet.java
```diff
@@ -594,44 +599,194 @@ public int compare(Future<Long> o1, Future<Long> o2) {
...
+	        final long lh = fpL * mHigh;
+	        final long hl = fpH * mLow;
+	        final long hh = fpH * mHigh;
+
+	        // Combine four partial products including carry overs.
+	        long l = lh + (ll >>> 32);
+	        long h = hl + (l & MASK);
+	        l = hh + (l >>> 32);
+	        l = l + (h >>> 32);
+	        h = (h << 32) | (ll & MASK);
```
Calvin-L, line 764: I'm not entirely sure I follow the math here---it looks like `h` will hold the least-significant digits (random testing suggests `h == fp * multiplier` might be a valid postcondition here).

Should `l` then be the most-significant bits of the product? That can be easily computed with `Math.multiplyHigh(fp, multiplier)`, but that doesn't seem to be what `l` comes out to.

Calvin-L: UPDATE: Actually `l == Math.multiplyHigh(fp, multiplier)` when `fp >= 0`, which I think it always will be.

Perhaps this code can be simplified? `Math.multiplyHigh` might be implemented more efficiently by the JVM than we can manage ourselves.

Calvin-L (approved):

PR #1140, opened 2025-02-13
Create semantic error test corpus for TLA+

ahelwer:
These changes are sort of the complement of #1133, where instead of writing assertions about correct TLA+ parser behavior we develop a set of specifications that should trigger specific errors & be rejected. Don't be intimidated by the large number of files changed! Almost all of them are just tiny .tla files that trigger specific errors.

The development of this corpus will pretty much be guided by code coverage of the semantic checking code. This will give us a good base for standardization of error codes as proposed in https://github.com/tlaplus/rfcs/issues/15.

This will be a draft PR for a while but feedback on the initial approach is welcome; while most of this is test code, it will involve relatively minor changes to SANY itself: each call to `errors.addError` et. al will be annotated with a specific enumerated error code. I try to match the existing function parameter indentation. Occasionally I add a comment about the error, especially if the error is suspected to be unreachable. These codes will not be exposed in user-facing error messages until this approach has stabilized.

List of functional changes to non-test code beyond adding error code annotations:
- In `SymbolTable.java`, moved the `REDEFINED_BUILT_IN_SYMBOL` check above the `MULTIPLY_DEFINED_SYMBOL` check; the latter encompasses the former due to the `currentBinding.getKind() != symbol.getKind()` condition (`BuiltInKind` is a possible kind), so it was never possible to trigger the `REDEFINED_BUILT_IN_SYMBOL` error path and a less-descriptive error was being output when the user tried to redefine a built-in operator.
- In `SubstInNode.java`, removed an unnecessary `@SuppressWarnings("unused")` annotation on the `addExplicitSubstitute` method (these tests confirm the method is definitely used!)
- In `SubstInNode.java`, the Java compiler found that one of the error reports was unreachable. This is because a variable was first checked with `instanceof` (function terminated if evaluated to false), then checked to not be null, and the error report only triggered if the variable was null; since `instanceof` returns false when given null, the error report was unreachable. This code block was simplified to remove the unreachable error report.
- Removed some APIs in the `Errors` class that are now no longer used by any caller.


ahelwer:
@lemmy @Calvin-L this is now finally complete & ready for review! Fear not the gigantic file change count, almost all of them are tiny .tla files triggering parse errors and most of the changes to actual code is just tagging `addError` calls with a specific error code enum parameter.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/ErrorCode.java
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * The standardized error code value.
+   */
+  public final int value;
+
+  /**
+   * The error's level of seriousness.
+   */
+  public final ErrorLevel level;
```
Calvin-L, line 167: Consider hiding fields behind getters.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/Errors.java
```diff
@@ -30,11 +30,14 @@ public class Errors {
 
   public static class ErrorDetails {
 
+    public final ErrorCode code;
```
Calvin-L, line 33: Consider hiding this field behind a getter.

Calvin-L (approved):
I have not looked at every test case, but the source and test code changes look good to me. I'm very happy for the increased test coverage. :)

My one quibble would be to hide the new public fields behind getters. Unlike C#, Java doesn't offer a way to change how `x.f` works without refactoring every usage---but we can change how `x.f()` works by just changing the implementation of `f()`.

(That is an exceedingly minor complaint though, and I'll leave it to your judgement. The practice is only really necessary for projects that need to offer a stable ABI. I concede that Java developers are needlessly afraid of `public final` fields, even though they are generally very safe.)

tlatools/org.lamport.tlatools/src/tla2sany/semantic/ErrorCode.java
```diff
@@ -0,0 +1,196 @@
...
+  /**
+   * Standardized warnings. These should not cause a parse failure.
+   */
+  EXTENDED_MODULES_SYMBOL_UNIFICATION_AMBIGUITY (1000, ErrorLevel.WARNING),
+  INSTANCED_MODULES_SYMBOL_UNIFICATION_AMBIGUITY (1001, ErrorLevel.WARNING),
+
+  /**
+   * Standardized errors. These should cause a parse failure.
+   */
+  EXTENDED_MODULES_SYMBOL_UNIFICATION_CONFLICT (2000, ErrorLevel.ERROR),
```
lemmy, line 93: Should these error codes be disjoint from the error codes in [tlc2.output.EC.java](https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/src/tlc2/output/EC.java)?

tlatools/org.lamport.tlatools/test/tla2sany/semantic/error_corpus/E2000_Extended.tla
```diff
@@ -0,0 +1,6 @@
+Used by E2000_Test as an EXTENDS target; the + operator type is a CONSTANT,
```
lemmy, line 1: These tests appear to be an excellent food for LLMs.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/ErrorCode.java
ahelwer: Ah good call, yes they should. I will change their values to an appropriate unused X000 block.

ahelwer: Sorted & renamed using this bash script:
```sh
ERROR_CODE_PATH="/home/ahelwer/src/tlaplus/java-tools/tlatools/org.lamport.tlatools/src/tla2sany/semantic/ErrorCode.java"
sed -i -e "s/$1/$2/g" "$ERROR_CODE_PATH"
git add "$ERROR_CODE_PATH"
for file in *$1*.tla ; do
  sed -i -e "s/$1/$2/g" "$file" ;
  git add "$file" ;
  git mv $file ${file//$1/$2} ;
done
```

ahelwer:
@lemmy this one seem good to you?

lemmy (approved):

PR #1144, opened 2025-02-18
Fill mandatory reason field to comply with the protocol (fixes #1143)

### LINKED ISSUE #1143: Debugger doesn't work with LSP4IJ
## Description
TLC debugger doesn't work with some DAP clients (e.g. [LSP4IJ](https://github.com/redhat-developer/lsp4ij)) due to the DAP protocol violation in TLC debugger regarding `StoppedEvent`.

## Steps to Reproduce
1. Install Intellij IDEA 2023.2 community edition
2. Install [LSP4IJ plugin](https://plugins.jetbrains.com/plugin/23257-lsp4ij) on intellij
3. Download tla2tools.jar (SHA1 sum: `57ececee2c0f9915bd4c47fab99525c475b8cd87`)
4. Prepare `test.tla` and `test.cfg` in `/path/to/work-dir`
    * ```
      ---- MODULE test ----
      VARIABLES x
      
      Init == x = TRUE
      
      Next == x' = ~x
      
      Spec == Init /\ [][Next]_x
      ====
       ```
    * ```
       INIT Init
       NEXT Next
      ```
6. Create `Debug Adapter Protocol` Debug configuration like below:
    * Image
    * Image
7. Set breakpoint at `Next`
    * Image
8. Debug the spec by running `Debug 'TLC Debugger'`

## Expected Behavior
The debugger launches and stops at the breakpoint (Precisely, it should stop at `Init` even without breakpoint because TLC debugger sends `StoppedEvent` at the beginning though)

## Actual Behavior
The debugger doesn't stop at the expected point. Also, TLC hangs with waiting commands from DAP client.

## Steps Taken to Fix
N/A as the cause is already found.

### Summary
The cause is the missing `reason` field in [`StoppedEvent`](https://microsoft.github.io/debug-adapter-protocol/specification#arrow_left-stopped-event) event sent from TLC Debugger server, which is mandatory by protocol specification.

### Details
- LSP4IJ does incoming event validation using LSP4J's [`ReflectiveMessageValidator`](https://github.com/redhat-developer/lsp4ij/blob/0.10.0/src/main/java/com/redhat/devtools/lsp4ij/dap/client/DAPClient.java#L130)
- `StoppedEvent` from TLC Debugger server fails to the validation due to the missing `reason` field, which causes the event to be discarded, resulting in failing to stop as reported.

### Suggested fix
- Fill StoppedEvent's `reason` fields in TLC Debugger to make it protocol-compliant

## Your Environment
 - TLC version: v1.8.0 (https://github.com/tlaplus/tlaplus/commit/5e1673e4b0baf64e6fbed6106fbbcf29d812e514)
 - Operating System: macOS Sonoma

ocadaruma:
As described in #1143, `StoppedEvent`'s `reason` field is now missing while it is mandatory, which causes some clients to fail to work with TLC debugger.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/debug/AttachingDebugger.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/debug/AttachingDebugger.java
@@ -40,7 +40,6 @@ import java.util.concurrent.Executors;
 
 import org.eclipse.lsp4j.debug.OutputEventArguments;
 import org.eclipse.lsp4j.debug.StoppedEventArguments;
-import org.eclipse.lsp4j.debug.StoppedEventArgumentsReason;
 import org.eclipse.lsp4j.debug.launch.DSPLauncher;
 
 import tlc2.tool.impl.Tool;
@@ -149,7 +148,10 @@ public class AttachingDebugger extends TLCDebugger {
 			
 			StoppedEventArguments eventArguments = new StoppedEventArguments();
 			eventArguments.setThreadId(0);
-			eventArguments.setReason(StoppedEventArgumentsReason.PAUSE);
+			// DAP mandates non-null reason in StoppedEvent by spec.
+			// Since StoppedEvent created here is to stop the execution voluntarily on launch which
+			// pre-defined reasons doesn't fit, we fill empty-string as default.
+			eventArguments.setReason("");
 			launcher.getRemoteProxy().stopped(eventArguments);
 		});
 		
--- a/tlatools/org.lamport.tlatools/src/tlc2/debug/AttachingDebugger.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/debug/AttachingDebugger.java
@@ -40,6 +40,7 @@ import java.util.concurrent.Executors;
 
 import org.eclipse.lsp4j.debug.OutputEventArguments;
 import org.eclipse.lsp4j.debug.StoppedEventArguments;
+import org.eclipse.lsp4j.debug.StoppedEventArgumentsReason;
 import org.eclipse.lsp4j.debug.launch.DSPLauncher;
 
 import tlc2.tool.impl.Tool;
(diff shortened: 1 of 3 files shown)


tlatools/org.lamport.tlatools/src/tlc2/debug/TLCStackFrame.java
```diff
@@ -669,6 +669,8 @@ public StoppedEventArguments getStoppedEventArgument() {
 			// the term "exception" appears in the front-end.
 			eventArguments.setReason(StoppedEventArgumentsReason.EXCEPTION);
 			eventArguments.setText(this.exception.getMessage().replaceAll("(?m)^@!@!@.*", ""));
+		} else {
+			eventArguments.setReason(StoppedEventArgumentsReason.STEP);
```
lemmy, line 673: Why `STEP` and not `PAUSE`, or `BREAKPOINT`, ...?  Does the [spec](https://microsoft.github.io/debug-adapter-protocol/specification) mandate `STEP` as a default?

ocadaruma: That was just because I found reason-missing SEA here when testing step-over, without decent intention.
The spec doesn't mention about default value so using empty string sounds more appropriate. Let me fix.

tlatools/org.lamport.tlatools/src/tlc2/debug/AttachingDebugger.java
```diff
@@ -148,6 +149,7 @@ public CompletableFuture<Void> launch(Map<String, Object> args) {
 			
 			StoppedEventArguments eventArguments = new StoppedEventArguments();
 			eventArguments.setThreadId(0);
+			eventArguments.setReason(StoppedEventArgumentsReason.PAUSE);
```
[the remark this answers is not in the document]
ocadaruma: Why I chose `PAUSE` was because it sounds like natural for me here, where the debugger stops voluntary on launching.
Due to the same reason as https://github.com/tlaplus/tlaplus/pull/1144#discussion_r1959641855, using empty string sounds more appropriate.

lemmy (approved):

lemmy:
Thanks for this PR!  It’s great to see that the IntelliJ TLA+ plugin will also support the TLA+ Debugger. :-)

PR #1146, opened 2025-02-18
Handle excepts in FcnLambdaValue.toTuple()

### LINKED ISSUE #1145: `FcnLambdaValue.toTuple()` ignores `this.excepts`
## Description
Originally reported on the mailing list: https://groups.google.com/d/msgid/tlaplus/5da8f6e5-04fd-4488-9e43-eeeb35a2cc7cn%40googlegroups.com

I minimized the example input and did a little debugging to diagnose the issue.

```tla
---- MODULE BadExcept ----

EXTENDS Naturals, Sequences

F(seq) == SubSeq(seq, 2, 2)

ASSUME F([[i \in 1..4 |-> 0] EXCEPT ![2] = 1])[1]
     = F(<<0, 1, 0, 0>>                      )[1]

====
```

## Steps to Reproduce
1. Create BadExcept.tla as above
2. Create an empty BadExcept.cfg
3. `tlc2 BadExcept.tla`

## Expected Behavior
The assumption is true because
```tla
[[i \in 1..4 |-> 0] EXCEPT ![2] = 1]
```
is equal to
```
<<0, 1, 0, 0>>
```

## Actual Behavior
```
Error: Assumption line 7, col 8 to line 8, col 49 of module BadExcept is false.
```

## Possible Fix
It is possible to work around this problem with careful use of `TLCEval` (requires `EXTENDS TLC`). For instance, the assumption passes if `F` is instead defined as
```
F(seq) == SubSeq(TLCEval(seq), 2, 2)
```

## Notes

The root cause seems to be that [`FcnLambdaValue.toTuple()`](https://github.com/tlaplus/tlaplus/blob/6f1eb447ca79e8a67ec46fa2c986704bdae381e1/tlatools/org.lamport.tlatools/src/tlc2/value/impl/FcnLambdaValue.java#L588) ignores `this.excepts`. So:
 - `SubSeq` is called with a `FcnLambdaValue` as its first argument
 - [It calls `toTuple()`](https://github.com/tlaplus/tlaplus/blob/6f1eb447ca79e8a67ec46fa2c986704bdae381e1/tlatools/org.lamport.tlatools/src/tlc2/module/Sequences.java#L273) on its first argument
 - The `EXCEPT ![2] = 1` gets ignored

Note that the `[1]` in the assumption is necessary for reproducing the issue because it prevents TLC from eagerly evaluating the function constructor (i.e. it will stay as a `FcnLambdaValue` because `EvalControl.KeepLazy` is set).

The original mailing list post reports that addition of `PrintT` fixes the problem. This appears to be because `FcnLambdaValue.toString` _does_ account for excepts. I suspect that the printed value was not actually the one being used by TLC, and that nothing was actually fixed by `PrintT`. It only helped reveal the discrepancy.

## Your Environment
MacOS, OpenJDK 11, TLC 4e398d4d4c1fda53988b1689a20027f3da6dcf6d

Calvin-L:
Fixes #1145.

The `toTuple()` method does not take `this.excepts` into account.  As a result, TLC might mis-evaluate certain expressions involving functions.

To fix the problem, this commit replaces manual calls to `tool.eval` with calls to `select(Value)`, which correctly handles excepts.

### DIFF
--- a/tlatools/org.lamport.tlatools/test-model/Github1145.tla
+++ b/tlatools/org.lamport.tlatools/test-model/Github1145.tla
@@ -1,0 +1,14 @@
+\* Copyright (c) 2025, Oracle and/or its affiliates.
+---- MODULE Github1145 ----
+
+EXTENDS Naturals, Sequences
+
+F(seq) == SubSeq(seq, 2, 2)
+
+ASSUME F([[i \in 1..4 |-> 0] EXCEPT ![2] = 1])[1]
+     = F(<<0, 1, 0, 0>>                      )[1]
+
+ASSUME F([[i \in {1, 2, 3, 4} |-> 0] EXCEPT ![2] = 1])[1]
+     = F(<<0, 1, 0, 0>>                              )[1]
+
+====
--- a/tlatools/org.lamport.tlatools/src/tlc2/value/impl/FcnLambdaValue.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/value/impl/FcnLambdaValue.java
@@ -1,5 +1,6 @@
 // Copyright (c) 2003 Compaq Corporation.  All rights reserved.
 // Portions Copyright (c) 2003 Microsoft Corporation.  All rights reserved.
+// Copyright (c) 2025, Oracle and/or its affiliates.
 // Last modified on Wed 12 Jul 2017 at 16:10:00 PST by ian morris nieves
 //      modified on Wed  4 Jul 2007 at 17:31:23 PST by lamport
 //      modified on Thu Dec  6 21:46:34 PST 2001 by yuanyu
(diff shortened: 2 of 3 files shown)


tlatools/org.lamport.tlatools/test-model/Github1145.tla
```diff
@@ -0,0 +1,14 @@
+\* Copyright (c) 2025, Oracle and/or its affiliates.
+---- MODULE Github1145 ----
+
+EXTENDS Naturals, Sequences
+
+F(seq) == SubSeq(seq, 2, 2)
```
ahelwer, line 6: (for my own understanding) why is this operator necessary? Is it that the functions will not otherwise be coerced to a sequence?

ahelwer (approved):

tlatools/org.lamport.tlatools/test-model/Github1145.tla
Calvin-L: It just helps reduce duplication. I believe the test will still fail in current master if you inline `F`.

lemmy (approved):

PR #1148, opened 2025-02-19
LazyValue no longer inherits from Value

Calvin-L:
LazyValue supports none of the functionality of Value:
 - it can't be compared for equality
 - it can't be tested for membership
 - it can't be serialized
 - it can't be converted to other Value types
 - it can't be passed to Java operator overrides

It seems tempting to treat LazyValue as a not-yet-evaluated value, but that doesn't work out in practice.  The meaning of a LazyValue depends on the current behavior and whether it is being used under a prime operator or inside `ENABLED`.  TLC does not (and has never) allowed LazyValue inside another LazyValue or inside other composite values like SetEnumValue.

This change should make us a little more confident in TLC's existing (and future) code: there is greatly reduced risk of LazyValue "leaking" into places it doesn't belong.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/value/impl/LazyValue.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/value/impl/LazyValue.java
@@ -1,30 +1,25 @@
 // Copyright (c) 2003 Compaq Corporation.  All rights reserved.
 // Portions Copyright (c) 2003 Microsoft Corporation.  All rights reserved.
-// Copyright (c) 2023, Oracle and/or its affiliates.
+// Copyright (c) 2023, 2025, Oracle and/or its affiliates.
 // Last modified on Wed 12 Jul 2017 at 16:10:00 PST by ian morris nieves
 //      modified on Mon 30 Apr 2007 at 15:30:08 PST by lamport
 //      modified on Thu Feb  8 21:23:55 PST 2001 by yuanyu
 
 package tlc2.value.impl;
 
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
-
 import tla2sany.semantic.SemanticNode;
 import tlc2.tool.EvalControl;
-import tlc2.tool.FingerprintException;
 import tlc2.tool.TLCState;
 import tlc2.tool.coverage.CostModel;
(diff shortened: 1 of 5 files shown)


tlatools/org.lamport.tlatools/src/tlc2/value/impl/LazyValue.java
```diff
@@ -121,6 +117,10 @@ public LazyValue(SemanticNode expr, Context con, final boolean cachable, final C
     }
   }
 
+  public CostModel getCostModel() {
```
ahelwer, line 120: Do you generally think it is necessary to define getter methods for public final fields?

ahelwer (approved):
Approved as long as you've searched the codebase for any sub-casts to LazyValue

tlatools/org.lamport.tlatools/src/tlc2/value/impl/LazyValue.java
Calvin-L: Definitely not. Good style dictates this should be a getter; performance dictates this should be a public field; we have both because `Value` has both and I don't want to fix all the usages just yet.

Calvin-L: I've added a TODO for this.

Calvin-L:
> any sub-casts to LazyValue

...this leads to a thorny topic, which is why I didn't get to it right away (and then forgot about this PR).

With this change, casts to LazyValue are guaranteed to be safe (in the sense that if they would throw an exception now, they would have thrown an exception before, too). This is because at run-time, every object that was a LazyValue before this change is still a LazyValue after this change. Nice!

But: this PR does introduce a new related risk. Casts from `Object` to `IValue` or to `Value` might now fail at run-time where they would have previously succeeded. If at run-time `Object x` holds a `LazyValue`, then `(Value)x` will now throw `ClassCastException`.

Answering the second concern turns out to be very difficult because TLC makes many casts from `Object` to `Value` and `IValue`. I have audited as many of those casts as I can. All the ones I looked at are safe---but that's no proof. :( As evidence of safety, TLC's tests do pass with this change. :) The tests don't achieve 100% coverage, but they exercise lots of LazyValue weirdness because it's been such a pain point in the past.

Issue #1170, opened 2025-04-18
Support company-level execution statistics


lemmy:
TLC has been collecting and reporting anonymized, PII-free [execution statistics](https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/src/util/ExecutionStatisticsCollector.md) for many years. This data is stored and [publicly shared](https://exec-stats.tlapl.us) by INRIA. These statistics help us understand the hardware and software environments of our users, and provide a rough estimate of our user base size (see [analytics](http://metabase.tlapl.us/public/dashboard/cf7e1a79-19b6-4be1-88bf-0a3fd5aa0dec)).

Importantly, TLC’s exec statistics collection is **opt-in**, whereas many open-source tools (e.g., VSCode) default to collecting execution statistics unless users explicitly opt out. To be clear, **this proposal is not about changing the opt-in model**—such a decision belongs to the TLA+ Foundation board.

That said, larger companies exploring or adopting TLA+ want TLA+ usage trends. Unfortunately, our current exec statistics don’t support this use case:

 **a)** Anonymized statistics prevent identifying executions tied to a particular organization.  
 **b)** There is currently no mechanism to enable opt-in exec statistics across an entire organization.

In theory, companies could instruct employees to [manually opt-in](https://github.com/tlaplus/vscode-tlaplus/issues/114) and apply a common prefix to their installation IDs, but this is difficult to coordinate at scale and often impractical. Furthermore, some companies may be unwilling or unable to share exec statistics **publicly**.

To support this enterprise use case, **TLC should allow:**
1. Companies to automatically opt-in all installations.
2. Execution statistics to be reported to a **company-controlled endpoint**.

Currently, when exec stats are enabled, TLC opens an HTTP connection to `esc01.tlapl.us` at startup. We propose enhancing this behavior as follows:

- TLC should attempt to resolve a DNS record named `tlaplus-esc01` (value can be overridden with `-Dutil.ExecutionStatisticsCollector.domain`). 
- If the DNS query results in `NXDOMAIN`, TLC proceeds with the current default behavior (reporting to `esc01.tlapl.us`).  
- If the DNS record successfully resolves (e.g., to `tlaplus-esc01.mycompany.com` [compare DNS search domain](https://en.wikipedia.org/wiki/Search_domain)), TLC will assume opt-in is enabled and instead report statistics to that endpoint (`tlaplus-esc01.mycompany.com`).

[...]

PR #1172, opened 2025-04-21
Support company-level execution statistics

lemmy:
https://github.com/tlaplus/tlaplus/issues/1170


ahelwer:
Is it possible to distinguish between the public DNS entry resolving and a private DNS entry resolving?

I think it would be nice to make usage statistics opt-in even if the public DNS entry resolves, if that is technically feasible. That way a code change would be necessary to change usage statistics to opt-out instead of just a DNS change.

lemmy:
> Is it possible to distinguish between the public DNS entry resolving and a private DNS entry resolving?
> 
> I think it would be nice to make usage statistics opt-in even if the public DNS entry resolves, if that is technically feasible. That way a code change would be necessary to change usage statistics to opt-out instead of just a DNS change.

To clarify, DNS records themselves are all public. The distinction comes from how domain names are resolved within a specific environment. For example, the company "Frob" configures its machines with `frob.com` as the DNS search domain. If Frob creates the record `tlaplus-execution-stats-collection01.frob.com`, then any corporate-owned machine resolving `tlaplus-execution-stats-collection01` will automatically resolve it to `tlaplus-execution-stats-collection01.frob.com` due to the search domain.  This implicit resolution acts as an opt-in mechanism: simply resolving that name enables execution statistics to be sent to `tlaplus-execution-stats-collection01.frob.com` instead of the default `esc01.tlapl.us`.

lemmy:
## Security Threat Analysis

Let `tesc` be `tlaplus-execution-stats-collection01`.

### DNS Hijacking & Cache Poisoning Risks
A machine owned by the organization "Frob" may be tricked into reporting telemetry data to a malicious endpoint—such as `tesc.iamevil.com`—instead of the legitimate `tesc.frob.com`. This could happen if an attacker manages to:
* Poison the machine's DNS cache, or
* Hijack the machine’s or the network’s DNS resolver.

### Cross-Network Telemetry Collection
Separately, there’s a scenario where TLC  might report to an unintended but non-malicious domain such as `tesc.bar.net`. This could occur if: A Frob employee connects their device to a network operated by another organization, say "Bar", and Bar operates its own telemetry collection endpoint at `tesc.bar.net`, and the machine resolves `tesc` according to Bar’s DNS. This is not necessarily an attack, but a consequence of the environment in which the machine is operating.

### Protections and Limitations
DNSSEC can protect against DNS poisoning and hijacking by ensuring DNS responses are cryptographically signed. However, this is beyond the control or scope of TLC itself. Design Consideration: TLC is designed to be easy to deploy in Frob's environment, requiring only the creation of a DNS record and an HTTPS endpoint. `tesc` has been chosen to make collisions unlikely.

Potential but Unused Mitigation: TLC could have been designed to allow machines to opt in to a specific `tesc` domain only once and then "lock" that domain, ensuring all future telemetry reports go exclusively to, for example, `tesc.frob.com`. However, this approach was not implemented.

### Role of TLS Certificates
What does the SSL/TLS certificate at `tlaplus-execution-stats-collection01.frob.com` protect?

TLC resolves `tesc` to `tesc.frob.com` at the DNS level before attempting an HTTPS connection. It will then connect specifically to `tesc.frob.com`, which must present a valid TLS certificate for that hostname.

Importantly, if someone tries to resolve and connect to just `tesc` (e.g., https://tesc), the TLS handshake will fail unless the certificate includes a wildcard like `tesc.*` in its Subject Alternative Name. However, no legitimate certificate authority will issue such a wildcard, thereby preventing the misuse of ambiguous or misleading DNS names.

tlatools/org.lamport.tlatools/src/util/ExecutionStatisticsCollector.java
```diff
@@ -60,75 +64,131 @@ public String toString() {
...
+		// file containing the NO_ESC_STR.
+		final File udcFile = new File(pathname);
+		if (udcFile.exists()) {
+			try (BufferedReader br = new BufferedReader(new FileReader(udcFile))) {
+				line = br.readLine();
+				if (NO_ESC_STR.equals(line.trim())) {
+					return;
+				}
+			} catch (Exception swallow) {
+			}
```
Calvin-L, line 129: If the file is unreadable for some unknown reason, it might be better to return or throw instead of silently proceeding. (Only a successful read of the file can prove that it _doesn't_ contain `NO_ESC_STR`.)

Maybe something like
```java
try (...) {
   ...
} catch (FileNotFoundException | NoSuchFileException swallow) {
    // file does not exist; continue because the user has not expressed a preference
} catch (IOException e) {
    throw new RuntimeException("Unable to read " + udcFile, e);
}
```

lemmy: The idea is to not bother the user with exec stats even if something is amiss.  For example, we would have to think about it a (TLC)RuntimeException interferes with the REPL, model-checking, or simulation. I’m open to changing the default to disable execution statistics in both cases if the file is unreadable.

lemmy: https://github.com/tlaplus/tlaplus/pull/1172/commits/a410fdf9c35743e28a63bf48ba1874d0dad58fce

PR #1173, opened 2025-04-24
Raise a warning if a state-level formula is used as a PROPERTY.

lemmy:
`Warning: The formula Inv is a state-level formula, but it was used as a PROPERTY (or PROPERTIES), where a temporal formula is typically expected. State-level formulas used as PROPERTY or PROPERTIES are only checked in the initial state. To verify that the formula Inv holds in all states of every behavior, use INVARIANT Inv instead. Alternatively, applying the "always" temporal operator (□) to the state-level formula Inv changes it into a temporal formula, asserting that Inv holds in all states of every behavior.`

[Feature][TLC]

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/output/EC.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/output/EC.java
@@ -257,6 +257,7 @@ public interface EC
     public static final int TLC_LIVE_CANNOT_EVAL_FORMULA = 2251;
     public static final int TLC_LIVE_ENCOUNTERED_NONBOOL_PREDICATE = 2252;
     public static final int TLC_LIVE_FORMULA_TAUTOLOGY = 2253;
+    public static final int TLC_LIVE_FORMULA_STATE_LEVEL = 2255;
 
     
     public static final int TLC_EXPECTED_VALUE = 2215;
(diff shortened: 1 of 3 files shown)


lemmy:
We need a better way—both conceptually and technically—to provide extended help like the example below. Trying to squeeze this level of detail into a standard warning message, like the one shown above, isn’t practical. At the same time, posting it somewhere online, such as on https://docs.tlapl.us/ or GitHub, makes it difficult to discover when users actually need it. One idea is for TLC to support an “extended warning/error” mode, where it would provide detailed explanations for each issue. Thoughts?

# When to use INVARIANT/INVARIANTS and when to use PROPERTY/PROPERTIES with TLC?

The `INVARIANT` (or `INVARIANTS`) keyword in a TLC configuration file is used to specify a condition that must hold in all reachable states of the system. This is typically how safety properties are expressed—asserting that "something bad never happens." For instance, an invariant might state that a certain variable is always non-negative.

However, the same condition can also be written as a temporal property using TLA+ syntax. In this case, assumming an invariant `I`, it would be written as `[]I` (read as "always I") in the TLA+ specification. Then, in the config file, instead of using `INVARIANT I`, you would use `PROPERTY I`.

Important note: If you switch from `INVARIANT I` to `PROPERTY I` without updating the definition of `I` to include the temporal `[]` operator, the meaning changes drastically. Without the `[]`, TLC will only check that `I` holds in the initial states—not in all reachable states—potentially missing violations of the intended property.

Calvin-L:
In https://github.com/tlaplus/rfcs/issues/15 there's a related suggestion:
> documentation websites can have an error code section with expanded explanations 

Although I definitely like the idea of an offline “extended warning/error” mode. Something like
```
tlc --verbose-errors [...normal args...]
```
Or if you don't want to re-run the model checker:
```
tlc --explain [errorcode]
```

lemmy:
Git commit that is logically part of this PR: 21974dcc0bf06649cc00a76a724b9bee8d7da8eb

PR #1180, opened 2025-04-30
When an invariant is violated, show the values of any \A-bound names

Calvin-L:
This commit makes TLC output some additional helpful information for invariants of the form

    \A x \in S: P(x)

Specifically, TLC will now point to the souce location of P and give the value of x that makes P false.

For instance, given a simple spec like

    ---- MODULE InvariantViolationExplanation ----
    EXTENDS Naturals
    VARIABLE i
    Init == i = 1
    Next == i < 10 /\ i' = i + 1
    Safe == \A x \in 0..i: x+x /= i
    ====

TLC will now output

    Error: Invariant Safe is violated.
    More information: The expression at line 6, col 24 to line 6, col 31 of module InvariantViolationExplanation is FALSE (when x = 1)

(In addition to the behavior leading to the violation, which contains the value of `i`.)

Having the specific value of x can make the violation much easier to digest.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
@@ -7,6 +7,7 @@ import java.io.StringWriter;
 import java.io.Writer;
 import java.text.DecimalFormat;
 import java.text.SimpleDateFormat;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.HashSet;
 
@@ -498,6 +499,9 @@ public class MP
             break;
         case EC.TLC_INVARIANT_VIOLATED_INITIAL:
             b.append("Invariant %1% is violated by the initial state:\n%2%");
+            if (parameters.length > 2) {
+                b.append("\n%3%");
+            }
             break;
         case EC.TLC_PROPERTY_VIOLATED_INITIAL:
             b.append("Property %1% is violated by the initial state:\n%2%");
@@ -519,6 +523,9 @@ public class MP
             break;
         case EC.TLC_INVARIANT_VIOLATED_BEHAVIOR:
             b.append("Invariant %1% is violated.");
+            if (parameters.length > 1) {
+                b.append("\n%2%");
+            }
             break;
         case EC.TLC_INVARIANT_VIOLATED_LEVEL:
         	b.append("The invariant %1% is not a state predicate (one with no primes or temporal operators).");
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/ITool.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/ITool.java
@@ -1,5 +1,6 @@
 /*******************************************************************************
  * Copyright (c) 2019 Microsoft Research. All rights reserved. 
+ * Copyright (c) 2025, Oracle and/or its affiliates.
  *
  * The MIT License (MIT)
  * 
@@ -126,6 +127,17 @@ public interface ITool extends TraceApp, OpDefEvaluator {
 	TLCState enabled(SemanticNode pred, Context c, TLCState s0, TLCState s1);
 	TLCState enabled(SemanticNode pred, IActionItemList acts, Context c, TLCState s0, TLCState s1);
 	TLCState enabled(SemanticNode pred, IActionItemList acts, Context c, TLCState s0, TLCState s1, CostModel cm);
+
+	/**
+	 * Evaluate the given action, and if it evaluates to FALSE, return information about why.
+	 *
+	 * @param act the action
+	 * @param s0 the current state
+	 * @param s1 the next state (or {@link TLCState#Empty})
+	 * @return information about why the action evaluates to FALSE, or {@code null} if the action evaluates to TRUE
+	 * @throws util.Assert.TLCRuntimeException if the action does not evaluate to a boolean value
+	 */
+	FalseExpressionWithDetails checkValidity(Action act, TLCState s0, TLCState s1);
 
 	boolean isValid(ExprNode expr, Context ctxt);
 	
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/ModelChecker.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/ModelChecker.java
@@ -559,15 +559,15 @@ public class ModelChecker extends AbstractChecker
         return false;
 	}
 
-	final boolean doNextSetErr(TLCState curState, TLCState succState, boolean keep, int ec, String param) throws IOException, WorkerException {
+	final boolean doNextSetErr(TLCState curState, TLCState succState, boolean keep, int ec, String... params) throws IOException, WorkerException {
 		synchronized (this)
 		{
 		    if (this.setErrState(curState, succState, keep, ec))
 		    {
-		    	if (param == null) {
+		    	if (params == null) {
 		    		MP.printError(ec);
 		    	} else {
-		    		MP.printError(ec, param);
+		    		MP.printError(ec, params);
 		    	}
 				this.trace.printTrace(curState, succState);
 				this.theStateQueue.finishAll();
@@ -1181,10 +1181,11 @@ public class ModelChecker extends AbstractChecker
 				// Check properties of the state:
 				if (!seen || forceChecks) {
 					for (int j = 0; j < tool.getInvariants().length; j++) {
-						if (!tool.isValid(tool.getInvariants()[j], curState)) {
+						FalseExpressionWithDetails falseExpr = this.tool.checkValidity(this.tool.getInvariants()[j], curState, TLCState.Empty);
+						if (falseExpr != null) {
 							// We get here because of invariant violation:
 							MP.printError(EC.TLC_INVARIANT_VIOLATED_INITIAL,
-									new String[] { tool.getInvNames()[j].toString(), tool.evalAlias(curState, curState).toString() });
+									new String[] { tool.getInvNames()[j].toString(), tool.evalAlias(curState, curState).toString(), falseExpr.prettyPrint() });
 							if (!TLCGlobals.continuation) {
 								this.errState = curState;
 								returnValue = EC.TLC_INVARIANT_VIOLATED_INITIAL;
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/impl/DebugTool.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/impl/DebugTool.java
@@ -1,5 +1,6 @@
 /*******************************************************************************
  * Copyright (c) 2020 Microsoft Research. All rights reserved. 
+ * Copyright (c) 2025, Oracle and/or its affiliates.
  *
  * The MIT License (MIT)
  * 
@@ -48,6 +49,7 @@ import tlc2.tool.INextStateFunctor;
 import tlc2.tool.INextStateFunctor.InvariantViolatedException;
 import tlc2.tool.IStateFunctor;
 import tlc2.tool.ITool;
+import tlc2.tool.FalseExpressionWithDetails;
 import tlc2.tool.TLCState;
 import tlc2.tool.TLCStateFun;
 import tlc2.tool.TLCStateMutExt;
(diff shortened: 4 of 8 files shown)


Calvin-L:
I'm interested in feedback on the phrasing of the error message and exactly what information it spits out. (And also of course on the implementation code.)

I think there are probably some improvements to make before this gets merged.

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/DebugTool.java
```diff
@@ -123,20 +124,28 @@ public boolean isValidAssumption(final ExprNode assumption) {
...
-	public boolean isValid(Action act, TLCState state) {
+	public InvariantViolation findFalsifyingContext(Action invariant, TLCState s0, TLCState s1) {
+		// Need this override for breakpoints to work
+		return isValid(invariant, s0, s1)
+			? null
+			: new InvariantViolation(invariant.pred, Context.Empty);
+	}
+
+	@Override
+	public boolean isValid(Action act, TLCState s0, TLCState s1) {
```
Calvin-L, line 135: I had to rework the way `DebugTool` overrides `isValid` or else there are hangs and failures in the debugger tests. I _think_ this new approach is better because
 - `isValid(Action, TLCState)` (with 1 state) is implemented in terms of this method
 - so overriding `isValid(Action, TLCState, TLCState)` suffices to handle both overloads

But there could be something I'm missing.

tlatools/org.lamport.tlatools/src/tlc2/tool/ModelChecker.java
```diff
@@ -1181,10 +1182,11 @@ public Object addElement(final TLCState curState) {
 				// Check properties of the state:
 				if (!seen || forceChecks) {
 					for (int j = 0; j < tool.getInvariants().length; j++) {
-						if (!tool.isValid(tool.getInvariants()[j], curState)) {
+						InvariantViolation counterexample = this.tool.findFalsifyingContext(this.tool.getInvariants()[j], curState, TLCState.Empty);
+						if (counterexample != null) {
 							// We get here because of invariant violation:
 							MP.printError(EC.TLC_INVARIANT_VIOLATED_INITIAL,
-									new String[] { tool.getInvNames()[j].toString(), tool.evalAlias(curState, curState).toString() });
+									new String[] { tool.getInvNames()[j].toString(), tool.evalAlias(curState, curState).toString(), counterexample.formatPretty() });
```
lemmy, line 1189: [nit] The terminology elsewhere in TLC is "prettyprint".

tlatools/org.lamport.tlatools/src/tlc2/tool/ITool.java
```diff
@@ -127,6 +129,17 @@ public interface ITool extends TraceApp, OpDefEvaluator {
...
+	/**
+	 * Evaluate the given invariant in the given state and, if it evaluates to FALSE, return information about why.
+	 *
+	 * @param invariant the invariant
+	 * @param s0        the current state
+	 * @param s1        the next state (or {@link TLCState#Empty})
+	 * @return information about why the invariant evaluates to FALSE, or {@code null} if the invariant evaluates to TRUE
+	 * @throws util.Assert.TLCRuntimeException if the invariant does not evaluate to a boolean value
+	 */
+	InvariantViolation findFalsifyingContext(Action invariant, TLCState s0, TLCState s1);
```
lemmy, line 141: Could this be generalized to `ITool#findInvalidAssignment*` methods to align with the existing `ITool#isValid*` methods? I would even be supportive of a refactoring of the return values of the various `isValid` methods. Additionally, should we introduce a constant level variant that accepts an `ExprNode` instead of an `Action`?

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/InvariantViolation.java
```diff
@@ -0,0 +1,58 @@
...
+
+import tla2sany.semantic.SemanticNode;
+import tlc2.util.Context;
+
+/**
+ * Information about an invariant violation.  This class has no particular structure, only a single
+ * {@link #formatPretty()} method to display the information.  Today it shows a location of a relevant FALSE
+ * expression and the values of any {@code \A}-bound variables.
+ */
+public class InvariantViolation {
```
lemmy, line 13: The naming is inconsistent, using "InvariantViolation" in one place and "find ... Context" in another. It might be clearer to rename "InvariantViolation" to "Assignment" (after verifying it doesn’t conflict with `tlc2.model.Assignment`). Additionally, it’s worth considering whether `InvariantViolation` and `tlc2.value.impl.CounterExample` should be merged into one. Combining them could simplify the process of including this information when counterexamples are serialized in various formats (such as JSON or TLA+).

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/Tool.java
```diff
@@ -3486,9 +3486,112 @@ else if (val == null) {
...
+              s0,
+              s1,
+              invariant.cm,
+              invariant.con);
+    }
+
+    /**
+     * Helper for {@link ITool#findFalsifyingContext(Action, TLCState, TLCState)}.
+     */
+    private InvariantViolation findFalsifyingContext(SemanticNode expr, Context c, TLCState s0, TLCState s1, CostModel cm, Context relevantAssignments) {
```
lemmy, line 3503: I didn't review the logic, assuming it has been largely copied from `isValid`?!

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/Tool.java
```diff
@@ -3486,9 +3486,112 @@ else if (val == null) {
...
+            // fall thru to default case
+        }
+
+        Value val = this.eval(expr, c, s0, s1, EvalControl.Clear, cm);
+        if (!(val instanceof BoolValue)) {
+            Assert.fail(EC.TLC_EXPECTED_VALUE, new String[]{"boolean", expr.toString()}, expr, c);
+        }
+        return ((BoolValue) val).val
+                ? InvariantViolation.NO_VIOLATION
+                : new InvariantViolation(expr, relevantAssignments);
```
lemmy, line 3589: This could simply return `Context` if we can be certain that `expr` is guaranteed to always be a `SymbolNode` instead of a `SemanticNode`.

tlatools/org.lamport.tlatools/src/tlc2/util/Context.java
```diff
@@ -236,6 +239,23 @@ public Context deepCopy() {
 		}
 		return new Context(this.name, this.value, this.next.deepCopy());
 	}
+
+	public Context replaceSuffix(Context oldSuffix, Context newSuffix) {
```
lemmy, line 243: Does this mutate the existing `Context` instance?  If not, should this rather be static?

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/DebugTool.java
```diff
@@ -123,20 +124,28 @@ public boolean isValidAssumption(final ExprNode assumption) {
 	}
 
 	@Override
-	public boolean isValid(Action act, TLCState state) {
+	public InvariantViolation findFalsifyingContext(Action invariant, TLCState s0, TLCState s1) {
+		// Need this override for breakpoints to work
```
lemmy, line 128: What exactly doesn't work without this override?

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/DebugTool.java
lemmy: I don't see any issues with the proposed change.

Note: You can use `DebugTool#noDebug` to obtain an instance of `ITool`, which evaluates expressions outside of the debugger's control.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -498,6 +499,9 @@ private static String getMessage0(int messageClass, int messageCode, String[] pa
             break;
         case EC.TLC_INVARIANT_VIOLATED_INITIAL:
             b.append("Invariant %1% is violated by the initial state:\n%2%");
+            if (parameters.length > 2) {
+                b.append("\nMore information: %3%");
```
lemmy, line 503: The prefix "More information: " (used here and below) doesn’t provide much value to readers. How can we make deserialization easier for downstream tools like our VSCode extension?  Can we avoid regex-based parsing, or make it dead simple?

tlatools/org.lamport.tlatools/src/tlc2/util/Context.java
Calvin-L: It constructs a new `Context`. Contexts are (thankfully!) immutable.

Perhaps `withDifferentSuffix` might be a better name? `with` seems to be the word Java developers usually use for methods that return a modified copy (as opposed to modifying in-place).

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/Tool.java
Calvin-L: My thinking was that `expr` would point to the more specific part of the invariant under the `\A` binder:
```
\A x \in S: ....
            ^^^^
            expr
```
Having used this branch for a few days, I'm not convinced that adds much value.

Eliminating the `InvariantViolation` class and just returning a `Context` would certainly solve the [other discussion](https://github.com/tlaplus/tlaplus/pull/1180#discussion_r2070527414) about naming. Even so, part of me wants to keep the wrapper class as a kind of documentation. The _meaning_ of the returned context is different than normal: it useful for evaluating expressions, it's part of an error report.

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/Tool.java
Calvin-L: Well, mostly copied from `evalApplImpl` (which is what you get if you inline a few methods). But essentially yes. :)

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/InvariantViolation.java
Calvin-L: I agree. I'll see if I can come up with something better.

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/DebugTool.java
Calvin-L: The old `isValid` method was just a wrapper to `eval`. The new `findFalsifyingContext` method actually bypasses some calls to `eval`---and something about that causes a test hang. I know that's not a very clear explanation, but I don't fully understand the problem myself yet.

tlatools/org.lamport.tlatools/src/tlc2/util/Context.java
lemmy: Or perhaps a new constructor `Context(Context orig, Context oldSuffix, Context newSuffix)`?

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
ahelwer: The answer is probably to implement a LSP API instead of investing in parsing the text output of the tool.

lemmy: Ideally, TLC would be fully machine-readable (and controllable).  Better yet, in a way that eliminates the need for parsing altogether (think proper IPC). Unfortunately, I believe this is beyond the scope if this issue.

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/Tool.java
Calvin-L: > Having used this branch for a few days, I'm not convinced that adds much value.

Actually I've changed my mind. The specific false subexpression _does_ provide value for invariants with many conjuncts. `TypeOK` invariants usually follow that pattern, and it can be very helpful to see exactly which conjunct was false.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
Calvin-L: It is an obscure term, and I'm open to suggestions.

Maybe "Values of relevant names" would be a bit friendlier?

ahelwer: I kinda like the term. It was immediately obvious what it meant from the PR title, and I've struggled to find any better term (using the clunky "bound identifiers") in the build-your-own-TLA-tools tutorial.

PR #1181, opened 2025-05-01
Add extended documentation how to handle `\E s \in Seq(S):...` and `\A s \in Seq(S):...`

lemmy:
[Feature][TLC]

### DIFF
--- a/docs/seq-unenumerable.md
+++ b/docs/seq-unenumerable.md
@@ -1,0 +1,95 @@
+# Replacing `Seq` with `BoundedSeq` for Finite-State Model Checking with TLC
+
+```tla
+----- MODULE Simple -----
+EXTENDS Sequences
+
+S == {"a", "b"}
+
+VARIABLE log
+
+Spec == log = <<>> /\ [][\E s \in Seq(S) : log' = log \o <<s>>]_log
+====
+```
+
+## Problem
+
+When you run the above TLA+ specification through TLC, you will encounter the following error:
+
+```
+TLC encountered the non-enumerable quantifier bound
+Seq({"a", "b"})
+line 8, col 35 to line 8, col 40 of module Simple
+In TLA+, Seq(S) represents the set of all finite sequences whose elements come from the set S. Even when S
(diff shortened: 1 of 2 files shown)


lemmy:
DNS CNAME record has been requested.

lemmy:
This is inspired by how Rust explains its errors:
* https://github.com/rust-lang/rust/tree/master/compiler/rustc_error_codes/src/error_codes
* https://doc.rust-lang.org/error_codes/

lemmy:
@ahelwer @FedericoPonzi as per https://github.com/tlaplus/rfcs/issues/15

lemmy:
I'm considering using a static site generator to format and present the extended help files (in Markdown) that will eventually be hosted at http://explain.tlapl.us. Here are some potential options:

- [Jekyll](https://jekyllrb.com/) (Ruby)
- [MkDocs](https://www.mkdocs.org/) (Python)
- [mdBook](https://rust-lang.github.io/mdBook/index.html) (Rust)
- [Hugo](https://gohugo.io/) (Go)

For reference:
- https://foundation.tlapl.us and https://conf.tlapl.us/ currently use Hugo.
- https://apalache-mc.org/docs/ uses mdBook.

It's not clear if any of these tools support decent TLA+ syntax highlighting out of the box. @ahelwer ?

https://github.com/peaceiris provides GH actions to deploy Hugo, mdbook, ...

ahelwer:
I don't believe any of them currently support TLA+ syntax highlighting. I've been meaning to write regex highlighting files for Hugo for TLA+ since forever since I also use that as a SSG for my own site (I think I usually use Haskell or Idris for TLA+ blocks).

Closest thing is I think someone got Jekyll to work with tree-sitter for code highlighting and we have a tree-sitter grammar for TLA+ but I don't know how mature or maintained that Jekyll plugin is.

I would support using Hugo just because I'm familiar with it and we also use it for the foundation website.

lemmy:
Since none of the four contenders seem to support TLA+ syntax highlighting (right away), we might as well take the simpler approach and serve `explain.tlapl.us` directly from the master branch.  A more sophisticated solution with proper syntax highlighting can always be added later.

Issue #1184, opened 2025-05-10
Raise warning when definition or declaration appears as Record field name


lemmy:
The formula `SeemsObvious` is false for any value of `C1`, except for the one value `"C1"`. 

```tla
---- MODULE Foo ----
CONSTANTS C1
SomeSet == {C1}

SomeRecord == 
  [
    C1 |-> 1
  ]

SeemsObvious == 
    DOMAIN SomeRecord = SomeSet

\* A more real-world scenario.
Next == \E v \in SomeSet: SomeRecord[v] = 1

\* TLAPS proves this theorem
THEOREM DOMAIN SomeRecord = {"C1"} BY DEF SomeRecord
====
```

This appears sufficiently counterintuitive to warrant a warning. ~Unfortunately, such a warning can only be raised during model checking in the general case.~ SANY could issue a warning when it sees the symbol of the `CONSTANT C1` appear as a record field. Thoughts?

/cc @craft095

ahelwer:
Ooooh, tricky. I agree SANY could detect this in the record constructor. It would not be able to catch the case where they declare `CONSTANT C1` *after* constructing the record, though.

lemmy:
Since CONSTANTS typically appear first in most specifications, adding a check in the record constructor seems like a good starting point. Additional checks, even those that may require multiple passes, can always be introduced later.

lemmy:
Warning could be raised in `processRcdForms` by checking if `symbolTable.resolveSymbol(labels[lvi]) != null`.

https://github.com/tlaplus/tlaplus/blob/b3e5c565d38f717912d267dd86174f496519d169/tlatools/org.lamport.tlatools/src/tla2sany/semantic/Generator.java#L4046-L4089

lemmy:
> It would not be able to catch the case where they declare `CONSTANT C1` _after_ constructing the record, though.

Since symbols in TLA+ must be declared or defined **before** they are used, most users will likely recognize that the domain of `SomeRecord` cannot be `42`.

```tla
---- MODULE Foo ----

SomeRecord == 
  [ C1 |-> 1 ]

C1 ==
   42
====
```

However, this example makes me wonder if detecting this scenario might actually require a second pass. When parsing `C1`, SANY will already have encountered `"C1"`, so the check could potentially be the inverse of the one introduced in https://github.com/tlaplus/tlaplus/pull/1185.

ahelwer:
[...]

PR #1185, opened 2025-05-10
Raise warning when symbol appears in Records.

lemmy:
Addresses Github issue #1184
https://github.com/tlaplus/tlaplus/issues/1184

[Feature][TLC]

```
The field name "SomeField" in the record constructor is identical to the existing definition or declaration
named SomeField, located at line 5, col 1 to line 6, col 4 of module RiverCrossingFlashlight.
The field in the record will not take the value of the SomeField definition or declaration.
In TLA+, field names in records are strings, regardless of any similarly named declarations or definitions.
Therefore, DOMAIN [SomeField |-> ...] = {"SomeField"} holds true.
```


ahelwer (approved):

ahelwer:
Changed look good. Should also add a test case to the error corpus to trigger this code.

PR #1188, opened 2025-05-16
Fixed JavaCC warnings introduced by Unicode work

ahelwer:
Warnings were added in PR #896
Warning types fixed:
1. Warning:  "\u2227" cannot be matched as a string literal token at line 1261, column 33. It will be matched as  `<AND>`.
2. Warning: Line 1316, Column 3: Non-ASCII characters used in regular expression.

There were three warnings of type (1), for `<AND>`, `<OR>`, and `<IN>`. The
reason for this warning was that these three tokens function both as
infix operators and also keywords. The keyword tokens were called `<AND>`,
`<OR>`, and `<IN>`. During the unicode work the ∧, ∨, and ∈ synonyms were
added to the `<AND>`, `<OR>`, and `<IN>` token definitions. However, they were
also defined in separate tokens named `<op_infix_land_uc>`,
`<op_infix_lor_uc>`, and `<op_infix_in_uc>` (the same as all other Unicode
operator variants). The warning was just saying that these three _uc token
definitions were being ignored in favor of `<AND>`, `<OR>`, and `<IN>`. The
solution was simply to delete the ignored _uc token definitions.

The warning of type (2) was because the ℕ, ℤ, and ℝ characters were
added to the general `<IDENTIFIER>` token regex. This was fixed by setting
the UNICODE_INPUT option to true. JavaCC now generates a "note" instead
of a warning saying to make sure to use a reader with the proper input
encoding, which indeed we do.

There is one remaining warning about a lookahead ambiguity in the CASE statement. This warning was present before the Unicode work.

[SANY]


tlatools/org.lamport.tlatools/javacc/tla+.jj
```diff
@@ -79,7 +79,7 @@ config.jj.
 options {
   LOOKAHEAD = 1;
   JAVA_UNICODE_ESCAPE = false; // disable processing of \\u
-  UNICODE_INPUT = false;
+  UNICODE_INPUT = true;
```
lemmy, line 82: What does this particular line do? The PR is about the removal of warnings.

ahelwer: That fixes the "Warning: Line 1316, Column 3: Non-ASCII characters used in regular expression." warning. Or turns it into a "note", really.

Calvin-L: More broadly, what does the `UNICODE_INPUT` option actually control? Does it affect the behavior of the generated parser, or does it affect how javacc reads `tla+.jj`?

ahelwer: Other than silencing the warning: isolating the diff from toggling the setting on & off, it adds the following snippet to two places in `TLAplusParserTokenManager.java`:
```java
int hiByte = (int)(curChar >> 8);
int i1 = hiByte >> 6;
long l1 = 1L << (hiByte & 077);
```
`curChar` is the current character in the stream, of `char` datatype, which in Java is 16 bits. Variables `i1` and `l1` are never used so can be ignored. `hiByte` is only used in the definitions of `i1` and `l1` so can also be ignored. So basically it's a no-op.

The [JavaCC documentation](https://javacc.github.io/javacc/documentation/grammar.html) states:

>When set to true, the generated parser uses uses an input stream object that reads Unicode files. By default, ASCII files are assumed. This option is ignored if either of options USER_TOKEN_MANAGER, USER_CHAR_STREAM is set to true.

lemmy: It would be helpful to include this information as a code comment.

lemmy (approved):
LGTM - everything beyond `tlatools/org.lamport.tlatools/javacc/tla+.jj` is generated, right?

ahelwer:
Correct, tla+.jj is the only hand-modified file.

ahelwer:
@Calvin-L does this look good?

Issue #1186, opened 2025-05-12
Feature request: More control over warnings in TLA+ tools


craft095:
In #1184 a new warning is proposed. I like the idea of adding warnings for confusing cases, and I believe such warnings are very important for those who are just starting to use TLA+. However, if we want to add more warnings, it would be great to have more control over them — the ability to disable, enable, or treat a warning as an error.
For example, in our case (automotive safety), we have to treat all warnings as errors because some indicate truly dangerous situations, and there's no way to selectively enable or disable specific warnings.

lemmy:
There’s value in supporting three distinct message levels: *Info*, *Warning*, and *Error*. For example, issue #1184 falls somewhere between an info and a warning, making its classification debatable. In fact, message severity will often be subjective, and/or depend on the context in which the tools is used. To handle this, each message should be tagged with a key, and our tools should let users customize the severity level—allowing them to reclassify messages as info, warning, or error based on their own context.

lemmy:
@craft095 Would you be able to drive this work?  May I assign this issue to you?

ahelwer:
# Feature Proposal
Internally, we should tag all SANY log messages with one of the following ascending levels:
1. Trace
1. Debug
1. Info
1. Lint (for things like #1185)
1. Warning
1. Error

Then, two more optional command-line parameters should be added to SANY:
- `-logLevel=`, which is set to `Info` by default
- `-errorLevel=`, which is set to `Error` by default

The function of `-logLevel=` is to only output log messages to stderr or stdin if they are the given level or above. So, `-logLevel=Error` would implement @hwayne's requested "silent mode" feature in #1072 (at least for SANY) by suppressing logging of messages like "Semantic processing of module Naturals" (which would be either Debug or Info level).

The function of `-errorLevel=` is primarily to control whether a nonzero exit code is returned if a message with the given level or higher is emitted. Secondarily it also controls whether messages are routed to stdout or stderr. Associated with this, I believe we should deprecate the `-error-codes` flag. SANY should simply return a nonzero exit code on error by default.

lemmy:
The TLA+ tools do not follow the typical `-flag=val` (equals) command-line pattern. Also, are you expecting TLC to use the same parameters when it internally invokes SANY? For example, something categorized as a lint warning might actually originate from TLC when it e.g. encounters an unusual entry in a .cfg file.

ahelwer:
[...]

PR #1194, opened 2025-06-10
SANY error logging: aborts are now errors with exceptions

ahelwer:
These changes support #1186

Remove unnecessary delineation to support adding more levels

Before these changes, there were separate methods in the Errors class for adding "aborts" vs. adding "errors". The only actual difference between these was that adding an abort also threw an exception if the caller set a boolean parameter to true. Every caller to Errors::addAbort() set this parameter to true. So, the Errors::addAbort() method has been unified with the Errors::addError() method. The Errors::addError() method has also been modified so it returns an AbortException that can be optionally thrown by the caller. Internally, aborts and errors go into the same list.

Why make these changes? We want to support additional logging levels for better control of SANY output. Within a schema of additional log levels, the distinction between errors and aborts adds unnecessary complexity to both the internal logic of the Errors class and the external API. Since the only benefit of this distinction was throwing an exception, something that the caller can now easily do, it makes sense to simplify the logging code by removing the distinction.

There was some abort-specific code in the toolbox IDE. In one case this was modified to make use of the new standardized error code, and in the other was translated to a functional equivalent.

There was one additional change: throwing an exception at the top of the Generator::generateNumerableStep() method made the Java compiler complain that the rest of the method was unreachable, which was true, because the reason the abort was being thrown was to protect the method from being called at all. A comment indicates confusion about why the method exists, because it isn't used. The method was thus deleted.

[SANY][Refactoring]

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
@@ -284,7 +284,7 @@ public class SANY {
           if (semanticErrors.getNumMessages() > 0) {
             syserr.println("Semantic errors:\n\n" + semanticErrors);
             // indicate fatal error during semantic analysis or level-checking
-            if ( semanticErrors.getNumAbortsAndErrors() > 0 ) {
+            if ( semanticErrors.getNumErrors() > 0 ) {
               spec.errorLevel = 4;
             } // end if
           } // end if
@@ -306,7 +306,7 @@ public class SANY {
         }
         
         // indicate fatal error during semantic analysis or level-checking
-        if ( semanticErrors.getNumAbortsAndErrors() > 0 ) { 
+        if ( semanticErrors.getNumErrors() > 0 ) { 
           spec.errorLevel = 4;
         }
       }
--- a/toolbox/org.lamport.tla.toolbox/src/org/lamport/tla/toolbox/spec/parser/ModuleParserLauncher.java
+++ b/toolbox/org.lamport.tla.toolbox/src/org/lamport/tla/toolbox/spec/parser/ModuleParserLauncher.java
@@ -31,7 +31,9 @@ import tla2sany.modanalyzer.ParseUnit;
 import tla2sany.modanalyzer.SpecObj;
 import tla2sany.parser.ParseException;
 import tla2sany.semantic.AbortException;
+import tla2sany.semantic.ErrorCode;
 import tla2sany.semantic.Errors;
+import tla2sany.semantic.Errors.ErrorDetails;
 import tla2sany.semantic.ExternalModuleTable;
 import util.FilenameToStream;
 import util.ToolIO;
@@ -412,12 +414,13 @@ public class ModuleParserLauncher
 
                 // Test for special case of importing unknown module. (added 5
                 // Sep 2009 by lamport)
-                String[] abortMsgs = result.getParseErrors().getAborts();
-                if ((abortMsgs.length > 0) && (abortMsgs[0].indexOf("Cannot find source file for module") != -1))
-                {
-                    parsingModuleIndex = abortMsgs[0].indexOf("imported in module ") + 19;
-                    nameToFind = abortMsgs[0].substring(parsingModuleIndex, abortMsgs[0].indexOf(".",
-                            parsingModuleIndex + 1));
+                for (final ErrorDetails abortDetails : result.getParseErrors().getErrorDetails()) {
+                    if (abortDetails.getCode() == ErrorCode.MODULE_FILE_CANNOT_BE_FOUND) {
+                        final String abortMsg = abortDetails.getMessage();
+                        parsingModuleIndex = abortMsg.indexOf("imported in module ") + 19;
+                        nameToFind = abortMsg.substring(parsingModuleIndex, abortMsg.indexOf(".",
+                                parsingModuleIndex + 1));
+                    }
                 }
 
                 // Correct the capitalization of nameToFind, if necessary.
@@ -489,11 +492,12 @@ public class ModuleParserLauncher
                     // the abort in parseErrors
                     if (result.getParseErrors() != null)
                     {
-                        String[] aborts = result.getParseErrors().getAborts();
+                        String[] aborts = result.getParseErrors().getErrors();
                         if (aborts.length > 0)
                         {
-                            // error message
-                            message = aborts[0];
+                            // The error message triggering the abort will be
+                            // the most recent, so get the last message.
+                            message = aborts[aborts.length - 1];
                         }
                     }
                     // Unless this is the one abort in which err.moduleName can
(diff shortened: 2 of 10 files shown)


toolbox/org.lamport.tla.toolbox/src/org/lamport/tla/toolbox/spec/parser/ModuleParserLauncher.java
```diff
@@ -412,12 +414,13 @@ private void processParsingErrors(IProject project, ParseResult result)
...
-                    parsingModuleIndex = abortMsgs[0].indexOf("imported in module ") + 19;
-                    nameToFind = abortMsgs[0].substring(parsingModuleIndex, abortMsgs[0].indexOf(".",
-                            parsingModuleIndex + 1));
+                for (final ErrorDetails abortDetails : result.getParseErrors().getErrorDetails()) {
+                    if (abortDetails.getCode() == ErrorCode.MODULE_FILE_CANNOT_BE_FOUND) {
+                        final String abortMsg = abortDetails.getMessage();
+                        parsingModuleIndex = abortMsg.indexOf("imported in module ") + 19;
+                        nameToFind = abortMsg.substring(parsingModuleIndex, abortMsg.indexOf(".",
+                                parsingModuleIndex + 1));
+                    }
```
ahelwer, line 423: I think this code is a good exemplar of what parser clients can do with the new standardized error codes! There remains the fragile kludge of parsing the error message to extract information, though... perhaps the `ErrorDetails` class should adopt the approach of containing an `Object[]` array that is interpolated into a format string, so clients like this can directly access the array of objects to extract specific information from the error message? This would be a place where variants really shine but unfortunately Java doesn't have language support for that at the same level as Rust & OCaml.

lemmy: Why `Object[]` and not `String[]`?  What types do you envision?

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -306,7 +306,7 @@ public static void frontEndSemanticAnalysis(SpecObj spec,
         }
         
         // indicate fatal error during semantic analysis or level-checking
-        if ( semanticErrors.getNumAbortsAndErrors() > 0 ) { 
+        if (semanticErrors.isFailure()) {
```
lemmy: nit: The method name `Errors#isFailure` seems misleading, as errors are inherently failures—so the distinction doesn't quite make sense. Since the `Errors` class is essentially a custom collection, a method like `Errors#isEmpty` would be more consistent with naming conventions used in other collection types.

toolbox/org.lamport.tla.toolbox/src/org/lamport/tla/toolbox/spec/parser/ModuleParserLauncher.java
```diff
@@ -489,11 +492,11 @@ else if (message.indexOf("Lexical {error: EOF reached, possibly open comment sta
                     // the abort in parseErrors
                     if (result.getParseErrors() != null)
                     {
-                        String[] aborts = result.getParseErrors().getAborts();
+                        String[] aborts = result.getParseErrors().getErrors();
                         if (aborts.length > 0)
                         {
                             // error message
-                            message = aborts[0];
+                            message = aborts[aborts.length - 1];
```
lemmy, line 500: Why does the index change, why no longer the first but the last elem?

lemmy (approved):
See my comments.

toolbox/org.lamport.tla.toolbox/src/org/lamport/tla/toolbox/spec/parser/ModuleParserLauncher.java
ahelwer: We could probably get away with strings. The only non-string things that immediately come to mind are ints, for things like "operator accepts expressions of maximum level x, but expression was of level y"

lemmy: It would be better if TLC prints the human readable names constant, state, action, and temporal.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: Perhaps it depends on the direction we want to take the `Errors` class. I am envisioning it as a sort of one-stop shop for all messages emitted by SANY, so they can be filtered, tagged, routed & output all in one place. Warnings are already included in `Errors`, and probably that is where linter messages will be recorded as well. Maybe we'll rename the class eventually.

In this particular case I foresee changing this call to something like `anyMessagesOfLevelOrGreater(ErrorLevel level)`, so the caller can control what failure means. Or the caller instantiates `Errors` with an `ErrorLevel` parameter controlling what indicates a failure, and `isFailure()` relies on that. For now it's probably good to keep this as a call to `semanticErrors.getNumErrors()`.

toolbox/org.lamport.tla.toolbox/src/org/lamport/tla/toolbox/spec/parser/ModuleParserLauncher.java
ahelwer: This is a bit tricky to explain, but the reason why it's `aborts[0]` is because the code is implicitly assuming the only message in `aborts` is the error they're looking for. Given that every addition of an abort triggered an `AbortException`, and nothing in the codebase recovers from an `AbortException`, this assumption was fine (although if any additions of aborts ever *didn't* trigger an `AbortException` this assumption would have been wrong and thus this code would be buggy). Personally I would have chosen to write it as `aborts[aborts.length - 1]`, since semantically what we are looking for here is "the last abort error message" - it just happens to also be the first abort error message.

Since we are now co-mingling errors and aborts, we can no longer assume there will be only a single error message in `aborts`, since some recoverable error messages could have previously been added. So we take the last one, which will always be the one which caused the abort.

lemmy: Please add this as a code comment.

ahelwer: Possibly we could also attach the error details to the AbortException. I will look into this.

toolbox/org.lamport.tla.toolbox/src/org/lamport/tla/toolbox/spec/parser/ModuleParserLauncher.java
ahelwer: Some other examples I thought of are errors like "operator of arity x called with y parameters" or "parameter n of operator is itself an operator expecting m parameters" but also that information does not seem very actionable by clients. If we are going to include this expected payload format in the hypothetical future standardization of these error codes, I think string parameters make the most sense.

toolbox/org.lamport.tla.toolbox/src/org/lamport/tla/toolbox/spec/parser/ModuleParserLauncher.java
ahelwer: Added a code comment here as requested. In PR #1197 I also took the first step to addressing this, enabling a future change where this code doesn't need to look through the error messages at all.

lemmy:
@Calvin-L Are you reviewing as well?

ahelwer:
@lemmy is this good to merge?

lemmy:
Still waiting to hear back from @Calvin-L.

Calvin-L (approved):
Sorry for the delay. I have no objections here!

PR #1197, opened 2025-06-11
SANY: add ErrorDetails to SemanticException

ahelwer:
Somewhat related to #1186 and #1194 in a yak-shaving sense.

External SANY users can now directly get cause of exception

Previously, if SANY threw a SemanticException then users catching that exception would have to separately go to the Errors instance to hopefully get the error that caused the exception to be thrown. Now, the ErrorDetails instance is packed inside the SemanticException instance so users can get the exception cause directly. SemanticException is the user-facing wrapper for an AbortException, which is what is thrown by the Errors class and is used throughout the semantic-level checker (and parts of the syntactic checker).

One more piece of the puzzle remains: there are several places within the syntax part of SANY which nonetheless use the Errors class to trigger AbortException. Unfortunately, the SANY class throws away this AbortException instance and turns it into a ParseException. ParseException is a class that is auto-generated by JavaCC, so it cannot be modified to contain an ErrorDetails instance. Thus if users catch a ParseException they still need to go to an Errors instance to try to find the error which caused it. Future work will perhaps try to fix this problem.

[SANY][Refactor]


tlatools/org.lamport.tlatools/src/tla2sany/semantic/NumeralNode.java
```diff
@@ -39,20 +39,25 @@ public class NumeralNode extends ExprNode {
...
+       radix = 16;
      } else {
-    	 throw new AbortException();  // This shouldn't happen.
+       errors.addAbort(
+         ErrorCode.INTERNAL_ERROR,
+         stn.getLocation(),
+         "Unknown numeral format: " + num,
+         true
+       );
      }
```
ahelwer, line 60: Moved some whitespace (mix of tabs and spaces) around here because it was impossible to make the indentation look non-wonky in the diff otherwise.

tlatools/org.lamport.tlatools/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -45,7 +45,7 @@ org.eclipse.jdt.core.compiler.problem.missingEnumCaseDespiteDefault=disabled
 org.eclipse.jdt.core.compiler.problem.missingHashCodeMethod=ignore
 org.eclipse.jdt.core.compiler.problem.missingOverrideAnnotation=ignore
 org.eclipse.jdt.core.compiler.problem.missingOverrideAnnotationForInterfaceMethodImplementation=enabled
-org.eclipse.jdt.core.compiler.problem.missingSerialVersion=warning
+org.eclipse.jdt.core.compiler.problem.missingSerialVersion=ignore
```
Calvin-L, line 48: I endorse this change. 👍 

I am no expert on Java's serialization framework, but my current understanding is that
 - [the automatically-generated `serialVersionUID` suffices](https://github.com/projectlombok/lombok/wiki/WHY-NOT:-serialVersionUID)
 - long-term we should be moving away from Java's serialization framework (historically it has been riddled with security holes and other sharp edges---even at Oracle our internal guidance is to steer well clear of this feature)

(But @lemmy you might want to weigh in if you have any additional thoughts.)

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SemanticException.java
```diff
@@ -1,4 +1,56 @@
...
+   * The internal error wrapped by this exception.
+   */
+  private final AbortException internalException;
+
+  /**
+   * Initializes a new instance of the {@link SemanticException} class.
+   *
+   * @param internalException The internal exception causing the error.
+   */
+  public SemanticException(AbortException internalException) {
```
Calvin-L, line 22: This constructor should either:
 1. take `ErrorDetails` and `Errors` instead of an `AbortException`, or
 2. call `super(internalException)` to mark the abort as this exception's cause.

My preference is for the second, but I'd be happy with either.

Java exceptions are chained: each exception can point to a [cause](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Throwable.html#getCause()), so an exception can carry complete information about both the original error and which components caught and re-threw it.

By convention, an exception constructor that takes an exception as its argument is meant to build a link in the cause chain. It's a little weird that this class doesn't do that.

tlatools/org.lamport.tlatools/.settings/org.eclipse.jdt.core.prefs
lemmy: I would have preferred for this to be a dedicated commit, but I have no objections to ignoring this warning.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SemanticException.java
ahelwer: Okay! Thoughts on the new changes? The two getter methods call `this.getCause()` and cast it to an `AbortException`.

Calvin-L: Seems good to me!

Calvin-L (approved):

Issue #1198, opened 2025-06-13
Decide on & commit to code style guide


ahelwer:
In PR #987, @Calvin-L added a style guide to DEVELOPING.md. The style guide as it is can be read [here](https://github.com/tlaplus/tlaplus/blob/master/DEVELOPING.md#code-formatting-and-style).

@lemmy [proposes](https://github.com/tlaplus/tlaplus/pull/1195#issuecomment-2970677771) a different approach to code style which just uses Eclipse editor defaults regardless of surrounding code style (so tabs, basically).

I don't really care what the style guide ends up as long as we all agree to, commit to, and enforce (in PR reviews) a uniform style guide going forward. Personally I'm a two-space indentation guy, but editor settings are easy to change.

lemmy:
I don't think that creating and agreeing on a style guide is a good use of our time.  Especially that code styling becomes less and less of an issue in the AI age.

ahelwer:
Personally I find consistent indentation very useful for understanding code blocks.

lemmy:
I temporarily reformat the section of code I'm working on for local use, and then revert the changes afterward.

moiseev:
A passerby comment. A simple `.editorconfig` could go a long way towards the style enforcement (whatever the style is). It is supported by many editors/IDEs I know (I am not at all familiar with Eclipse). For a casual contributor it would be a bummer to get into bike-shedding in PR comments and/or need to install and master the new IDE to format the code. Likewise, it could create unwelcome burden to the maintainers if such casual contributions get merged and then need to be styled.

Less constructively, I would argue that code style uniformity is _more_ important in the age of AI as the complexity shifts from writing code to reading it, and style consistency is one of the few tools the reviewer has at their disposal. Apple's [goto fail](https://www.imperialviolet.org/2014/02/22/applebug.html) comes to mind. Very possible to miss that if tabs are sprinkled with spaces.

But again, just a passerby from the great internet. Take it or leave it.

lemmy:
This might very well be an issue in other projects, but I haven’t seen bike-shedding over code style in our PRs—mainly because we don’t have an enforced code style.  IMO, we should leave the existing code as-is. Contributors can use whatever code style they prefer in the parts they’re working on. The existing code shouldn’t be reformatted. If someone is making a small change, it might be a good idea to follow the surrounding style for consistency—but that should be a suggestion, not a rule.

[...]

PR #1200, opened 2025-06-19
Relax recommendation for tabs over spaces

Calvin-L:
Addresses #1198.

It was not my intention to be prescriptive here.  The default Eclipse settings use tabs.

Until we reach consensus on what style (if any) to standardize on, I don't think there's a reason to have a fine-grained rule in this slot. The number one rule above still stands:

 > 1. Modifications should copy the style of nearby code rather than change it.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -216,7 +216,7 @@ That said, it is worth following a few guiding principles:
 For new Java code we recommend (but do not require):
 
  - Each `public` class and method should have a javadoc description of its purpose and behavior.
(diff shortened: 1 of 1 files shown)


ahelwer:
Unfortunately @lemmy believes that the nearby code rule should not still stand.

lemmy:
I support the *recommendation* but not the *rule*.

Calvin-L:
I'm going to go ahead and merge this one, since I still think the suggestions in DEVELOPING.md should at least be compatible with the Eclipse defaults. We can continue to refine these recommendations in the future.

PR #1195, opened 2025-06-11
Raise warning when liveness property trivially true.

lemmy:
Closes Github issue #1193
https://github.com/tlaplus/tlaplus/issues/1193

[Feature][TLC]

```
Warning: Implied-temporal checking is a tautology.
The temporal formula Live at line 12, col 9 to line 12, col 50 of module Github1198 equals
the fairness constraints of the behavior specification Spec at line 14, col 1 to line 15, col 34 of module Github1198.
Therefore, checking whether Spec => Live using TLC will never result in a violation of Live.
```

Issue #1193, opened 2025-06-09
Raise warning when liveness property trivially true


lemmy:
Liveness and fairness are advanced concepts in TLA+, and it's not uncommon for beginner or novice users to mistakenly conjoin a fairness constraint `F` with the behavior specification `Spec`, and then attempt to check the property `F` itself. This results in a trivially true implication:

```tla
Spec /\ F => F
```

(Note that this issue is ignores the case where the implication is vacuously true due to safety part `S` of `Spec` is `FALSE`).

This kind of mistake can give a false sense of correctness and may be especially misleading for users who are still learning how to properly express and verify liveness properties.

### Suggested Improvement:

This spec bug will escape static analysis in the general. However, TLC could be enhanced to emit a warning in the special case where the property being checked is syntactically identical to a component of the specification. Such a warning would help guide newer users and improve the reliability of specifications verified using TLC.

lemmy:
@ahelwer @ahelwer @Calvin-L Comments or thoughts?

Calvin-L:
I think this is a good idea.

> This spec bug will escape static analysis in the general.

I actually think it is good to restrict the check to a simple syntactic one, and _not_ to do any deeper analysis. I think there exist specs + liveness properties that are both (1) important and (2) obviously true after a bit of unfolding definitions---but that should be cause for celebration, not alarm.

ahelwer:
Sounds like a good idea, I like all these new warnings. Will try to think of some myself.

PR #1201, opened 2025-06-21
Do not raise warning when liveness property is a refinement check when checking if property is trivially true

lemmy:
https://github.com/tlaplus/tlaplus/issues/1193
https://github.com/tlaplus/tlaplus/pull/1195


lemmy:
Manually checked and confirmed that none of the examples trigger the warning.

PR #1202, opened 2025-06-23
SANY: created unified errors.addMessage() error reporting method

ahelwer:
Also unified error reporting to single method

In commit be86598, error messages were tagged with a standardized error code. This code also contains an associated error level. Thus, in these changes the error reporting methods addError() and addWarning() are merged into a single method addMessage(), since the error level is already encoded in the standardized error code. The existing methods were tagged as deprecated. Some test code was updated to use the new method.

[SANY][Refactor]

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/Errors.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/Errors.java
@@ -1,33 +1,45 @@
-// Copyright (c) 2003 Compaq Corporation.  All rights reserved.
-// Portions Copyright (c) 2003 Microsoft Corporation.  All rights reserved.
-
-/***************************************************************************
-* Every Semantic node has an errors field that is an Errors object.  A     *
-* SpecObj object also has a few different kinds of Errors objects.  Here   *
-* are the relevant methods:                                                *
-*                                                                          *
-*    addWarning                                                            *
-*    addError                                                              *
-*    addAbort   : These methods add the indicated level of error.          *
-*                                                                          *
-*    isSuccess()                                                           *
-*    isFailure() : Indicates if addError or addAbort was called.           *
-*                                                                          *
-*    getNumErrors()                                                        *
-*    getNumAbortsAndErrors()                                               *
-*    getNumMessages()        : Return approximately obvious values.        *
-*                                                                          *
-*    toString() : Returns all the errors as a single string.               *
-***************************************************************************/
+/*******************************************************************************
+ * Copyright (c) 2003 Compaq Corporation. All rights reserved.
+ * Copyright (c) 2003 Microsoft Corporation. All rights reserved.
+ * Copyright (c) 2025 Linux Foundation. All rights reserved.
+ *
+ * The MIT License (MIT)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+ * of the Software, and to permit persons to whom the Software is furnished to do
+ * so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN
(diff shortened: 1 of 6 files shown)


toolbox/org.lamport.tla.toolbox/src/org/lamport/tla/toolbox/spec/parser/ModuleParserLauncher.java
```diff
@@ -536,7 +537,7 @@ else if (message.indexOf("Lexical {error: EOF reached, possibly open comment sta
             if (result.getSemanticErrors() != null)
             {
 
-                String[][] errors = { result.getSemanticErrors().getErrors(), result.getSemanticErrors().getWarnings() };
+                String[][] errors = { result.getSemanticErrors().getErrors(), result.getSemanticErrors().getMessagesOfAtLeastLevel(ErrorLevel.LINT).stream().map(ErrorDetails::toString).toArray(String[]::new) };
```
ahelwer: This is a bit unwieldy but really the entire loop below this should be refactored to use ErrorDetails. It parses the string error message to extract the location.

lemmy:
Would it be possible to retain the addError and addWarning methods (perhaps delegating to the new addMessage method) to minimize the size of the change set?

tlatools/org.lamport.tlatools/src/tla2sany/semantic/Errors.java
```diff
@@ -71,58 +83,124 @@ public boolean equals(Object o) {
...
+    final List<ErrorDetails> lint = this.getMessagesOfLevel(ErrorLevel.LINT);
+    ret.append((lint.size() > 0) ? "*** Lint: " + lint.size() + "\n\n" : "");
+    for (final ErrorDetails error : lint) {
       ret.append(error.toString() + "\n\n\n");
     }
 
     return ret.toString();
   }
+
+  private static boolean atLeastLevel(ErrorCode code, ErrorLevel base) {
```
lemmy: `ErrorLevel` appears to assume a total order. Are SANY messages actually totally ordered, or is `ErrorLevel` functioning more like an `ErrorCategory`, where values represent categories with no inherent or only partial ordering? Additionally, there seems to be some conceptual overlap with `severity`.

ahelwer: It is a total order. This will eventually be the basis for a --logLevel and --errorLevel CLI parameter to control output & what is interpreted as an error.

lemmy: My question is if there will always be a total order.  For example, one "category" could be heuristic-based warnings and another would be deterministic-based warnings.  I don't think we will be able to come up with a universially agreed on ordering.

ahelwer: Every logging system I know of manages to use a linear order of log levels. I don't see the purpose of categorizing messages in a non-linear hierarchy. What would be the use-case of that?

lemmy: Use-case: Heuristic-based vs deterministic-based warnings

ahelwer: A use case is like "user will be able treat Lint, Warning, and Error level messages as errors (output to stderr and return nonzero exit code) using the command line parameter `-errorLevel lint`". What is a comparable use case for heuristic-based vs. deterministic-based warnings?

lemmy: "Allow users to suppress deterministic warnings using some command-line parameter".

By the way, `errorLevel` isn't the best term to use here—even if it's commonly used by other logging frameworks.

ahelwer: It seems preferential to conform to common user experience conventions as much as possible. Software engineer expectation about tool behavior & parameters is a valuable repository of existing knowledge we can rely on instead of having to re-educate people about how our tool is special.

ahelwer:
Changed the `addWarning` and `addError` methods to just wrap `addMessage`. Marked as `@Deprecated` so future users use `addMessage` instead.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/Errors.java
lemmy: That design guideline wasn’t in place when I joined the project, and it hasn’t been formally established since. While software engineers are part of the TLA+ user base, they don’t represent it in full. Additionally, software engineering practices often follow trends or conventions that aren’t always grounded in lasting value. So while common UX patterns can be useful, their existence alone doesn’t justify adopting them.

ahelwer: Reduced scope of PR. This PR just unifies addWarning() and addError() into addMessage(), we can decide on Lint (or alternative name) levels and whether they are ordered later.

Calvin-L:
@ahelwer It looks like there are merge conflicts. This seems like a fine change to me; do you still plan to fix it up?

ahelwer:
Rebased! Thanks for checking, @Calvin-L.

ahelwer:
@Calvin-L looks like CI passes.

ahelwer:
@Calvin-L I don't think @lemmy has time to review this so it is probably good to merge.

ahelwer:
@lemmy could you merge if you don't have time to review? Thanks.

PR #1204, opened 2025-06-23
Do not ignore silently but raise warning if an invariant is a *constant-level* formula.

lemmy:
```
Warning: The invariant ConstLevelInvariant is a constant-level formula (i.e., it contains no variables, primes, or temporal operators) and evaluates to TRUE. To assert constant-level formulas in your spec, use ASSUME ConstInv, or to give the assumption a name such as YourAssumption, use ASSUME YourAssumption == ConstInv.
```

[Feature][TLC]

- [ ] Add new error code `2149` to to VSCode extension similar to https://github.com/tlaplus/vscode-tlaplus/commit/73058ac5aed4c04375b28b699aa746c75f2bf603

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/output/EC.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/output/EC.java
@@ -76,6 +76,7 @@ public interface EC
     public static final int TLC_INVARIANT_VIOLATED_BEHAVIOR = 2110;
     public static final int TLC_INVARIANT_EVALUATION_FAILED = 2111;
     public static final int TLC_INVARIANT_VIOLATED_LEVEL = 2146;
+    public static final int TLC_INVARIANT_CONSTANT_LEVEL = 2149;
     public static final int TLC_ACTION_PROPERTY_VIOLATED_BEHAVIOR = 2112;
     public static final int TLC_ACTION_PROPERTY_EVALUATION_FAILED = 2113;
     public static final int TLC_DEADLOCK_REACHED = 2114;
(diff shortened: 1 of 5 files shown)


lemmy:
The text below could be added as an extended explanation to the docs/ folder.  Not sure if this is necessary, or if the short error message is good enough.

----

In TLA+, `CONSTANT` values are fixed for the entire model and should **not** be verified as invariants. Invariants describe properties of **states** and must hold in **every reachable state**, but a `CONSTANT` is not part of the changing state—it’s more like a parameter or axiom of the system.

Here is a TLA+ module that demonstrates this idea:

```tla
----------------------------- MODULE CheckConstantAsAssume -----------------------------
EXTENDS Integers

(* --CONSTANTS-- *)
CONSTANT N  \* Do not check this as an invariant!

(* ASSUME for constant *)
ASSUME N \in Nat

(* VARIABLE for state-level changes *)
VARIABLE x

(* Initial predicate *)
Init == x = 0

(* Next-state relation *)
Next == x' = x + 1

(* A valid invariant: checks a state-dependent property *)
Inv == x <= N

(* INVALID invariant: this is not meaningful as an invariant! *)
(* BadInvariant == N > 0 *) \* This is a constant-level formula, not a state-level one.

```

### Explanation of the rule:

#### ❌ Do **not** do:

```tla
BadInvariant == N > 0
```

Why? Because `N` is constant; `N > 0` is **always either true or false**, and never changes with states. This expression does **not depend on variables**, so it's not meaningful to test as an invariant.

#### ✅ Do:

```tla
ASSUME N \in Nat
```

Use `ASSUME` to constrain `CONSTANT`s. This ensures the model checker only runs with valid constant values.

#### ✅ Also do:

```tla
Inv == x <= N
```

This is a legitimate invariant because it references both a `VARIABLE` (`x`) and a `CONSTANT` (`N`) in a way that verifies the behavior of the system over states.

---

### Rule (in plain language):

> **Never check a `CONSTANT` value using an `INVARIANT`. Instead, use `ASSUME` to restrict its value. `INVARIANT` should be used only with state-level formulas—i.e., formulas that depend on `VARIABLE`s.**

PR #1206, opened 2025-06-26
Added Nix dependencies shell file

ahelwer:
Updated DEVELOPING.md with Git requirement

Added a shell.nix file which, when users run nix-shell in the root directory of this repo, will drop the user into a shell with all required dependencies to build both the tools and toolbox. As part of the process of creating this shell.nix file, it was discovered that the toolbox build process has a dependency on Git. Thus the DEVELOPING.md instructions were updated to list Git as a dependency to build the toolbox.

Since both myself and @Calvin-L are nix users this seems like a good addition, especially since by default my system uses a Java version higher than 11. Nix users will look for & recognize the `shell.nix` file and know how to use it.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -15,6 +15,7 @@ There are other files and directories beyond these, but these are the most impor
 ├── CONTRIBUTING.md             # Info on contributing to the project
 ├── DEVELOPING.md               # This document
 ├── pom.xml                     # The Maven build system definition file
+├── shell.nix                   # Nix development dependencies shell file
 ├── toolbox/                    # All files related to the TLA⁺ Toolbox IDE
 ├── tlatools/
 │   └── org.lamport.tlatools/   # All files related to the TLA⁺ Tools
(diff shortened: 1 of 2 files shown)


lemmy:
> Since both myself and @Calvin-L are nix users this seems like a good addition, especially since by default my system uses a Java version higher than 11. Nix users will look for & recognize the `shell.nix` file and know how to use it.

This is not a justification for adding this—or anything else—to the TLA⁺ repository.

ahelwer:
You don't have to use nix for others to benefit! Quite a few projects these days have shell.nix or flake.nix files in their roots.

lemmy:
Adding tools like Nix to the TLA⁺ repository would create the expectation that Nix is officially supported. Since there is no consensus to support this development tool, it should not be included in the repository.

ahelwer:
In your view, what is the consensus mechanism by which support for users developing the TLA+ tools under nix should be added?

lemmy:
> In your view, what is the consensus mechanism by which support for users developing the TLA+ tools under nix should be added?

This has the support of both maintainers, and I don’t see a compelling argument in this PR for adding Nix beyond personal convenience. That rationale could easily justify adding a wide range of other development tools, which we should avoid.

lemmy:
This might be acceptable if there's a BIG FAT notice at the top of the file stating that this Nix file exists solely because some contributors find Nix useful. The notice should make it clear that:
a) Nix is unsupported,
b) the file may be removed at any time, and
c) nothing in the project should depend on Nix

Could this go into a `contrib/` folder?

PR #1207, opened 2025-06-26
SANY: parameterize parser for trace output

ahelwer:
Push ToolIO.out usage further up the stack

To provide greater control over SANY output, references to global static output streams like ToolIO.out should be replaced by class fields of type PrintStream. This can be done incrementally, pushing ToolIO.out references further up the stack by giving classes a PrintStream parameter and then providing ToolIO.out as a value for that parameter in the constructing class. Eventually, this process will concentrate ToolIO.out references near the top-level static main methods, thus unlocking the ability for users to control output with CLI parameters.

In furtherance of #1186


tlatools/org.lamport.tlatools/src/tla2sany/parser/TLAplusParser.java
```diff
@@ -7765,42 +7786,6 @@ final private boolean jj_2_74(int xla) {
     finally { jj_save(73, xla); }
   }
 
-  final private boolean jj_3R_210() {
```
lemmy, line 7768: This reformatting doesn't belong here.

ahelwer: All the jj methods are automatically generated by JavaCC

ahelwer:
Superseded by #1209

PR #1209, opened 2025-06-27
SANY: centralize log output streams

ahelwer:
Remove calls to static ToolIO.out and ToolIO.err streams

To facilitate better control over SANY's output, all instances where SANY outputs human-readable messages have been changed to go through a centralized interface called SanyOutput.

Often throughout SANY's codebase, output is written to ToolIO.out or or ToolIO.err, or to a random PrintStream error output parameter that happens to be in scope (even if the message is not an error). These have all been changed to now flow through the SanyOutput interface.

Three implementations of the SanyOutput interface are also provided: OutErrSanyOutput, which splits messages between output and error streams, SimpleSanyOutput, which uses a single stream, and SilentSanyOutput which silences all output. TLC can inject its own implementation of SanyOutput to get output to flow through its own output machinery.

Supersedes #1207, and very directly furthers #1186.


ahelwer:
One thing that needs to be specified a bit better here: the concept of "tool mode" for SANY. As-is, I have defined tool mode as putting all output to stderr. Tool mode is used from the toolbox, but not used from TLC (to enable TLC to insert its own streams for output control, ultimately implementing #1186). Do we want that? Is SANY tool mode even worthwhile to have? The only place this changes the behavior of SANY is in ParseUnit.java.

lemmy:
If "Tool Mode" means that SANY's existing textual output is merely enriched to make parsing easier, I would argue that this mode offers no real advantage over the easily parsable output provided by the XMLExporter. However, is the XMLExporter output currently a superset of SANY's command-line output—or could it be made into one?

ahelwer:
All right we can punt on the tool mode issue. ParseUnit will keep its dependency on the static ToolIO.getMode() method and maybe it can be picked up later in some kind of SanySettings type refactoring.

ahelwer:
With regard to XMLExporter, before these changes it did its best to suppress all SANY output since it was trying to write its main XML output to stdout as well (and having unstructured log messages co-mingled with the XML would be a problem). Now with the new `SanyOutput` class we can get the SANY methods it calls to only output if they encounter an error, in which case the XML will not be able to be produced anyway. Also, the user will enjoy an error message if they try to convert their spec to XML and it has a parse error.

lemmy:
I recommend including SANY's output in the structured XML generated by XMLExporter. This would eliminate the need for consumers of XMLExporter to separately read from stdout or stderr. Additionally, it would spare them from having to parse the unstructured output produced by SANY.

ahelwer:
That does sound like a good idea, although perhaps it can go in another PR.

ahelwer:
@lemmy this has been updated so SANY now takes an interface to which it writes its output; TLC can implement and inject its own implementation of the interface so log messages can be redirected through its own output machinery.

tlatools/org.lamport.tlatools/javacc/tla+.jj
```diff
@@ -90,18 +90,27 @@ PARSER_BEGIN(TLAplusParser)
...
 import tla2sany.utilities.Vector;
 import tlc2.output.EC;
 import util.Assert;
 import util.TLAConstants;
-import util.ToolIO;
 import util.UniqueString;
 
 public class TLAplusParser implements tla2sany.st.SyntaxTreeConstants, ParseTree {
 
+  private SanyOutput out = new SilentSanyOutput();
```
lemmy, line 109: Can this be final?

tlatools/org.lamport.tlatools/src/tla2sany/output/LogLevel.java
```diff
@@ -0,0 +1,60 @@
+/*******************************************************************************
```
lemmy, line 1: I'm not convinced that relying solely on log levels is sufficient. Instead, each message should have a unique identifier, allowing users to selectively suppress individual messages by their ID. That said, it's perfectly reasonable to group predefined sets of message IDs into categories such as Errors, Warnings, and so on.

tlatools/org.lamport.tlatools/javacc/tla+.jj
ahelwer: Unfortunately not, because there are several constructors generated by JavaCC; these changes include a new constructor we define which lets users supply a SanyOutput instance and then calls the generated constructors, but if users call those generated constructors directly then the out field needs to have some value.

tlatools/org.lamport.tlatools/src/tla2sany/output/LogLevel.java
ahelwer: I agree, and definitely view this as a transitional step toward unique message identifiers like TLC has.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -90,14 +105,6 @@ public class SANY {
    * level-checking, and returns a Specification object, which is the
    * root of the semantic graph of the specification.
    *
-   * It also sends progress indications and a text error stream
-   * suitable for printing on System.out or System.err to the
-   * PrintStream "syserr" (unless "syserr" is null).  The error stream
-   * is also saved in Errors objects that are components of the
-   * Specification object returned.  The caller may prefer to pass
-   * syserr = null, and use the these Errors objects to provide
-   * feedback to the human spec-writer in some other way.
```
Calvin-L, line 99: Parts of this comment should be preserved. The bits about `Errors` are still true (I think?) and can be left here. The bit about printing to `syserr` belongs on the new `frontEndMain` override above.

tlatools/org.lamport.tlatools/src/tla2sany/output/SanyOutput.java
```diff
@@ -0,0 +1,56 @@
...
+   * @param format A string into which args is interpolated.
+   * @param args Parameters to interpolate into the format string.
+   */
+  public void log(LogLevel level, String format, Object... args);
+
+  /**
+   * For the rare cases requiring an explicit stream associated with the
+   * given {@link LogLevel}, for example {@link Exception#printStackTrace(PrintStream)},
+   * retrieve a stream to which output can be printed.
+   *
```
Calvin-L, line 52: `PrintStream` is `Closeable`. Say whether callers must close, may close, or must _not_ close the returned object. (Looks like callers must _not_ close the returned object; it often points to some system-level print stream.)

tlatools/org.lamport.tlatools/src/tla2sany/output/SimpleSanyOutput.java
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * The stream receiving non-silenced output.
+   */
+  private final PrintStream out;
+
+  /**
+   * The null stream receiving output to silence.
+   */
+  private final PrintStream silent =
+      new PrintStream(OutputStream.nullOutputStream());
```
Calvin-L: Nit: this construction shows up a few times in this package. You could consider making this a package-private static constant and just referencing it from the various places it's needed.

Calvin-L (approved):

ahelwer:
@lemmy is this good to merge?

lemmy:
Thankfully, Olav caught the regression #1221 early, which made diagnosing it with bisection straightforward. That said, this should serve as a reminder that even seemingly simple refactorings carry a real risk of introducing regressions. The value of such changes must be weighed against that risk.

TLA+ users depend on TLC to verify systems that are often mission- or even safety-critical. If we make a mistake, the consequences could be severe — potentially even putting lives at risk.

ahelwer:
That is true, and we must also consider the risk of TLA+ sliding into irrelevance due to stagnation.

lemmy:
It seems this PR also introduced a regression in our VSCode extension: https://github.com/tlaplus/vscode-tlaplus/issues/433.

@ahelwer Could you please take a look when you have a moment? It would be great to get your input on resolving this. Thanks.

Issue #719, opened 2022-03-01
Replace `java.io.File` with a filesystem-independent representation


lemmy:
The TLA+ tools use `java.io.File` in (too) many places.  The assumption that every input is a file causes problems and makes it impossible to run in environments (such as sandboxes) that lack a file system. 

A few tla2tools.jar classes that rely on `java.io.File`:
- `NamedInputStream`
- `util.FileUtil`
- `FilenameToStream#resolve`
- TLAClass, SpecObj, SpecProcessor, ...

A representation should be chosen that:
- is independent of a file system
- allows multiple reads of the same TLA+ module/config (InputStream can only be read once!)

Related:
- https://github.com/tlaplus/tlaplus/issues/718
- https://github.com/tlaplus/tlaplus/issues/378
- https://github.com/tlaplus/tlaplus/issues/688
- Run TLC in a browser
- Dynamically (at runtime) parse additional TLA+ expressions

lemmy:
Given recent developments with MCP servers (see https://github.com/microsoft/genaiscript/issues/1809#issuecomment-3144927176), the need to decouple SANY/TLC from the filesystem has become more urgent. This change increases the priority of addressing this issue.

ahelwer:
I will think about the best way to tackle this.

PR #1210, opened 2025-07-01
SANY: replace custom Stack class with Deque

ahelwer:
Reduce maintenance burden by replacing our internal custom stack class with Deque (double-ended queue), the modern Java equivalent. We also get static type safety using generics. Although Java has a Stack class which more closely corresponds to our custom stack class, its usage is not recommended.

The functional changes are as follows:
- Replace all uses of Stack with a Deque<T> interface implemented by an ArrayDeque<T>
- Replace stack.push() with stack.addFirst()
- Replace stack.pop() with stack.removeFirst()
- Replace stack.peek() with stack.peekFirst()
- Replace stack.get(0) with stack.peekLast()
- Replace all iteration from top of stack to bottom with regular iteration over Deque

The Deque interface also does not support adding null elements, so one piece of code in TLAplusParser using null as a flag value has been rewritten to use the number of elements in the stack.

[SANY][Refactor]


ahelwer:
cc @Calvin-L, I think you will like these changes

tlatools/org.lamport.tlatools/src/tla2sany/utilities/Stack.java
```diff
@@ -1,39 +0,0 @@
...
-  }
-
-  public final Object push( Object obj) {
-    super.addElement( obj );
-    return obj;
-  }
-
-  public final int search( Object obj ){
-   for (int lvi = size-1; lvi >= 0; lvi-- )
-     if ( obj == info[ lvi ] ) return lvi;
```
Calvin-L, line 36: I'm all for removing this class (and the other redundant data structure classes we have lying around), but I am not 100% certain this class is 1:1 equivalent with `Deque`.

Actually, I'm 100% certain it is _not_ equivalent, because this `search()` method and the `contains()` method inherited from `tla2sany.utilities.Vector` both use reference equality instead of `.equals()`. See this comment on a similar PR from long ago:
https://github.com/tlaplus/tlaplus/pull/328#issuecomment-542160722

I've done a manual audit of all the types we store in a `Stack`. Here they are:
 - `Generator.pair`
 - `OpApplNode`
 - `UniqueString`
 - `Context`

For _these specific types_ it happens to be safe to interchange `equals()` and `==`. Because of that fact (and the good test coverage we have for SANY these days), I'll approve this PR.

----

But I will also point out that I think there is a better process for safely replacing these classes, sketched out toward the bottom of this monster comment:
https://github.com/tlaplus/tlaplus/pull/756#issuecomment-1231029602

(To summarize the monster comment, instead of deleting this class, I think this PR could be two very cautious commits:

 1. Add as many tests for this class as possible, especially covering the weird reference equality behavior.
 2. Change this class to `public class Stack<T> extends ArrayDeque<T>`, then override/implement methods as needed so that the tests from (1) still pass AND all the uses of this class can remain completely unchanged.

Follow-up commits can progressively inline methods, add type arguments, and generalize type signatures until `Stack` is completely gone---or until only its awkward differences from the standard library remain.)

ahelwer: That approach makes sense! I will do that when I eventually get around to refactoring out the custom `Vec` class.

lemmy: Such refactoring is a very low priority due to the associated risks and limited value (code deduplication) it provides.

ahelwer:
@Calvin-L I think this is good to merge unless @lemmy wants to review?

lemmy:
> @Calvin-L I think this is good to merge unless @lemmy wants to review?

I won’t be able to review this because the associated risks and limited value (code deduplication) outweigh the priority of this refactoring.

ahelwer:
Okay so to be clear, this is a personal prioritization thing and you are not vetoing a merge?

lemmy:
I'm not vetoing this PR, but I strongly oppose similar refactorings in the future. There are many other refactorings with a significantly higher return on investment than deduplicating a Collection class that hasn’t changed in 20 years. For example, refactoring File to InputStream would move us meaningfully closer to running TLC in the browser.

ahelwer:
I am not always the master of whatever thing in the codebase most irritates me in any given moment (in this case, it's that the contents of our custom Stack class was indecipherable in the Eclipse debugger) but it is good to know there is appetite for a File -> InputStream type refactoring. That will be covered by the new SANY API that has been in purgatory the past few months.

ahelwer:
This is good to merge @lemmy @muenchnerkindl

lemmy:
> [...] it is good to know there is appetite for a File -> InputStream type refactoring. That will be covered by the new SANY API that has been in purgatory the past few months.

https://github.com/tlaplus/tlaplus/issues/719

Issue #1212, opened 2025-07-17
TLCEval can silently fail due to how a constant operator is referenced


fhackett:
I have found a scenario where I cannot use `TLCEval` to force a constant expression to evaluate once. It seems related to how the variable is referenced.

Consider the following specification:
```tla
---- MODULE MustEvaluateDefnOnce ----
EXTENDS Integers, TLC

defn == TLCEval(Print("defn", { x : x \in 1 .. 5 }))
\* Works:
\* defn == TLCEval(Print("defn", 1 .. 5))

derivedDefn == TLCEval(Print("derivedDefn", { i \in 1 .. 5 : i + 1 \in defn }))

\* Same problem:
\* derivedDefn == TLCEval(Print("derivedDefn", { i \in defn : i + 1 \in defn }))

\* Works:
\* derivedDefn == TLCEval(Print("derivedDefn", { i \in defn : i + 1 <= 5 }))

VARIABLE x, y

Init ==
    /\ y = Print("y", defn)
    /\ x = Print("x", derivedDefn)

Next ==
    /\ 1 \in derivedDefn
    /\ UNCHANGED <<x, y>>

====
```

When checked (with the usual `INIT Init NEXT Next` config), this model should print exactly 1 line starting with `defn` and exactly 1 line starting with `derivedDefn`. Like this:
```
"defn" {1, 2, 3, 4, 5}
"y" {1, 2, 3, 4, 5}
"derivedDefn" {1, 2, 3, 4}
"x" {1, 2, 3, 4}
```

For `derivedDefn` this works as expected, but for `defn` the evaluation is never cached. Instead, it does this:
```
"defn" {1, 2, 3, 4, 5}
"y" {1, 2, 3, 4, 5}
"defn" {1, 2, 3, 4, 5}
"defn" {1, 2, 3, 4, 5}
"defn" {1, 2, 3, 4, 5}
"defn" {1, 2, 3, 4, 5}
"defn" {1, 2, 3, 4, 5}
"derivedDefn" {1, 2, 3, 4}
"x" {1, 2, 3, 4}
```

The problem is, `defn` is actually a simplification some non-trivial setup steps, involving loading and post-processing potentially large amounts of data using `IODeserialize`. None of the caching type operators seem to do anything.

I've tried my best to minimize the issue down to the simplest case where it still happens, and it apparently comes down to the example spec I've included.

My environment the VSCode extension with no special configuration, version 2025.7.41720.

[...]

PR #1214, opened 2025-07-23
Make TLCCache work in a constant context

fhackett:
It's not a fundamental fix for `TLCEval`, but this addresses #1212 for me. As described in the issue, I was unable to make TLC consistently cache a value. `TLCCache` only worked for variable-level formulas, and `TLCEval` silently fails depending on how the definition is referenced (e.g., it stops working if you refer to your constant from inside a set comprehension). Given that `TLCEval` is very hard to fix, I followed Markus's advice to pivot.

Instead of working on `TLCEval`, I extended `TLCCache` to work in constant contexts. This means, as long as you can make a sufficiently unique "closure" object that identifies your value, you can cache anything you want at the constant or variable levels (no applications for other levels come to mind yet).

The intended application is in combination with IOUtils, to ensure your datafile is not silently reloaded in a loop.

Inventory of changes:
- `TLCCache` has a 2nd code path, where values are cached in the provided `Tool`. I do this by adding a concurrent hash map, so that multiple workers can safely use this code path, and I don't have to worry about optimizing read and write concurrency management.
- A test that uses `TLCGet`/`TLCSet` to confirm a cached value is computed only once, from assumptions, to init, to next-state.

One thing I'm assuming: it's sufficient to call `.normalize()` on a value you're going to cache across workers, so that you can safely reference it across said workers without accidentally doing thread unsafe lazy evaluation. I think that's probably reasonable, from how I understand `.normalize()`, but do correct me if I'm wrong.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/TLCExt.tla
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/TLCExt.tla
@@ -168,8 +168,7 @@ TLCCache(expression, closure) ==
   (* Equals the value that an earlier evaluation of the expression evaluated to *)
   (* iff the closure of the earlier evaluation and this evaluation are equal.   *)
   (*                                                                            *)
-  (* Currently only implemented for state-level formulas.  See TLC!TLCEval for  *)
-  (* constant-level formulas.                                                   *)
+  (* Works in constant and variable-level contexts.                             *)
   (******************************************************************************)
   expression
 
--- a/tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
@@ -57,7 +57,6 @@ import tlc2.tool.impl.Tool;
 import tlc2.util.Context;
 import tlc2.util.FP64;
 import tlc2.util.IdThread;
-import tlc2.value.IValue;
 import tlc2.value.Values;
 import tlc2.value.impl.BoolValue;
 import tlc2.value.impl.CounterExample;
@@ -336,7 +335,9 @@ public class TLCExt {
 
 		if (expr.getLevel() == LevelConstants.ConstantLevel) {
 			final Value key = tool.eval(closure, c, s0, s1, control, cm).normalize();
-			return tool.getOrSetCached(key, (final Value _ignored) -> tool.eval(expr, c, s0, s1, control, cm).normalize());
+			return tool.getOrSetCached(key, () -> {
+				return tool.eval(expr, c, s0, s1, control, cm).normalize();
+			});
 		} else if ( expr.getLevel() == LevelConstants.VariableLevel) {
 			final int key = expr.hashCode() ^ closure.hashCode() ^ tool.eval(closure, c, s0).hashCode();
 
--- a/tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
@@ -57,6 +57,7 @@ import tlc2.tool.impl.Tool;
 import tlc2.util.Context;
 import tlc2.util.FP64;
 import tlc2.util.IdThread;
+import tlc2.value.IValue;
 import tlc2.value.Values;
 import tlc2.value.impl.BoolValue;
 import tlc2.value.impl.CounterExample;
@@ -333,8 +334,10 @@ public class TLCExt {
 		final ExprOrOpArgNode expr = args[0];
 		final ExprOrOpArgNode closure = args[1];
 
-		if (expr.getLevel() == LevelConstants.VariableLevel) {
-
+		if (expr.getLevel() == LevelConstants.ConstantLevel) {
+			final Value key = tool.eval(closure, c, s0, s1, control, cm);
+			return tool.getOrSetCached(key, (final Value _ignored) -> tool.eval(expr, c, s0, s1, control, cm));
+		} else if ( expr.getLevel() == LevelConstants.VariableLevel) {
 			final int key = expr.hashCode() ^ closure.hashCode() ^ tool.eval(closure, c, s0).hashCode();
 
 			final Value value = s0.getCached(key);
@@ -345,7 +348,7 @@ public class TLCExt {
 			return s0.setCached(key, tool.eval(expr, c, s0, s1, control, cm));
 		}
 
-		// For constant, action or temporal formulas, all we do is to evaluate the TLA+
+		// For action or temporal formulas, all we do is to evaluate the TLA+
 		// expression--nothing is cached!
 		return null; // Returning null here causes Tool.java to evaluate the original expression.
 	}
--- a/tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
@@ -335,8 +335,8 @@ public class TLCExt {
 		final ExprOrOpArgNode closure = args[1];
 
 		if (expr.getLevel() == LevelConstants.ConstantLevel) {
-			final Value key = tool.eval(closure, c, s0, s1, control, cm);
-			return tool.getOrSetCached(key, (final Value _ignored) -> tool.eval(expr, c, s0, s1, control, cm));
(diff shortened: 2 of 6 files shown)


tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
```diff
@@ -333,8 +333,12 @@ public static Value tlcEval2(final Tool tool, final ExprOrOpArgNode[] args, fina
 		final ExprOrOpArgNode expr = args[0];
 		final ExprOrOpArgNode closure = args[1];
 
-		if (expr.getLevel() == LevelConstants.VariableLevel) {
-
+		if (expr.getLevel() == LevelConstants.ConstantLevel) {
+			final Value key = tool.eval(closure, c, s0, s1, control, cm).normalize();
+			return tool.getOrSetCached(key, () -> {
```
lemmy: Have you considered `tla2sany.semantic.SemanticNode.setToolObject(int, Object)` instead of introducing a Map in Tool?

fhackett: No I hadn't - I tried to understand how that API works, and it seems it allows storing run specific data in some AST node, if you have a unique ID for it.

In principle I can imagine storing the cache as such a tool object (I can also see how to allocate a specific ID), but I'm confused - maybe I misunderstood how thread safety is done here. Are tool objects / ASTs isolated per thread / worker? If not, those tool objects are unprotected array manipulations, and given TLCCache would get called for the first time at any point, I couldn't guarantee to not do bad things when initializing the cache.

Related: I tried some more testing "live" in TraceLink, and found (1) a bug in this PR, that somehow the field can still be null, and (2) some perf questions, where the concurrent map might be slowing somethings down. Working on those regardless.

lemmy: Yes, you would need to introduce explicit locking in `TLCCache`, similar to how it’s done in `TLCEval.java`.  The `ReaderWriterLock` should provide sufficient concurrency.

fhackett: I was part way through writing a message explaining why I thought that was unsafe, but I changed my mind.
My original comment was about how if TLCEval and TLCCache both have their own lock that protects the array underlying `setToolObject` , then surely you have 2 locks protecting 1 thing, and that will lead to trouble.

I actually think it's fine, but the reasoning is subtle. Any of these private locks protect specifically the AST "under" the `TLCEval(...)` of `TLCCache(...)` call. If I have the nonsensical but possible combination `TLCCache(TLCEval(???), {})`, and both caching mechanisms activate at the same time, we are still protecting 2 arrays with 2 locks, relatively speaking, because we can never be sharing the same AST node.

In that case, I think I get it. I pushed an alternative version that fixes a bug I found and also follows your suggestion.

tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/TLCExt.tla
```diff
@@ -168,8 +168,7 @@ TLCCache(expression, closure) ==
   (* Equals the value that an earlier evaluation of the expression evaluated to *)
   (* iff the closure of the earlier evaluation and this evaluation are equal.   *)
   (*                                                                            *)
-  (* Currently only implemented for state-level formulas.  See TLC!TLCEval for  *)
-  (* constant-level formulas.                                                   *)
+  (* Works in constant and variable-level contexts.                             *)
```
lemmy: "state-level" is more commonly used compared to "variable-level" .

tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
```diff
@@ -326,15 +328,64 @@ public static synchronized Value tlcModelValue(final Value val) {
 		return ModelValue.add(str.val.toString());
 	}
 	
+	private static final StampedLock tlcEval2Lock = new StampedLock();
```
lemmy: Could you explain why a stamped instead of a ReaderWriteLock?

fhackett: I read about its read locking (by far the most common case here) being faster than ReaderWriterLock, something about recursive locking support / the cache effect of timestamps. But also, it looked interesting and I wanted to try it.

While I don't have a strong opinion, I am now a little curious if the difference is measurable at all between the two locking methods. I have some specs that make extensive use of `TLCCache`, so there is some merit in empirically checking if this matters. I'll reply with results when I get back to this.

lemmy: The performance difference has to be noticeable to justify the additional complexity arounds `StampLock`'s API compared to `ReaderWriterLock`.

tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
```diff
@@ -326,15 +328,64 @@ public static synchronized Value tlcModelValue(final Value val) {
...
+				}
+				if (value != null) {
+					return value;
+				}
+				
+				tlcEval2Lock.unlock(stamp);
+				stamp = 0;
+				
+				// Note: do this outside of the write lock, otherwise recursive TLCCache invocations that miss
+				// the cache will deadlock trying to write to the cache while an outer call holds the lock.
```
lemmy: Reentrancy wouldn't be an issue with `ReentrantReadWriteLock`.

tlatools/org.lamport.tlatools/test-model/ConstantContextTLCCache.tla
```diff
@@ -0,0 +1,29 @@
+---- MODULE ConstantContextTLCCache ----
+EXTENDS Integers, TLC, TLCExt
+
+COUNT_ID == 1
+
+ASSUME TLCSet(COUNT_ID, 0)
```
lemmy, line 10: Consider adding a couple of lines to explain what's going on in this test.

tlatools/org.lamport.tlatools/test/tlc2/module/ConstantContextTLCCacheTest.java
```diff
@@ -0,0 +1,23 @@
+package tlc2.module;
```
lemmy, line 26: Please add the usual MIT license header.

tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
fhackett: I checked. In a 5 minute 40 seconds run that calls `TLCCache` constantly, no difference is measurable. I changed it to the simpler lock.

lemmy (approved):
LGTM

Will merge soon unless @Calvin-L and/or @ahelwer also want to review.

tlatools/org.lamport.tlatools/src/tlc2/module/TLCExt.java
```diff
@@ -326,15 +328,70 @@ public static synchronized Value tlcModelValue(final Value val) {
...
+				}
+				
+				// The element is still missing. Populate it (may recursively read and/or write lock)..
+				value = tool.eval(expr, c, s0, s1, control, cm);
+				// This is needed to ensure the value is usable from any thread going forward.
+				value.initialize();
+				
+				cache.put(key, value);
+				return value;
+			} finally {
```
lemmy, line 386: I would hope for TLC to crash if anything goes south in between try/finally.  At least, `TLCEval` makes that assumption, and, thus, doesn't use try/finally.

fhackett: I see. My main purpose for this is actually just code convenience, not necessarily exception safety. I wanted to use something like a try-with-resources, but locks don't support those. So, to avoid having to remember the unlock cases at each return, I just crammed them all in a finally block.

I can find a different structure if you like, but it might be harder to read. Or maybe this was just informational, in which case sure, that makes sense.

fhackett: Actually I'm more confused after checking the code - `TLCEval` does also use try/finally https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/src/tlc2/module/TLCEval.java#L114

I didn't think I entirely made up the pattern... anyhow, let me know if you recommend changing something, and/or what you meant to say. I might be missing context.

PR #1215, opened 2025-07-23
SANY: Start to sort out syntax parser error reporting

ahelwer:
Emit PlusCal divergence warnings through Errors class

Currently, error reporting in SANY's first stage (the syntax parser) is a bit of a mess. The current state of things:
 1. The TLAplusParser class throws ParseException instances, which are caught and turned into AbortException instances, which are caught and turned back into ParseException instances. The error message is output to the log but loses its attachment to the exceptions along the way.
 2. PlusCal translation divergence warnings are output directly to the log instead of tagged with a standardized error code and stored in the Errors class like other anomalous occurrences.
 3. Warning and error messages stored in the Errors class are both output at the Errors log level, instead of separated appropriately.

These changes currently do not really touch the first issue, but handle the second and third. Another error message is also tagged with a best-effort location.

[SANY][Refactor]


tlatools/org.lamport.tlatools/src/tla2sany/modanalyzer/SpecObj.java
```diff
@@ -272,12 +271,20 @@ protected ParseUnit findOrCreateParsedUnit(String name, Errors errors, boolean f
                 * produces this error.  Hopefully, we can use this to attach a     *
                 * location to the error message.                                   *
                 *******************************************************************/
-                throw errors.addError(
-                    ErrorCode.MODULE_FILE_CANNOT_BE_FOUND,
-                    Location.nullLoc,
-                    "Cannot find source file for module " + name +
-                    ((nextExtenderOrInstancerModule == null) ? "" : " imported in module " + nextExtenderOrInstancerModule.getName()) + "."
-                );
+                if (nextExtenderOrInstancerModule == null) {
```
ahelwer, line 276: I am skeptical that this branch can actually be hit. I was playing around trying to trigger it but was unable to engineer a scenario where a module fails to be imported from nowhere in particular.

lemmy:
As I recall, the VSCode extension is unaware of SANY’s divergence warnings, but the Toolbox handles them.

ahelwer:
The scope of this PR was narrowed so downstream consumers won't need to update anything.

ahelwer:
>On all four, either Calvin or I have left comments explaining why the PR has not yet been accepted.

@lemmy ?

tlatools/org.lamport.tlatools/src/tla2sany/modanalyzer/SpecObj.java
```diff
@@ -871,8 +878,8 @@ private void resolveNamesBetweenModuleAndExtention(ModulePointer extenderModule,
     /**
      * This invokes {@code loadSpec(rootExternalModuleName, errors, false);}
      */
-    public boolean loadSpec(final String rootExternalModuleName, final Errors errors, SanyOutput out) throws AbortException {
-    	return loadSpec(rootExternalModuleName, errors, false, out);
+    public void loadSpec(final String rootExternalModuleName, final Errors errors, SanyOutput out) throws AbortException {
```
FedericoPonzi: Because the method is public, changing the return type is a breaking change :( if some consumer is doing anything with this boolean, their code won't compile anymore.

tlatools/org.lamport.tlatools/src/tla2sany/modanalyzer/SpecObj.java
```diff
@@ -885,7 +892,7 @@ public boolean loadSpec(final String rootExternalModuleName, final Errors errors
 	 *                           followed by a validation of them if they exist
 	 * @see Validator
 	 */
-	public boolean loadSpec(final String rootExternalModuleName, final Errors errors, final boolean validateParseUnits, SanyOutput out)
+	public void loadSpec(final String rootExternalModuleName, final Errors errors, final boolean validateParseUnits, SanyOutput out)
```
FedericoPonzi: Same here, revert if not needed.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/ErrorCode.java
```diff
@@ -165,7 +165,10 @@ public enum ErrorCode {
    */
   EXTENDED_MODULES_SYMBOL_UNIFICATION_AMBIGUITY (4800, ErrorLevel.WARNING),
   INSTANCED_MODULES_SYMBOL_UNIFICATION_AMBIGUITY (4801, ErrorLevel.WARNING),
-  RECORD_CONSTRUCTOR_FIELD_NAME_CLASH (4802, ErrorLevel.WARNING);
+  RECORD_CONSTRUCTOR_FIELD_NAME_CLASH (4802, ErrorLevel.WARNING),
+  PLUSCAL_ALGORITHM_AND_TRANSLATION_CHANGED_SINCE_LAST_TRANSLATION (4803, ErrorLevel.WARNING),
+  PLUSCAL_ALGORITHM_CHANGED_SINCE_LAST_TRANSLATION (4804, ErrorLevel.WARNING),
+  PLUSCAL_TRANSLATION_CHANGED_SINCE_LAST_TRANSLATION (4805, ErrorLevel.WARNING);
```
FedericoPonzi, line 171: do we later need to update https://github.com/tlaplus/vscode-tlaplus/blob/42f604eeeb6f4eb463cb5bc5496f33d7a542d755/src/parsers/tlcCodes.ts as well?

ahelwer: No, these error-specific IDs are not yet exposed or consumed publicly.

ahelwer:
Thanks for the review, @FedericoPonzi! I tested this build in the VS Code extension using the "Parse Module" command, but unfortunately it just automatically re-translates the PlusCal before parsing so the warning isn't even triggered (looking at the SANY output pane). Maybe I'm just doing it wrong but I can't ever get the Parse Module command to actually produce underlined errors in any TLA+ file, no matter what error it produces.

FedericoPonzi:
I think this is the usual behaviour with syntax errors on pluscal:
Screenshot 2026-01-17 22:02:33

ahelwer:
Ah okay, I mean just in general TLA+ files, not PlusCal. Similar to https://github.com/tlaplus/vscode-tlaplus/issues/433 I can't get error highlighting working in my copy of VS Code. This happens even with the built-in tools. The error appears in the SANY output pane but nothing is highlighted.

ahelwer:
Ah I figured out the issue. It has to do with symlinks. I will open an issue against the VS Code extension.

ahelwer:
Okay after working around https://github.com/tlaplus/vscode-tlaplus/issues/480 I can confirm this PR's build works just fine in the VS Code extension. Thanks for the review, @FedericoPonzi!

PR #1218, opened 2025-07-26
WIP

just-now:
[WIP] Work in progress. Additional input is needed.

> Review and address remaining TODOs
> Map.of("_DumpTraceFilePrefix", String.format("%s_", numOfGenTraces.get())))

Is it really needed to have `DumpTraceFilePrefix` here? 
Simulator produces one trace per run only. The user of TLC might parametrize just file name to distinguish traces between runs? WDYT?

> Ensure errorCode is handled correctly in all cases (Simulator.java, SingleThreadedSimulator.java)

Added one more place in the code where, presumably, `errorCode` needs to be updated.

> Add tests for the new functionality

Started writing tests. Faced with the following exception in the simulator:

```
expr == line 13, col 5 to line 15, col 87 of module _JsonTrace
pop == TLCStackFrame [node=line 9, col 1 to line 9, col 17 of module Github1217]
```

```
 [junit] java.lang.AssertionError
    [junit]     at tlc2.debug.TLCDebugger.popFrame(TLCDebugger.java:781)
    [junit]     at tlc2.debug.TLCDebugger.popFrame(TLCDebugger.java:787)
    [junit]     at tlc2.tool.impl.DebugTool.constLevelEval(DebugTool.java:356)
    [junit]     at tlc2.tool.impl.DebugTool.evalImpl(DebugTool.java:230)
    [junit]     at tlc2.tool.impl.DebugTool.eval(DebugTool.java:192)
    [junit]     at tlc2.tool.impl.Tool.isValid(Tool.java:3513)
    [junit]     at tlc2.tool.impl.DebugTool.isValid(DebugTool.java:145)
    [junit]     at tlc2.tool.impl.Tool.checkPostConditionWithContext(Tool.java:3578)
    [junit]     at tlc2.tool.impl.Tool.checkPostConditionWithCounterExample(Tool.java:3569)
    [junit]     at tlc2.tool.Simulator.simulate(Simulator.java:404)
    [junit]     at tlc2.tool.Simulator.simulate(Simulator.java:325)
    [junit]     at tlc2.TLC.process(TLC.java:1252)
```

It looks like the last frame doesn't contain `_JsonTrace` action. Any idea how to fix it?

> Write documentation (maybe?! yes please!)

Let's discuss what documentation is needed.
`-invlevel` is already documented, maybe it's worth having an example how to use this command option. If so, please tell me what file shall I add it into. In case if there's something else I can do with the documentation related to `-invlevel`, please tell me.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/Simulator.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/Simulator.java
@@ -358,6 +358,8 @@ public class Simulator {
 
 			// If the result is an error, print it.
 			if (result.workerId() == -1) {
+				// This might happen when thread pool fails to spawn a thread.
+				errorCode = EC.GENERAL;
 				runningWorkers.clear();
 				break;
 			} else if (result.isError()) {
@@ -399,9 +401,8 @@ public class Simulator {
 				// see tlc2.tool.Worker.doPostCheckAssumption()
 				if (result.error().hasTrace()) {
 					error.errorCode = Math
-							//TODO numOfGenTraces is not ideal because it is not monotonically increasing.
 							.max(this.tool.checkPostConditionWithCounterExample(result.error().getCounterExample(),
-									Map.of("_DumpTraceFilePrefix", String.format("%s_", numOfGenTraces.get()))), error.errorCode);
+									Map.of("_DumpTraceFilePrefix", String.format(""))), error.errorCode);
 				} else {
 					error.errorCode = Math.max(this.tool.checkPostCondition(), error.errorCode);
 				}
--- a/tlatools/org.lamport.tlatools/test-model/Github1217.cfg
+++ b/tlatools/org.lamport.tlatools/test-model/Github1217.cfg
@@ -1,0 +1,1 @@
+SPECIFICATION Spec
--- a/tlatools/org.lamport.tlatools/test-model/Github1217.tla
+++ b/tlatools/org.lamport.tlatools/test-model/Github1217.tla
@@ -1,0 +1,20 @@
+------ MODULE Github1217 ------
+EXTENDS Integers
+
+VARIABLE x
+
+Init ==
+    x = 0
+
+Add == x' = x + 1
+Sub == x' = x - 1
+Reset == x' = 0
+
+Next ==
+    \/ Add
+    \/ Sub
+    \/ Reset
+
+Spec == Init /\ [][Next]_<<x>>
+---------------------------
+=====
--- a/tlatools/org.lamport.tlatools/test/tlc2/tool/Github1217Test.java
+++ b/tlatools/org.lamport.tlatools/test/tlc2/tool/Github1217Test.java
@@ -1,0 +1,88 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Microsoft Research. All rights reserved.
+ *
+ * The MIT License (MIT)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+ * of the Software, and to permit persons to whom the Software is furnished to do
+ * so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN
+ * AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * Contributors:
+ *   Anatoliy Bilenko - initial implementation
+ ******************************************************************************/
+package tlc2.tool;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertEquals;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.Files;
+
+import org.junit.Test;
+
+import tlc2.output.EC;
+import tlc2.tool.liveness.ModelCheckerTestCase;
+import util.FileUtil;
+
+
+public class Github1217Test extends ModelCheckerTestCase {
+
+	private static final Path dumpFilePath = Paths.get(System.getProperty("java.io.tmpdir"), "Github1217.json");
+
+	public Github1217Test() {
+		super("Github766", new String[] { "-note", "-config", "Github1217.cfg", "-simulate", "-invlevel", "10",
+						  "-dumpTrace", "json", "Github1217.json", "Github1217.tla"},
+		EC.ExitStatus.VIOLATION_SAFETY);
+	}
+
+  // java -cp dist/tla2tools.jar tlc2.TLC -note -simulate -invlevel 10 -dumptrace json out.json test-model/Github1217.tla
+
+
+	@Test
+	public void testSpec() throws IOException {
+
+		assertTrue(recorder.recorded(EC.TLC_FINISHED));
+		// assertTrue(recorder.recordedWithStringValues(EC.TLC_STATS, "51", "50", "47"));
+		// assertTrue(recorder.recordedWithStringValue(EC.TLC_SEARCH_DEPTH, "3"));
+
+		// Assert POSTCONDITION.
+		//assertFalse(recorder.recorded(EC.TLC_ASSUMPTION_FALSE));
+		//assertFalse(recorder.recorded(EC.TLC_ASSUMPTION_EVALUATION_ERROR));
+
+		//assertTrue(Files.exists(Github1217Test.dumpFilePath));
+		//System.out.println(Github1217Test.dumpFilePath);
+		//assertEquals(Github1217Test.dumpFilePath.toString(), 0.0, 1.0);
+
+		// If the file exist, simply compare it to a correct and manually checked version.
+		// try (final InputStream expected = getClass().getResourceAsStream("Github1147.dot");
+		// 		final FileInputStream actual = new FileInputStream(Github1147Test.dumpFilePath.toFile());) {
+		// 	BufferedReader expectedReader = new BufferedReader(new InputStreamReader(expected));
+		// 	BufferedReader actualReader = new BufferedReader(new InputStreamReader(actual));
+		// 	while (expectedReader.ready() && actualReader.ready()) {
+		// 		String expectedLine = expectedReader.readLine();
+		// 		String actualLine = actualReader.readLine();
+		// 		assertEquals(expectedLine, actualLine);
+		// 	}
+
+		// 	assertEquals(expectedReader.ready(), actualReader.ready());
+		// }
+
+		// assertZeroUncovered();
+	}
+}


lemmy:
> [WIP] Work in progress. Additional input is needed.
> 
> > Review and address remaining TODOs
> > Map.of("_DumpTraceFilePrefix", String.format("%s_", numOfGenTraces.get())))
> 
> Is it really needed to have `DumpTraceFilePrefix` here? Simulator produces one trace per run only. The user of TLC might parametrize just file name to distinguish traces between runs? WDYT?

Remember the `-continue` parameter. If it’s present, the simulator will create multiple files.

> [...]
> 
> > Add tests for the new functionality
> 
> Started writing tests. Faced with the following exception in the simulator:
> 
> ```
> expr == line 13, col 5 to line 15, col 87 of module _JsonTrace
> pop == TLCStackFrame [node=line 9, col 1 to line 9, col 17 of module Github1217]
> ```
> 
> ```
>  [junit] java.lang.AssertionError
>     [junit]     at tlc2.debug.TLCDebugger.popFrame(TLCDebugger.java:781)
>     [junit]     at tlc2.debug.TLCDebugger.popFrame(TLCDebugger.java:787)
>     [junit]     at tlc2.tool.impl.DebugTool.constLevelEval(DebugTool.java:356)
>     [junit]     at tlc2.tool.impl.DebugTool.evalImpl(DebugTool.java:230)
>     [junit]     at tlc2.tool.impl.DebugTool.eval(DebugTool.java:192)
>     [junit]     at tlc2.tool.impl.Tool.isValid(Tool.java:3513)
>     [junit]     at tlc2.tool.impl.DebugTool.isValid(DebugTool.java:145)
>     [junit]     at tlc2.tool.impl.Tool.checkPostConditionWithContext(Tool.java:3578)
>     [junit]     at tlc2.tool.impl.Tool.checkPostConditionWithCounterExample(Tool.java:3569)
>     [junit]     at tlc2.tool.Simulator.simulate(Simulator.java:404)
>     [junit]     at tlc2.tool.Simulator.simulate(Simulator.java:325)
>     [junit]     at tlc2.TLC.process(TLC.java:1252)
> ```
> 
> It looks like the last frame doesn't contain `_JsonTrace` action. Any idea how to fix it?

You can ignore this by overriding `runWithDebugger` in your unit test. For instance, this is also done in https://github.com/tlaplus/tlaplus/blob/6479e53ae1e418a1bfe31677ad45012fe8ed7008/tlatools/org.lamport.tlatools/test/tlc2/tool/Github971eTest.java#L60-L63

> > Write documentation (maybe?! yes please!)
> 
> Let's discuss what documentation is needed. `-invlevel` is already documented, maybe it's worth having an example how to use this command option. If so, please tell me what file shall I add it into. In case if there's something else I can do with the documentation related to `-invlevel`, please tell me.

Documenting `-invlevel` would be beneficial, but it’s not within the scope of this PR. Instead, we should document how the behavior of `POSTCONDITION` in the context of simulation changed. Specifically, we should explain that `POSTCONDITION` is now evaluated multiple times when `-continue` is present.

just-now:
It took some time to return back to this PR. 

I'll write down the documentation for `POSTCONDITION` changes and `-continue` tomorrow.

tlatools/org.lamport.tlatools/src/tlc2/tool/Simulator.java
```diff
@@ -399,9 +401,8 @@ protected int simulate(final StateVec initStates) throws InterruptedException {
 				// see tlc2.tool.Worker.doPostCheckAssumption()
 				if (result.error().hasTrace()) {
 					error.errorCode = Math
-							//TODO numOfGenTraces is not ideal because it is not monotonically increasing.
 							.max(this.tool.checkPostConditionWithCounterExample(result.error().getCounterExample(),
-									Map.of("_DumpTraceFilePrefix", String.format("%s_", numOfGenTraces.get()))), error.errorCode);
+									Map.of("_DumpTraceFilePrefix", String.format("%s_", System.currentTimeMillis()))), error.errorCode);
```
lemmy: The `currentTimeMillis` causes the file names to be unpredictable. While I don’t necessarily suggest that it should be `numOfGenTraces.get`, `numOfGenTraces.get` does have the advantage of allowing you to deduce the name of the dump file from `TLCGet(“stats”)`.

just-now: OK. I'll return previous implementation back.

just-now:
- I've updated the documentation.

- Also, I have a question:

In case if `-dumpTrace` is used with in the following way  `-dumpTrace json /tmp/xyz.json` with a path with `/`s adding a prefix with `_DumpTraceFilePrefix \o _JsonTraceFile` gives the following `%s_/tmp/xyz.json` format string. I presume this is not something we want. 

What's the best way to fix it?

I can implement a formula `PathJoin` and call it the following way `JsonSerialize(PathJoin(_DumpTraceFilePrefix, _JsonTraceFile), CounterExample)`.

```tla
PathJoin(prefix, file) ==
  LET normPrefix ==
        IF StringLen(prefix) > 0 /\ SubSeq(prefix, StringLen(prefix), StringLen(prefix)) = "/"
        THEN SubSeq(prefix, 1, StringLen(prefix) - 1)
        ELSE prefix
      normFile ==
        IF StringLen(file) > 0 /\ SubSeq(file, 1, 1) = "/"
        THEN SubSeq(file, 2, StringLen(file))
        ELSE file
  IN normPrefix \o "/" \o normFile
  ```
  
  or we can do something else.

**Comment by lemmy** [comment]
Careful, the proposed PathJoin implementation would be platform-dependent due to the use of `/`. Instead of parsing and splitting a single command-line argument, consider adding a second, optional command-line parameter to specify the output path.

File: tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_JsonTrace.tla [hunk 2]
```diff
@@ -5,13 +5,14 @@ LOCAL INSTANCE Json
...
+LOCAL _JsonTraceFileDirectory == ""
 
 LOCAL _JsonTraceFile ==
     "CounterExample.json"
 
 LOCAL _JsonTrace ==
     IF CounterExample.state = {} THEN TRUE ELSE
-        /\ JsonSerialize(_DumpTraceFilePrefix \o _JsonTraceFile, CounterExample)
-        /\ PrintT("CounterExample written: " \o _DumpTraceFilePrefix \o _JsonTraceFile)
+        /\ JsonSerialize(_JsonTraceFileDirectory \o _DumpTraceFilePrefix \o _JsonTraceFile, CounterExample)
```
lemmy: It would be useful to allow specifying an output directory for all dumptrace export formats. I propose refactoring the existing `jsondir` option into something more general, such as `dumpdir`, and making it available across all formats.

just-now:
I refactored current implementation and if it'll be considered optimal, I will update all other TLA+ modules related to dumps from the list:
 - `_TLAPlusCounterExample`,
 - `_TLCTrace`,
 - `_TLCTracePlain`,
 - `_TLCActionTrace`,
 - `_DotTrace`.

Also, I will add a documentation for `-dumpDir`.

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
```diff
@@ -72,10 +72,10 @@
 
 /**
  * Main TLC starter class.
- * 
+ *
```
lemmy: Please minimize formatting changes.

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
```diff
@@ -655,6 +658,14 @@ else if (index < args.length)
                     printErrorMsg("Error: A file name for dumping states required.");
                     return false;
                 }
+            } else if (args[index].equalsIgnoreCase("-dumpDir")) {
```
lemmy: I'd have assumed `dumpDir` to be a subcommand/parameter of `dumptrace`.

just-now: Ah. OK.

just-now: Done.

just-now:
Please tell me if it's a good time to return back to this [comment](https://github.com/tlaplus/tlaplus/pull/1218#issuecomment-3211117056)

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
```diff
@@ -657,46 +657,41 @@ else if (index < args.length)
...
 					} else {
 						printErrorMsg("Error: Unknown format " + fmt + " given to -dumpTrace.");
 						return false;
 					}
+                    pcs.add(pc);
+
+                    // Parse the second optional parameter
+                    if ((index + 2) < args.length) {
+                        if ("dumpdir".equalsIgnoreCase(args[index])) {
+                            index++; // consume "-dumpdir".
```
lemmy: Dangling dash, i.e., `-dumpdir` vs `dumpdir` in comment.

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
```diff
@@ -1588,7 +1586,7 @@ private void printUsage()
 														"show only the differences between successive states when\n"
 															+ "printing trace information; defaults to printing\n"
 															+ "full state descriptions", true));
-    	sharedArguments.add(new UsageGenerator.Argument("-dumpTrace", "format file",
+    	sharedArguments.add(new UsageGenerator.Argument("-dumpTrace", "format file [dumpdir dir]",
```
lemmy: Reading this makes me wonder if we should abandon `dumpdir` in favor of positional parameters for consistency. In that case, the command would be `format file dir`. For instance, `json Trace.json /path/to/dir`. The only downside is that one would have to specify a file even if they would be content with the default.

lemmy:
> I refactored current implementation and if it'll be considered optimal, I will update all other TLA+ modules related to dumps from the list:
> 
> * `_TLAPlusCounterExample`,
> * `_TLCTrace`,
> * `_TLCTracePlain`,
> * `_TLCActionTrace`,
> * `_DotTrace`.

I’ve added new comments, but they’re unrelated to refactoring the other modules. Please proceed with refactoring the other modules.

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
just-now: What about having positional parameters in the path, specifying their placeholders in the path-string itself? 
In `handleParameters()` we can check expected number of positional `%d`-parameters. It could be so that there's not positional parameters at all.

Example: `json /tmp/%d_xyz.json`.

lemmy: Good idea!

just-now: OK, let me try to draft it.

just-now: > Example: json /tmp/%d_xyz.json.

I implemented passing the path to dumpfile with a format string in it. I'm not sure that something I did is not an overengineering.

lemmy:
@just-now Please force push your two commits with signed-off commits.

```
There are 2 commits incorrectly signed off. This means that the author(s) of these commits failed to include a Signed-off-by line in their commit message.

To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).
```

lemmy:
@just-now The test failures might be caused by unrelated changes.  Consider rebasing this branch on top of HEAD/master.

tlatools/org.lamport.tlatools/src/tlc2/tool/Simulator.java
```diff
@@ -358,6 +360,8 @@ protected int simulate(final StateVec initStates) throws InterruptedException {
 
 			// If the result is an error, print it.
 			if (result.workerId() == -1) {
+				// This might happen when thread pool fails to spawn a thread.
+				errorCode = EC.GENERAL;
```
lemmy, line 368: This change doesn't seem related to `-dumptrace`.  Could you please move this into a separate PR?

tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_JsonTrace.tla
```diff
@@ -4,14 +4,11 @@ LOCAL INSTANCE TLCExt
 LOCAL INSTANCE Json
 LOCAL INSTANCE Sequences
 
-LOCAL _DumpTraceFilePrefix == ""
-
-LOCAL _JsonTraceFile ==
-    "CounterExample.json"
+LOCAL _DumpTraceFile == ""
```
lemmy: This change is going to collide with https://github.com/tlaplus/tlaplus/pull/1297.  Moreover, there is no "convention" for the name for the outfile CONSTANT in each dumptrace format.  Thus, this would only work for _JsonTrace.

PR #1222, opened 2025-08-01
SANY bugfix: syntax parser errors were incorrectly silenced

### LINKED ISSUE #1221: Parser errors not showing details in 1.8
## Description
I was trying out TLA+ using the new 1.8 release following the [tutorial](https://learntla.com/core/operators.html) on the website. In one of the first steps, the tutorial mentions a specific parse error that may happen if you write = instead of == for a definition, which I wanted to try out. However, when I tried that on my end, it did not give me any parsing details. Since we are planning on using TLA+ for a course soon, so detailed parser errors are appreciated so that student don't get stuck with them too much.

## Steps to Reproduce
1. Open the TLA+ Toolbox.
2. Create a new empty specification `test.tla`.
3. In this specification, write `x = 0`.
4. Save the specification.

## Expected Behavior
In the Parsing Errors sidebar on the right, a parsing error appears with detailed information of the form:
```
In module test
Was expecting “=====” or more Module Body, encountered "$Operator" at line $n…
Could not parse module test from file test.tla
```

## Actual Behavior
In the Parsing Errors sidebar on the right, a parsing error appears the following contents:
```
In module test

Could not parse module test from file test.tla
```

## Steps Taken to Fix
I have tried different grammatical mistakes, such as using == for equality and writing multiple lines of EXTENDS to see whether these do work properly, but they give the same parser warning without any details. I have also tried to use the previous release, namely 1.7.4, and there it does give detailed parser warnings.

## Possible Fix
No idea, there is probably something internally going wrong with moving parsing errors to the UI.

## Your Environment
 - Toolbox version: 1.8
 - Operating System: Windows 11
Actually, looking at the version numbers and dates in the "About" menu, which are all at most April 2020, I seems that the 1.8 release might actually contain a 5 year old version of the tool, but the release files are only 5 days old according to Github...

ahelwer:
As part of PR #1209, SANY output was centralized into a single logging class. These changes extended to the TLAplusParser class, which was given an additional constructor accepting a SanyOutput parameter. Unfortunately the existing constructors were auto-generated by JavaCC so could not be removed, and the code constructing TLAplusParser in ParseUnit.java was neglected to be updated to use the new constructor accepting the SanyOutput parameter. By default the calls to SanyOutput in TLAplusParser were sent to a SilentSanyOutput instance, and so this had the effect that all detailed parse error messages occuring at the syntax level and below were silenced. The parse failure itself was recorded, but the specific reason for the failure was not output.

These changes update ParseUnit to use the new TLAplusParser constructor which accepts a SanyOutput parameter, thus ensuring all logging in that class flows through the same SanyOutput instance used by the rest of the parse run. A regression test system is also added to ensure syntax parser errors are not inadvertently silenced in the future. A full fix would involve onboarding SANY's syntax parser level to the standardized error code system, ensuring that all parse error messages are triggered & output appropriately.

This bug was reported by Olav Bunte in #1221. This PR closes #1221.

[SANY][Bugfix]

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/modanalyzer/ParseUnit.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/modanalyzer/ParseUnit.java
@@ -308,7 +308,7 @@ public class ParseUnit {
         try 
         {
             // create parser object
-            parseTree = new tla2sany.parser.TLAplusParser(nis, StandardCharsets.UTF_8.name());
+            parseTree = new tla2sany.parser.TLAplusParser(nis, StandardCharsets.UTF_8.name(), out);
 
             // Here is the one true REAL call to the parseTree.parse() for a file;
             // The root node of the parse tree is left in parseTree.
--- a/tlatools/org.lamport.tlatools/test/tla2sany/output/RecordedSanyOutput.java
+++ b/tlatools/org.lamport.tlatools/test/tla2sany/output/RecordedSanyOutput.java
@@ -1,0 +1,117 @@
+/*******************************************************************************
+ * Copyright (c) 2025 The Linux Foundation. All rights reserved.
+ *
+ * The MIT License (MIT)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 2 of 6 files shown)


ahelwer:
@lemmy this is a one line fix, the  rest is test code.

lemmy (approved):

PR #1225, opened 2025-08-06
Refactor `NamedInputStream` to facilitate reading TLA+ modules from non-file sources

ahelwer:
In furtherance of #719. It took a while to find a decent entry point for refactoring SANY to not rely on files, but `NamedInputStream` is one part of the puzzle. This PR does not remove SANY's dependency on files but is scoped in a way that sets up further changes elsewhere.

The `NamedInputStream` class, before these changes, was the main method used inside SANY to access the contents of module files. That is unchanged, but the class is now a thin wrapper around the `ModuleSourceCode` class which is one small part of the new SANY API that has been cooking in the test directory. Since lifetime management is not Java's strong suit, input streams are de-emphasized in favor of keeping a byte array of file contents around in memory. `ByteArrayInputStream` instances can be created over that byte array on demand, and it doesn't matter whether they're closed or not.

The `ModuleSourceCode` class is written for a world where modules can come from many possible places, not necessarily files. The broad strategy here is for `ModuleSourceCode` to wear `NamedInputStream`'s skin for a bit until the rest of SANY stops using `NamedInputStream`'s deprecated file-specific methods. Then the rest of SANY can slowly migrate to using `ModuleSourceCode` directly, and `NamedInputStream` can eventually be deleted. 

There are two commits in this PR. The first writes unit tests for `NamedInputStream`. The second changes `NamedInputStream` into a wrapper for `ModuleSourceCode` in a way where all the existing tests pass unchanged. `NamedInputStream` is also extensively documented during this process and various methods are marked deprecated.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/api/ModuleSourceCode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/api/ModuleSourceCode.java
@@ -22,6 +22,7 @@
  ******************************************************************************/
 package tla2sany.api;
 
+import java.io.ByteArrayInputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 
@@ -64,18 +65,18 @@ public class ModuleSourceCode {
    * desired (for advanced error reporting, for example) then the function
    * {@link ModuleSourceCode#getTextAsString} can be used.
    */
-  public final byte[] text;
+  private final byte[] text;
 
   /**
    * The origin of this module.
    */
-  public final ModuleOrigin origin;
+  private final ModuleOrigin origin;
 
   /**
    * The path at which this module was found, if sourced from
    * {@link ModuleOrigin#FILESYSTEM}. Null otherwise.
    */
-  public final Path path;
+  private final Path path;
 
   /**
    * Constructs a new instance of the {@link ModuleSourceCode} class.
@@ -91,11 +92,51 @@ public class ModuleSourceCode {
   }
 
   /**
+   * The text of this module's source code, as a UTF-8 encoded byte array.
+   *
+   * @return The module source code.
+   */
+  public byte[] getText() {
+    return this.text;
+  }
+
+  /**
+   * The origin of this module.
+   *
+   * @return The origin of this module.
+   */
+  public ModuleOrigin getOrigin() {
+    return this.origin;
+  }
+  
+  /**
+   * The path associated with this module, if it was sourced from the file
+   * system; otherwise null.
+   *
+   * @return The path associated with this module, if it has one.
+   */
+  public Path getPath() {
+    return this.path;
+  }
+
+  /**
    * Interpret the module text as a UTF-8 encoded string.
    *
    * @return The module text as a UTF-8 encoded string.
    */
   public String getTextAsString() {
     return new String(this.text, StandardCharsets.UTF_8);
+  }
+
+  /**
+   * Makes the module text available as a {@link ByteArrayInputStream}. This
+   * stream can either be closed or not, it does not matter; the underlying
+   * data is a byte array, not a file handle or anything else that needs to
+   * be explicitly released.
+   *
+   * @return The module text, as a {@link ByteArrayInputStream}.
+   */
+  public ByteArrayInputStream getTextAsInputStream() {
+    return new ByteArrayInputStream(this.text);
   }
 }
--- a/tlatools/org.lamport.tlatools/src/util/FileUtil.java
+++ b/tlatools/org.lamport.tlatools/src/util/FileUtil.java
@@ -386,7 +386,7 @@ public class FileUtil
             {
                 NamedInputStream nis = new NamedInputStream(sourceFileName, sourceModuleName, sourceFile);
                 return nis;
-            } catch (FileNotFoundException e)
+            } catch (IOException e)
             {
                 ToolIO.out.println("***Internal error: Unable to create NamedInputStream in toIStream method");
             }
--- a/tlatools/org.lamport.tlatools/src/util/NamedInputStream.java
+++ b/tlatools/org.lamport.tlatools/src/util/NamedInputStream.java
@@ -1,107 +1,246 @@
-// Copyright (c) 2003 Compaq Corporation.  All rights reserved.
-// Portions Copyright (c) 2003 Microsoft Corporation.  All rights reserved.
+/*******************************************************************************
+ * Copyright (c) 2003 Compaq Corporation. All rights reserved.
+ * Copyright (c) 2003 Microsoft Corporation. All rights reserved.
+ * Copyright (c) 2025 The Linux Foundation. All rights reserved.
+ *
+ * The MIT License (MIT)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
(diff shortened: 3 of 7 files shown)


ahelwer:
CI failure is due to tlaplus/examples change, fixed in #1226

lemmy:
> Since lifetime management is not Java's strong suit, input streams are de-emphasized in favor of keeping a byte array of file contents around in memory. 

Please highlight specific shortcomings in the existing JDK functionality for representing TLA+ modules in memory. Creating a custom solution contradicts the goal of eliminating bespoke implementations of functionality already provided by the JDK—such as collections.

ahelwer:
I don't understand what you are asking, but anyway we keep the module around in memory as a byte array and spin ByteArrayInputStream instances from it as needed. If they aren't closed then it isn't a big deal, as compared to FileInputStreams.

lemmy:
You've made a general claim about lifetime management of input streams. Please support this with a concrete example to justify the creation of a bespoke solution.

ahelwer:
Markus, I am doing this PR as a favor to you since you said you wanted to move the tools away from requiring files. I am uninterested in debating Java's lifetime management capabilities. There is nothing bespoke about a byte array, which is more integral to Java than anything in the standard library.

tlatools/org.lamport.tlatools/src/util/NamedInputStream.java
```diff
@@ -1,107 +1,237 @@
...
+ * safe to call under all circumstances. Several methods exist under the
+ * assumption that this class always refers to an underlying file with a
+ * valid path, which in the future is hoped not to be the case. These methods
+ * were written to function normally in the case that there *is* an actual
+ * underlying file, which will be the case until the rest of the codebase is
+ * changed to avoid using these methods and their implicit assumption of file
+ * existence. The unsafe methods have all been marked as deprecated.
  */
-public class NamedInputStream extends FileInputStream
+public class NamedInputStream implements Closeable
```
Calvin-L: This class is used directly by the Toolbox. Are these changes compatible with the current Toolbox code?

tlatools/org.lamport.tlatools/src/util/FileUtil.java
```diff
@@ -386,7 +386,7 @@ public static NamedInputStream createNamedInputStream(String name, FilenameToStr
             {
                 NamedInputStream nis = new NamedInputStream(sourceFileName, sourceModuleName, sourceFile);
                 return nis;
-            } catch (FileNotFoundException e)
+            } catch (IOException e)
             {
                 ToolIO.out.println("***Internal error: Unable to create NamedInputStream in toIStream method");
             }
```
Calvin-L, line 392: Why broaden the set of exceptions caught here?

This seems like a meaningful behavior change. Remember, `FileNotFoundException` is definitive evidence that the file does not exist. `IOException` is ambiguous: the file may or may not exist!

tlatools/org.lamport.tlatools/src/tla2sany/api/ModuleSourceCode.java
```diff
@@ -98,4 +128,15 @@ public ModuleSourceCode(byte[] text, ModuleOrigin origin, Path path) {
   public String getTextAsString() {
     return new String(this.text, StandardCharsets.UTF_8);
   }
+
+  /**
+   * Makes the module text available as a {@link InputStream}. The caller is
+   * responsible for lifetime management of the {@link InputStream} instance,
+   * although it is a {@link ByteArrayInputStream} so closing it is a no-op.
```
Calvin-L: I would remove this last clause ("although ..."), since it exposes implementation details.

I would also s/lifetime management/closing/, which seems a bit clearer, i.e. `The caller is responsible for closing the returned {@link InputStream}.`

Calvin-L (changes requested):
I left some small comments, but you should ignore them. After doing a little more thinking and poking at the source code, I have decided I am (gently) opposed to this change.

I am of course very happy to see progress on #719, and I'm also happy with the test-first approach here.

But I think this change further corrupts an already-messy part of the codebase. Now, `NamedInputStream` isn't even a stream! I don't see how this step makes progress. (Maybe I'm just failing to imagine the next step?)

No matter what we do, there will be some part of SANY that has to work with the `InputStream` API, and `NamedInputStream` as it exists today seems like a reasonable class that could be helpful in SANY's raw, stream-conscious layer.

----

I'm not certain what should be our first step toward #719. But I'm thinking it should be to create better abstractions around `FilenameToStream` (something like the `Resolver` interface we already have in the test code) and to start replacing raw `FilenameToStream` + `File` + `InputStream` with a higher-level `Resolver` + `ModuleSourceCode` + etc.

`SpecObj` might be a good place to start, since its use of `FilenameToStream` is relatively well-contained.

We will also have to revisit the Toolbox code when building our abstraction layer, since IIRC the Toolbox also uses `FilenameToStream` directly.

ahelwer:
So in my view the only place in the codebase that *needs* to use `InputStream` is the `TLAplusParser` syntax parser class. Everything in between reading in TLA+ module text from somewhere (file, jar, string) and giving it to `TLAplusParser` should use `byte[]`, and not one of the `InputStream` classes. The benefits of using `byte[]` instead of `InputStream` are several:
 -  Can read `byte[]` multiple times
 - Don't need to worry about lifetime management in a language with limited faculties for statically ensuring `close()` is called exactly once
 - One less piece of state to carry around - don't need to worry about how far into an input stream has been read, can access the `byte[]` wherever
 - If a class actually *needs* an `InputStream` then it is easy to construct a `ByteArrayInputStream` from a `byte[]`

So that is what these changes facilitate.

tlatools/org.lamport.tlatools/src/util/FileUtil.java
ahelwer: Since we read the file in directly in the constructor, it throws an `IOException` if the file cannot be read for some reason, which is distinct from the file not existing. `FileNotFoundException` is itself an `IOException`, though. Thus this catches both of those errors.

tlatools/org.lamport.tlatools/src/tla2sany/api/ModuleSourceCode.java
ahelwer: How should lifetime management requirements for the caller be exposed? It really does not matter whether the caller actually closes the given input stream. I suppose we could return a `ByteArrayInputStream` directly, which makes this clear to the caller (if they know how that class works).

tlatools/org.lamport.tlatools/src/util/NamedInputStream.java
ahelwer: Yes, because the class behavior is unchanged (to the extent verifiable with testing) when using files and the toolbox only uses files with this class.

tlatools/org.lamport.tlatools/src/tla2sany/api/ModuleSourceCode.java
ahelwer: Changed this to return a `ByteArrayInputStream` so it's clear that it doesn't matter whether the stream is closed or not.

ahelwer:
@Calvin-L with regard to:

>Now, NamedInputStream isn't even a stream!

This was done to minimize the diff. I would be happy to rename it to something else, but it's used widely in the codebase so would create quite a large rename diff.

lemmy:
> We will also have to revisit the Toolbox code when building our abstraction layer, since IIRC the Toolbox also uses `FilenameToStream` directly.

The Toolbox should be treated as a nice-to-have in this context. If the community steps up to migrate it, that’s great—but it shouldn’t block #719.

Issue #1227, opened 2025-08-11
Remove SANY syntax parser has "pragma" construct


ahelwer:
Another weird thing found when analyzing the syntax parser, similar to #1019, https://github.com/tlaplus/rfcs/issues/20, and https://github.com/tlaplus/rfcs/issues/25.

SANY has support for "pragmas", which take this form:
```tla
---> 1 foo 2 bar 3 baz
---- MODULE Test ----
====
```
Pragmas are enabled at the start of each compilation unit (module file); see: https://github.com/tlaplus/tlaplus/blob/9d3e8c6f90ed2644d23d9d1791efc3c7283c3878/tlatools/org.lamport.tlatools/javacc/tla%2B.jj#L1653-L1655

The pragma starts with the token `--->` and then is followed by a series of numbers or identifiers.

It is unclear what the purpose of these pragmas are. I could not find any tool which consumes them. Thus I propose they be removed.

lemmy:
Where do these pragmas end up in SANY's AST?

ahelwer:
They are ignored by the semantic level and are not exposed there. They are also not reachable from the root `TLAplusParser.parseTree` `SyntaxTreeNode` instance. It seems as though they can only be found by running `TLAplusParserTokenManager` directly. The parsing code in `TLAplusParser` processes these pragmas, but does not attach them to the parse tree.

PR #1228, opened 2025-08-13
SANY: Prototype incremental parsing functionality in unit tests

ahelwer:
In furtherance of #1096. Also relevant to #1229 and #1099.

This started out with tests for SANY's tokenizer and belchDEF method. It then extended to getting working prototypes of incremental parsing working. To facilitate this there were some minor visibility & functionality changes to the SANY's internal API:
- Changed the visibility of the `belchDEF()` method in `TLAplusParser` to facilitate unit testing
- Added/modified the custom constructors in `TLAplusParser` to accept a byte array and remove the `encoding` parameter since it is always set to `UTF-8`, then modified the constructor call-sites appropriately; parameters had to be re-ordered to avoid ambiguity with the auto-generated `TLAplusParser` constructors.
- Changed the visibility of the `processOperator()` and `generateExpression()` methods in `Generator` to facilitate unit testing
- Removed a side-effect from the `processOperator()` method  by returning an `OpDefNode` to be added to the module's operator definitions at the call-site instead adding it in the method itself, thus facilitating parsing operator definitions without an enclosing module

Found #1227 during this work.

[SANY][Testing]


ahelwer:
@lemmy this is largely test code with some relatively minor API changes so should be good to merge.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCDebuggerExpression.java
```diff
@@ -102,8 +101,7 @@ private static OpDefNode process(final SpecProcessor processor, final ModuleNode
 		final String wrapper = "---- MODULE %s ----\nEXTENDS %s\n%s == %s\n====";
 		String wrappedConditionExpr = String.format(wrapper, bpModName, rootModName, bpOpDef, conditionExpr);
 		byte[] wrappedConditionExprBytes = wrappedConditionExpr.getBytes(StandardCharsets.UTF_8);
-		InputStream sourceCode = new ByteArrayInputStream(wrappedConditionExprBytes);
-		TLAplusParser parser = new TLAplusParser(sourceCode, StandardCharsets.UTF_8.name());
+		TLAplusParser parser = new TLAplusParser(wrappedConditionExprBytes, StandardCharsets.UTF_8, new SilentSanyOutput());
```
lemmy: How about a TLAplusParser Ctor with `StandardCharsets.UTF_8` and `SilentSanyOutput()` as defaults?

lemmy:
The commit message mentions “prototype,” and there are several changes in the `TLAplusParser`. Are these changes refactoring (without altering the behavior) to enhance testing, or do they also introduce behavioral changes in `TLAplusParser`? If the latter, please provide details about the changes.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCDebuggerExpression.java
ahelwer: The first one is a very good idea, although I would like to force users as much as possible to provide a `SanyOutput` parameter of their own to avoid another #1221 issue.

ahelwer:
@lemmy ah, the "prototype" refers to prototyping what the code to incrementally parse opdefs & expressions looks like in the unit tests. In the near future I hope to use this to simplify both the expression debugging code and your new command-line invariant processing code. I've updated the top-level description with notes about the non-test API changes.

lemmy:
I understand you’re saying that this commit isn’t intended to change SANY’s behavior. If that’s the case, please add the `[Refactor]` or `[Test]` tag to the commit message. This will help our future selves easily exclude it when bisecting the git history to diagnose regressions.

ahelwer:
Done!

tlatools/org.lamport.tlatools/javacc/tla+.jj
```diff
@@ -1629,26 +1631,13 @@ CompilationUnit() : {
...
-    * that file                                                            *
-    *                                                                      *
-    *    static int defaultLexState = 0;                                   *
-    *                                                                      *
-    * leading us to guess that this is setting the lexical state to        *
-    * DEFAULT.                                                             *
-    ***********************************************************************/
+  tempASTN = Module() {
+    // Reset tokenizer state to default value after parsing module.
+    token_source.SwitchTo(token_source.defaultLexState);
```
lemmy, line 1636: `tla2sany.parser.TLAplusParserTokenManager.defaultLexState` is not final.  Still, passing `token_source.defaultLexState` instead of `0` constitutes no behavior change?

tlatools/org.lamport.tlatools/src/tla2sany/parser/TLAplusParser.java
```diff
@@ -7768,6 +7770,22 @@ final private boolean jj_2_74(int xla) {
     finally { jj_save(73, xla); }
   }
 
+  final private boolean jj_3_52() {
```
lemmy, line 7773: This change and the changes below are the result of non-determinism in the parser generator (JavaCC)?

tlatools/org.lamport.tlatools/javacc/tla+.jj
ahelwer: Correct, that field is auto-generated and not modified anywhere.

tlatools/org.lamport.tlatools/src/tla2sany/parser/TLAplusParser.java
ahelwer: Correct.

lemmy: Could we eliminate this non-determinism? Reviewing PRs related to the parser would become much easier.

ahelwer: Unfortunately probably not. The best way is to just ignore the auto-generated files in the PR diff. Unless github exposes functionality to not show diffs of certain files, the only other alternative is to remove the auto-generated files from source control entirely, which then has the issue that a clean checkout of the repository (before running the ant generate target) will show undefined symbol errors when opened in an IDE. However, I would support that if you want to do it.

It is also possible that upgrading the version of JavaCC we use would be more stable in this regard.

lemmy (approved):

ahelwer:
@lemmy I don't think we have bandwidth for dual signoff these days so this can probably just be merged.

lemmy:
Understood. I’m just a bit uneasy about the generation artifacts. Since they result from non-determinism in JavaCC, is it possible to manually revert these changes in the git commit?

ahelwer:
I wouldn't say it's nondeterminism - if you re-run JavaCC on the source files you'll get the same output every time - but something makes JavaCC sensitive so it reorders the generated code somewhat arbitrarily. I guess analogous to #1044.

As far as manually stripping these changes out, then we have the issue that the tla+.jj source file goes out of sync with the generated code.

PR #1235, opened 2025-08-26
Misc XMLExporter enhancements

lemmy:
To support:
* https://github.com/tlaplus/vscode-tlaplus/issues/371
* https://github.com/tlaplus/vscode-tlaplus/issues/378

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
@@ -36,6 +36,7 @@ import java.util.HashSet;
 import java.util.Hashtable;
 import java.util.Iterator;
 import java.util.Objects;
+import java.util.function.BiPredicate;
 import java.util.stream.Collectors;
 
 import org.w3c.dom.Document;
@@ -1495,7 +1496,7 @@ public class OpDefNode extends OpDefOrDeclNode
     }
   }
 
-  protected Element getSymbolElement(Document doc, SymbolContext context) {
+  protected Element getSymbolElement(Document doc, SymbolContext context, BiPredicate<SemanticNode, SemanticNode> filter) {
     Element ret = null;
     switch (getKind()) {
       case UserDefinedOpKind:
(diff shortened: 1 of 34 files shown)


tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
```diff
@@ -1500,6 +1502,13 @@ protected Element getSymbolElement(Document doc, SymbolContext context) {
         ret = doc.createElement("UserDefinedOpKind");
         ret.appendChild(appendText(doc,"uniquename",getName().toString()));
         ret.appendChild(appendText(doc,"arity",Integer.toString(getArity())));
+		if (getTreeNode().getAttachedComments().length > 0) {
```
ahelwer: You can use the `this.getPreComments()` API instead of using the tree node.

tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
```diff
@@ -187,12 +196,25 @@ static void moduleToXML(String... args) throws XMLExportingException {
...
-        //Element ext_e = externalModules[j].exportDefinition(doc, context);
-        Element ext_e = externalModules[j].export(doc, context);
-        rootElement.appendChild(ext_e);
-      }
+      
+		if (restricted) {
+			final BiPredicate<SemanticNode, SemanticNode> filter = ((s1, s2) -> {
+				if (s1 instanceof OpDefNode && s2 instanceof ModuleNode) {
+					final OpDefNode odn = (OpDefNode) s1;
+					return s2.equals(odn.getOriginallyDefinedInModuleNode());
```
ahelwer: Possibly this can be done without the second module parameter using the `OpDefNode.hasSource()` API call, which will return `true` if the operator was defined in another module imported with `INSTANCE`. I have not tested whether it will return `true` if imported using `EXTENDS`.

ahelwer: An even better solution would be to refactor the XML conversion code by tearing it all out of these classes into its own class in the `tla2sany.xml` namespace using the Visitor pattern. Then the filter could easily be part of the xml conversion class.

lemmy: The check has been expanded to include `OpDefOrDeclNodes` (VARIABLES and CONSTANTS), which lacks the `hasSource` attribute.

lemmy: > An even better solution would be to refactor the XML conversion code by tearing it all out of these classes into its own class in the `tla2sany.xml` namespace using the Visitor pattern. Then the filter could easily be part of the xml conversion class.

This makes sense, but I don’t have the bandwidth for such a refactoring.

tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
```diff
@@ -59,6 +62,8 @@ static final void printUsage(final PrintStream out) {
         "-o", "Offline mode; skip XML schema validation step.", true));
     args.add(new UsageGenerator.Argument(
         "-t", "Terse; format XML output without tabs or newlines.", true));
+    args.add(new UsageGenerator.Argument(
+        "-r", "Restrict mode; restrict output to current module.", true));
```
ahelwer: I think this explanation could be better. Perhaps "Only include declarations & definitions local to each module, not its imported modules"?

lemmy: Done: “Restrict mode: Include only declarations and definitions of the specified module, excluding extended or instantiated modules.” (I used “extended or instantiated” instead of “imported” because the latter is not TLA+ jargon.)

PR #1250, opened 2025-10-20
SANY: output everything to stdout by default

ahelwer:
Changes in PR #1209 split some SANY output between stderr and stdout. This was a breaking change to tools which parse SANY's command line output and might capture only one of stdout and stderr, like the VS Code extension. This changes SANY to only output to stdout by default. In the future, command line parameters can be added to better control SANY's output according to user preferences.

Changes are in response to https://github.com/tlaplus/vscode-tlaplus/issues/433

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
@@ -471,7 +471,7 @@ public class SANY {
 
     final SanyOutput out = new OutErrSanyOutput(
         ToolIO.out,
-        ToolIO.err,
+        Boolean.getBoolean(SANY.class.getName() + ".errors2stderr") ? ToolIO.err : ToolIO.out,
         LogLevel.INFO,
         LogLevel.ERROR
     );


tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -471,7 +471,7 @@ else if (args[i].toLowerCase().equals("-help")) {
 
     final SanyOutput out = new OutErrSanyOutput(
         ToolIO.out,
-        ToolIO.err,
+        ToolIO.out,
```
lemmy: Although it may seem improbable, I propose the following in case there are any potential adopters:

```java
Boolean.getBoolean(SANY.class.getName() + ".errors2stderr") ? ToolIO.err : ToolIO.out,
```

lemmy: `-Dtla2sany.drivers.SANY.errors2stderr=true` can then be mentioned in the commit message.

FedericoPonzi:
Thanks for the quick turnaround. Splitting each output stream seems reasonable IMO. Would it make sense to have a new version bump/release or something to avoid breaking the current version users? I think it was agreed to stick to rolling releases as these are less work, but some kind of versioning would still be useful for these kind of changes.

lemmy:
> Thanks for the quick turnaround. Splitting each output stream seems reasonable IMO. Would it make sense to have a new version bump/release or something to avoid breaking the current version users? I think it was agreed to stick to rolling releases as these are less work, but some kind of versioning would still be useful for these kind of changes.

The Toolbox and the TLA+ Tools follow semantic versioning at the OSGi layer. However, there is no versioning applied beyond that. In particular, the VSCode extension currently invokes the TLA+ tools via the command line and parses their raw output. Without a defined version header or structured output format, bumping version numbers alone would not help ensure compatibility.

From an architectural standpoint, having the VSCode extension invoke the TLA+ tools through the command line was a design mistake. A more robust and version-aware approach—such as using the Language Server Protocol or a similar IPC mechanism—would have been preferable.

Given our rolling release model, we must avoid introducing breaking changes. When such changes are unavoidable, we should provide a clear migration path to ensure a smooth transition for all known downstream consumers (e.g., the Toolbox and VSCode extension), as well as unknown or internal consumers within organizations.

PR #1257, opened 2025-11-21
Replace tla2sany.utilities.Vector with java.util.ArrayList

younes-io:
- Refactoring: Replaced the custom `tla2sany.utilities.Vector` with `java.util.ArrayList` across the parser, semantic analyzer, and other modules
- Verification: Verified with TokenizerTests, SemanticCorpusTests, and OffHeapDiskFPSetTest (all passed)
- Environment: Added `.devcontainer/` configuration for reproducible builds


ahelwer:
This is something I wanted to do eventually, but I suggest a change in approach: see Calvin's comment [here](https://github.com/tlaplus/tlaplus/pull/1210#discussion_r2208863998).

Calvin-L:
Our thinking around replacing these custom data structures is scattered in random comments like the one Andrew linked to. I've opened #1258 to make a central (and hopefully easier-to-discover) home for it.

tlatools/org.lamport.tlatools/src/tla2sany/utilities/Vector.java
```diff
@@ -1,154 +0,0 @@
...
-  public final void setElementAt( E obj, int i ) {
-    if (i < 0 || i >= size )
-      throw new ArrayIndexOutOfBoundsException();
-    else 
-      info[ i ] = obj;
-  }
-
-  public final boolean contains (E obj) {
-    for (int i = 0; i < size; i++) {
-      if ( info[ i ] == obj ) return true;
```
Calvin-L, line 115: Did you account for the fact that this `contains()` is not the same as `ArrayList.contains()`, despite its identical name? This one uses reference equality instead of `.equals()`.

Calvin-L:
@younes-io Thanks a ton for getting started on this change!

Can I ask you to split this PR up? Right now it does at least two things:
 - removes `tla2sany.utilities.Vector`
 - adds `.devcontainer` infrastructure and many changes to Eclipse project files

Those two things should be reviewed and merged separately.

tlatools/org.lamport.tlatools/src/tla2sany/utilities/Vector.java
younes-io: I checked the current branch: the only place where `ArrayList.contains` now sees an object with a meaningful `.equals` is `semanticAnalysisVector.contains(currentParseUnit.getName())`, a list of  module-name strings, where value equality is desired. Everywhere else we switched to `ArrayList` of classes without an `equals` override (`ParseUnit`, `ModulePointer`, `SymbolNode`...), so `ArrayList.contains` still behaves like reference checks.. 
Happy for extra eyes!

younes-io:
@Calvin-L 
> Can I ask you to split this PR up? Right now it does at least two things:

Thank you for the feedback, I have created separate PRs:
#1260 & #1261
#1259 

I will, therefore, be closing this one.

PR #1260, opened 2025-11-22
Refactor custom Vector usage to java.util.ArrayList

younes-io:
Replaced the custom tla2sany.utilities.Vector class with the standard java.util.ArrayList to improve maintainability and align with modern Java conventions. 

- Replaced Vector with ArrayList in parser, semantic, modanalyzer, and explorer packages
- Deleted tla2sany.utilities.Vector and VectorEnumeration
- Inlined custom appendNoRepeats logic in SpecObj.java
- Updated TLAplusParser.java to use Iterators instead of Enumerations.

Verified with regression tests including TokenizerTests and SemanticCorpusTests.

Related to @Calvin-L 's  #1258 

- [x] `tla2sany.utilities.Vector` (similar to `java.util.ArrayList`)
- [x] `tla2sany.utilities.VectorEnumeration` (enumerations are deprecated in favor of `java.util.Iterator`)

PR #1261, opened 2025-11-22
Refactor pcal package to use java.util.ArrayList instead of Vector

younes-io:
- Fix Ant JUnit wiring so `ant test` runs without external optional libs; add bundled `ant-junit` jars, taskdefs, and keep `test-class` ahead of `class` on junit classpaths (tlatools/org.lamport.tlatools/customBuild.xml)
- Restore PlusCal translator’s identifier scanning to include ADDED tokens (like `pc`), reintroducing `[self]` subscripts in generated TE specs and unbreaking PlusCal model-checking (tlatools/org.lamport.tlatools/src/pcal/TLAExpr.java)
- Keep precise parser error reporting for missing `with` bodies by catching the correct exception type (tlatools/org.lamport.tlatools/src/pcal/ParseAlgorithm.java)

Tests: `ant -f tlatools/org.lamport.tlatools/customBuild.xml test`

to be merged after #1260 is merged


younes-io:
Small follow‑up on this one. It covers the PlusCal refactor plus the Ant JUnit wiring and parse fixes. Thanks for the time.

@Calvin-L : could you please take a look at this one, if possible? Thank you

Issue #1258, opened 2025-11-21
Replace custom data structures with `java.util` equivalents


Calvin-L:
The TLA+ tools have some hand-rolled data structures that duplicate functionality already available in the [Java Collections Framework (JCF, `java.util` package)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/package-summary.html):

- [ ] `util.Set` (similar to `java.util.HashSet`)
- [ ] `util.InternTable` (similar to `java.util.HashMap`)
- [ ] `tla2sany.utilities.Vector` (similar to `java.util.ArrayList`)
- [ ] `tla2sany.utilities.VectorEnumeration` (enumerations are deprecated in favor of `java.util.Iterator`)
- [x] `tla2sany.utilities.Stack` --- #1210
- [ ] `tlc2.util.BitVector` (similar to `java.util.BitSet`)
- [ ] `tlc2.util.List` (similar to `java.util.LinkedList`)
- [ ] `tlc2.util.Vect` (similar to `java.util.ArrayList`)

There are good reasons to replace these classes (or large portions of their internals):
 - Less code for us to maintain
 - We benefit from the enormous amount of testing and formal methods already poured into the JCF
 - Potentially improved performance

There are also good reasons _not_ to replace these classes:
 - Risk of bugs (our custom implementations differ in subtle ways from the JCF collections)
 - Potentially degraded performance
 - [There are more important things to work on than replacing 20-year-old code that we know works](https://github.com/tlaplus/tlaplus/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22)

# It is entirely possible that we decide, based on evidence, to _retain_ these classes, not replace them

In this case, the only outcome would be a code comment justifying the custom implementation.

Documenting the justification should be considered as much a victory as replacing the class.

# Suggested process for making this refactor (or deciding not to)

 1. Keep your change tightly focused: work on _one_ replacement at a time.
 2. Write many new tests for the _existing_ implementation. Use the existing code to discover weird behaviors (like the [known discrepancy in `Vector.contains()`](https://github.com/tlaplus/tlaplus/pull/328#issuecomment-542160722)), and cover as many corner cases as you can find.
 3. Commit.
 4. Make the existing implementation extend its `java.util` equivalent, and change the implementation to delegate to `super`, ensuring that all tests still pass. If the new code is awkward or includes many special cases, then consider retaining the existing implementation instead.
[...]

PR #1262, opened 2025-11-22
refactoring `VectorEnumeration`

younes-io:
https://github.com/tlaplus/tlaplus/issues/1258

- [X] `tla2sany.utilities.VectorEnumeration` (enumerations are deprecated in favor of `java.util.Iterator`)

---

- Delegated to `ArrayList` while retaining legacy behaviors: identity-based `contains`, `insertElementAt` rejects `index == size`, `elements()` returns a snapshot. Documented in the class Javadoc
- Added regression tests for those quirks: `tlatools/org.lamport.tlatools/test/tla2sany/utilities/VectorTest.java`
- Ant test harness is self-contained (bundled ant-junit jars and taskdefs in `customBuild.xml`); `ant -f tlatools/org.lamport.tlatools/customBuild.xml test` now runs without external ant-junit
- Full suite run succeeded: `ant -f tlatools/org.lamport.tlatools/customBuild.xml test`

By the way, the wrapper is retained (for now) because `contains` uses reference equality and `elements()` is snapshot-based; changing to plain `List`/`Iterator` would alter observable semantics. @Calvin-L : what's your take on this? I kept the wrapper and documented the quirks. If you think identity isn't needed at call sites, we can switch to `List`/`Iterator` and drop the wrapper; otherwise we can keep it with the documented justification.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/utilities/Vector.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/utilities/Vector.java
@@ -16,8 +16,6 @@ import java.util.Iterator;
  * are audited for identity-based membership checks.
  */
 public class Vector<E> extends ArrayList<E> {
-  private static final long serialVersionUID = 5334890925756142170L;
-
   static int defaultSize = 10;
 
   public Vector() {
@@ -69,8 +67,8 @@ public class Vector<E> extends ArrayList<E> {
 
   @Override
   public final boolean contains(Object obj) {
-    for (Iterator<E> iterator = super.iterator(); iterator.hasNext();) {
-      if (iterator.next() == obj) {
+    for (E value : this) {
+      if (value == obj) {
         return true;
       }
     }


tlatools/org.lamport.tlatools/customBuild.xml
```diff
@@ -19,6 +19,16 @@
 	<!-- setup classpath -->
 	<path id="project.classpath" />
 
+	<!-- Provide the JUnit Ant tasks even when optional Ant libraries (ant-junit.jar) are not installed system-wide. -->
```
lemmy: Changes to `customBuild.xml` do not belong in this PR.
Also, please clarify under which circumstances these libraries are missing, since they should be present when the IDE is set up according to the instructions at: https://github.com/tlaplus/tlaplus/tree/master/general/ide

tlatools/org.lamport.tlatools/customBuild.xml
```diff
@@ -508,8 +518,8 @@
 				<pathelement location="lib/lsp/org.eclipse.lsp4j.jsonrpc.debug_0.21.1.v20230829-0012.jar"/>
 				<pathelement location="lib/CommunityModules.jar"/>
 				<pathelement location="test-model/UserModuleOverrideFromJar.jar" />
-				<pathelement path="${class.dir}" />
 				<pathelement path="${test.class.dir}" />
+				<pathelement path="${class.dir}" />
```
lemmy, line 511: Please minimize the git diff by avoiding unrelated code-style changes. This helps keep the review focused and easier to follow.

tlatools/org.lamport.tlatools/customBuild.xml
younes-io: My local Ant didn't have the optional ant-junit/ant-optional tasks, so I bundled  the JARs to run `ant -f tlatools/org.lamport.tlatools/customBuild.xml` test. I've reverted those build changes to keep this PR focused on the Vector  refactor

tlatools/org.lamport.tlatools/customBuild.xml
younes-io: reverted anyway

tlatools/org.lamport.tlatools/src/tla2sany/utilities/Vector.java
```diff
@@ -1,143 +1,100 @@
...
+ * <ul>
+ *   <li>{@link #contains(Object)} uses reference equality (==), not {@link Object#equals(Object)}.</li>
+ *   <li>{@link #insertElementAt(Object, int)} rejects {@code index == size()} (cannot append via insert).</li>
+ *   <li>{@link #elements()} returns a snapshot Enumeration taken at call time.</li>
+ * </ul>
+ * These differences make a drop-in replacement with {@link ArrayList} unsafe until all call sites
+ * are audited for identity-based membership checks.
+ */
+public class Vector<E> extends ArrayList<E> {
+  private static final long serialVersionUID = 5334890925756142170L;
```
Calvin-L: Including `serialVersionUID` is fine but unnecessary. There's a good overview [here](https://github.com/projectlombok/lombok/wiki/WHY-NOT:-serialVersionUID):

> (almost) all tools are wrong: They ship with code quality plugins that remark on serializable classes without a serialVersionUID field as being wrong. This is incorrect; the vast majority of your serializable classes **should not have a serialVersionUID field**, because this gets you the desired default behaviour of no binary compatibility [when you change the class].

tlatools/org.lamport.tlatools/src/tla2sany/utilities/Vector.java
```diff
@@ -1,143 +1,100 @@
...
+    super.set(i, obj);
   }
 
-  public final boolean contains (E obj) {
-    for (int i = 0; i < size; i++) {
-      if ( info[ i ] == obj ) return true;
+  @Override
+  public final boolean contains(Object obj) {
+    for (Iterator<E> iterator = super.iterator(); iterator.hasNext();) {
+      if (iterator.next() == obj) {
```
Calvin-L: This can be a bit simpler:
```java
for (E value : this) {
  if (value == obj) {
    ...
```
For-each loops desugar to almost exactly the code you have here.

(I see you've been very careful to use `super.iterator()` in this code, and it's true the for-each loop will call `this.iterator()` instead. But, since this class doesn't override `iterator()`, it will be equivalent either way.)

tlatools/org.lamport.tlatools/src/tla2sany/utilities/Vector.java
younes-io: Thank you for this very interesting feedback

PR #1263, opened 2025-11-25
Add documentation on ASSUME and INVARIANT usage in TLA+.

lemmy:
Created a new document outlining the differences between the `ASSUME` and `INVARIANT` constructs in TLA+. This includes explanations of constant-level and state-level formulas, along with examples of their usage. Additionally, reference this new documentation at explain.tlapl.us for better user guidance on constant-level formulas.

[Feature][TLC]

### DIFF
--- a/docs/assumptions-and-invariants.md
+++ b/docs/assumptions-and-invariants.md
@@ -1,6 +1,28 @@
 # When to use ASSUME/ASSUMPTION and when to use INVARIANT/INVARIANTS with TLC?
 
-The `ASSUME` (or `ASSUMPTION`) construct in TLA+ is used to introduce constant-level (level 0) formulas—facts that must hold about the constants and fixed structures of a specification. These constraints are evaluated before TLC begins exploring behaviors and are not interpreted as properties of system states.
-A constant-level (level 0) formula contains no state variables, no primed variables, and no temporal operators. It expresses purely mathematical truths about constants, sets, or other static objects. Such formulas might assert, for example, that a constant N is a natural number or that a given set S is nonempty. These belong in ASSUME clauses within the module.
+The `ASSUME` (or `ASSUMPTION`) keyword in TLA+ is used to introduce constant-level (level 0) formulas—facts that must hold about the constants and fixed structures of a specification. These constraints are evaluated before TLC begins exploring behaviors and are not interpreted as invariants or properties of system states.
+A constant-level (level 0) formula contains no state variables, no primed variables, and no temporal operators. It expresses purely mathematical truths about constants, sets, or other static objects. Such formulas might assert, for example, that a constant N is a natural number or that a given set S is nonempty.
 
 In contrast, `INVARIANT` (or `INVARIANTS`) entries in a TLC configuration file refer to state-level (level 1) formulas—predicates about the values of state variables in a single state. Although these predicates must be true in every reachable state, they are not temporal formulas themselves; rather, TLC interprets them as safety checks. An invariant might state that a counter never becomes negative or that two variables always remain equal.
+
+```tla
+---- MODULE Example ----
+CONSTANT N
+
+ASSUME N \subseteq (Nat \ {0})
+
+VARIABLE x
+
+...
+
+Inv == x \in N
+====
+```
+
+```cfg
+...
+CONSTANT N = {1, 2, 3}
+INVARIANT Inv
+```
+
+Note that `ASSUME` is a keyword in TLA+, whereas `INVARIANT` is a keyword used in TLC configuration files. Furthermore, assumptions introduced with `ASSUME` are automatically checked by TLC, while a state predicate such as `Inv` is checked only if the configuration file explicitly instructs TLC to do so by including it under an `INVARIANT` entry.
--- a/docs/invariants-and-properties.md
+++ b/docs/invariants-and-properties.md
@@ -3,7 +3,7 @@
 The `INVARIANT` (or `INVARIANTS`) keyword in a TLC configuration file is used to specify a state-level (level 1) formula—a condition that must hold in all reachable states of the system. This is typically how safety properties are expressed—asserting that "something bad never happens." For instance, an invariant might state that a certain variable is always non-negative.
 A state-level (level 1) formula is a predicate about the values of state variables in a single state. It contains state variables but no primed variables and no temporal operators.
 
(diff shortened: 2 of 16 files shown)


lemmy:
@craft095 @hwayne @muenchnerkindl If possible, please review the following TLC change. The main purpose of this update is to revise the documentation to improve the user experience.

docs/invariants-and-properties.md
```diff
@@ -1,14 +1,9 @@
...
 
-However, the same condition can also be written as a temporal property using TLA+ syntax. In this case,
-assuming an invariant `I`, it would be written as `[]I` (read as "always I") in the TLA+ specification.
-Then, in the config file, instead of using `INVARIANT I`, you would use `PROPERTY I`.
-Important note: If you switch from `INVARIANT I` to `PROPERTY I` without updating the definition of `I`
-to include the temporal `[]` operator, the meaning changes drastically. Without the `[]`, TLC will only
-check that `I` holds in the initial states—not in all reachable states—potentially missing violations
-of the intended property.
\ No newline at end of file
+However, the same condition can also be written as a temporal-level (level 3) formula using TLA+ syntax. In this case, assuming an invariant `I`, it would be written as `[]I` (read as "always I") in the TLA+ specification. Then, in the config file, instead of using `INVARIANT I`, you would use `PROPERTY []I`.
```
craft095: I don't think this syntax is supported: `PROPERTY []I`. It might be restated as
In this case, assuming an invariant `I`, it would be written as `P == []I` (read as "always I") in the TLA+ specification. Then, in the config file, instead of using `INVARIANT I`, you would use `PROPERTY P`.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -529,9 +529,10 @@ private static String getMessage0(int messageClass, int messageCode, String[] pa
         	}
             break;
         case EC.TLC_INVARIANT_CONSTANT_LEVEL:
-			b.append(
-					"The invariant %1% is a constant-level formula (i.e., it contains no variables, primes, or temporal operators) and evaluates to %2%. "
-					+ "To assert constant-level formulas in your spec, use ASSUME ConstInv, or to give the assumption a name such as YourAssumption, use ASSUME YourAssumption == ConstInv.");
+		b.append(
+				"The invariant %1% is a constant-level formula (i.e., it contains no variables, primes, or temporal operators) and evaluates to %2%. "
+				+ "To assert constant-level formulas in your spec, use ASSUME ConstInv. If you optionally want to give the assumption a name, write ASSUME YourAssumption == ConstInv instead. "
+				+ "See https://explain.tlapl.us/assumptions-and-invariants for additional details.");
```
craft095, line 535: BTW, `https://explain.tlapl.us/` gives 404. It would be nice to see the list of available articles instead.

docs/assumptions-and-invariants.md
```diff
@@ -0,0 +1,6 @@
+# When to use ASSUME/ASSUMPTION and when to use INVARIANT/INVARIANTS with TLC?
+
+The `ASSUME` (or `ASSUMPTION`) construct in TLA+ is used to introduce constant-level (level 0) formulas—facts that must hold about the constants and fixed structures of a specification. These constraints are evaluated before TLC begins exploring behaviors and are not interpreted as properties of system states.
+A constant-level (level 0) formula contains no state variables, no primed variables, and no temporal operators. It expresses purely mathematical truths about constants, sets, or other static objects. Such formulas might assert, for example, that a constant N is a natural number or that a given set S is nonempty. These belong in ASSUME clauses within the module.
```
muenchnerkindl: I find the last sentence of this paragraph too peremptory: as you say above, ASSUME clauses assert facts that should hold about the constant parameters. Constant-level formulas may be used elsewhere, for example
```
THEOREM \A n, a, b, c \in Nat \ {0} : n > 2 => a^n + b^n # c^n
```
and these should not appear in ASSUME clauses (also TLC may not be able to check them). I suggest dropping the last sentence.

docs/assumptions-and-invariants.md
```diff
@@ -0,0 +1,6 @@
...
+
+In contrast, `INVARIANT` (or `INVARIANTS`) entries in a TLC configuration file refer to state-level (level 1) formulas—predicates about the values of state variables in a single state. Although these predicates must be true in every reachable state, they are not temporal formulas themselves; rather, TLC interprets them as safety checks. An invariant might state that a counter never becomes negative or that two variables always remain equal.
```
muenchnerkindl: There is a dissymmetry between ``ASSUME`` in the previous paragraphs, which is a TLA+ keyword, and ``INVARIANT``, which appears in configuration files, not in TLA+ modules. Make it clear that assumptions appearing in a TLA+ module are checked without requiring any entry in configuration file?

docs/assumptions-and-invariants.md
hwayne: Here's my attempt at explaining the same:

> The `ASSUME` (or `ASSUMPTION`) construct in TLA+ is used to introduce constant-level (level 0) formulas (formulas that don't contain variables) as constraints. TLC checks all `ASSUME` constraints before generating initial states.
>
> `ASSUME` is mainly used to constrain the possible values of constants, for example to ensure that `NumThreads > 1` or `Readers \intersect Writers = {}`.

I also think it's worth pointing out that you can put `TLCSet` and `PrintT` in ASSUME statements but not sure how to best communicate that

lemmy: Auxiliary information like `TLCSet` and `PrintT` better goes into extended articles that we host at `explain.tlapl.us` and link to from the TLC message.  Example below:

https://github.com/tlaplus/tlaplus/blob/f910c54bb1da33fd415933b0a04f94bca36ed3a2/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java#L926

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
lemmy: This would require a separate PR that adds an `index.md` file at the root of the `docs/` directory. We’d also need to remember to update `index.md` whenever new articles are added.

docs/assumptions-and-invariants.md
lemmy: I believe mentioning `TLCSet` and `PrintT` is beyond the scope of this help file. However, I anticipate that we will eventually have links between help files and pointers to background reading material.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
lemmy: https://github.com/tlaplus/tlaplus/issues/1266

lemmy:
Thank you all for your feedback.

PR #1264, opened 2025-11-25
Restrict macOS signing to trusted PRs

younes-io:
**Purpose**: Gate the macOS signing steps in `pr.yml` so they only run on trusted sources (non-fork PRs or branches in the main repo).

## Rationale
Forked PRs don’t have access to the Apple signing secrets, so the macOS signing job fails and wastes CI time. This change prevents the signing steps from running when secrets aren’t available. Signing will remain unchanged for trusted PRs.

## Notes
- No other workflow behavior is affected; only the macOS signing steps are guarded.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -112,19 +112,20 @@ jobs:
       ## Sign Toolbox macOS zip file.
       ##
     - name: Set up Apple Certs
-      if: matrix.operating-system == 'macos-latest'
+      # Forked PRs do not have access to macOS signing secrets.
+      if: matrix.operating-system == 'macos-latest' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
       run: 'echo "$APPLE_CODESIGN_CERTS" > certs.pem'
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -112,19 +112,20 @@ jobs:
       ## Sign Toolbox macOS zip file.
       ##
     - name: Set up Apple Certs
-      if: matrix.operating-system == 'macos-latest'
+      # Forked PRs do not have access to macOS signing secrets.
+      if: matrix.operating-system == 'macos-latest' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
```
lemmy, line 116: Can `github.event_name != 'pull_request'`  be replaced with `github.event_name == 'push'?

younes-io: TThey aren't equivalent here. This workflow only runs on pull_request, so swapping to github.event_name == 'push' would make the mac signing steps never runat all in this file. With the current guard we allow trusted PRs and any other future non-PR events we might add

PR #1265, opened 2025-11-26
Improve error messages when action-level formulas are used as PROPERTY.

lemmy:
TLC has always prevented action-level formulas from being used as temporal properties in the configuration file. This commit significantly improves the error messages to better explain why such usage is incorrect and how to fix it.

Introduced two new error codes (TLC_CONFIG_PROPERTY_ACTION_LEVEL and TLC_CONFIG_PROPERTY_ACTION_LEVEL_AND_STUTTERING_SENSITIVE) with detailed explanations. The messages clarify that using an action-level formula as a PROPERTY would only check it on the initial action, which is typically not the intended behavior. They guide users to wrap action formulas with the □ (always) operator and, when necessary, make them stuttering insensitive using the [A]_v notation.

Added comprehensive test cases covering various action-level formula scenarios, including stuttering-sensitive and stuttering-insensitive cases.

[Feature][TLC]

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
@@ -1297,6 +1297,28 @@ public class MP
                     + "\nbut TLC can only handle behaviors of length up to 65535 states. The last\n"
                     + "state in the behavior is:\n%1%");
         	break;
+        case EC.TLC_CONFIG_PROPERTY_ACTION_LEVEL_SQUARE_A_SUB_V:
+            b.append("The formula %1% at %2% is an action-level formula (i.e., it contains no temporal operators). "
+            		+ "Only temporal-level (or state-level) formulas are allowed under PROPERTY or PROPERTIES. "
+            		+ "To check that action %1% holds in *every* step of the behavior, define a temporal property "
+            		+ "by applying the \"always\" temporal operator ([]) to the formula at %2% (compare page 90 of "
+            		+ "Specifying Systems at https://lamport.azurewebsites.net/tla/book.html).");
+            break;
+        case EC.TLC_CONFIG_PROPERTY_ACTION_LEVEL_ANGLE_A_SUB_V:
+            b.append("The formula %1% at %2% is an action-level formula (i.e., it contains no temporal operators). "
+            		+ "Only temporal-level (or state-level) formulas are allowed under PROPERTY or PROPERTIES. "
+            		+ "To check that infinitely many %1% steps occur, define a temporal property by applying "
+            		+ "the \"always eventually\" temporal operator combination ([]<>) to the formula at "
+            		+ "%2% (compare page 91 of Specifying Systems at https://lamport.azurewebsites.net/tla/book.html).");
+            break;
+        case EC.TLC_CONFIG_PROPERTY_ACTION_LEVEL:
+            b.append("The formula %1% at %2% is an action-level formula (i.e., it contains no temporal operators). "
+            		+ "Only temporal-level (or state-level) formulas are allowed under PROPERTY or PROPERTIES. "
+            		+ "To check that action %1% holds in *every* step of the behavior, define a temporal property: "
+            		+ "Let A be the formula at %2%; rewrite A as [][A]_v, where v is a state function (typically "
+            		+ "the tuple of variables) (compare page 90 of Specifying Systems at "
+            		+ "https://lamport.azurewebsites.net/tla/book.html).");
+            break;
         case EC.TLC_CONFIG_PROPERTY_NOT_CORRECTLY_DEFINED:
             b.append("The property %1% is not correctly defined.");
             break;
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java
@@ -1691,8 +1691,25 @@ public class SpecProcessor implements ValueConstants, ToolGlobals {
         {
             this.impliedTemporalVec.addElement(new Action(Specs.addSubsts(pred, subs), c));
             this.impliedTemporalNameVec.addElement(name);
+        } else if (level == 2)
+        {
+			if (pred instanceof OpApplNode
+					&& BuiltInOPs.getOpCode(((OpApplNode) pred).getOperator().getName()) == OPCODE_sa) {
+				// PROPERTY A where formula A == [ ... ]_v 
+				Assert.fail(EC.TLC_CONFIG_PROPERTY_ACTION_LEVEL_SQUARE_A_SUB_V,
+						new String[] { name, pred.getLocation().toString() });
+			} else if (pred instanceof OpApplNode
+					&& BuiltInOPs.getOpCode(((OpApplNode) pred).getOperator().getName()) == OPCODE_aa) {
+				// PROPERTY A where formula A == << ... >>_v 
+				Assert.fail(EC.TLC_CONFIG_PROPERTY_ACTION_LEVEL_ANGLE_A_SUB_V,
+						new String[] { name, pred.getLocation().toString() });				
+			} else {
+				Assert.fail(EC.TLC_CONFIG_PROPERTY_ACTION_LEVEL,
+						new String[] { name, pred.getLocation().toString() });
+        	}
         } else
         {
+        	// There is no other level beyond 0..3, but let's keep this branch to be on the safe side.
             Assert.fail(EC.TLC_CONFIG_PROPERTY_NOT_CORRECTLY_DEFINED, name);
         }
     }
--- a/tlatools/org.lamport.tlatools/src/tlc2/output/EC.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/output/EC.java
@@ -321,6 +321,9 @@ public interface EC
     public static final int TLC_CANT_HANDLE_CONJUNCT = 2239;
     public static final int TLC_CANT_HANDLE_TOO_MANY_NEXT_STATE_RELS = 2240;
     public static final int TLC_CONFIG_PROPERTY_NOT_CORRECTLY_DEFINED = 2241;
+    public static final int TLC_CONFIG_PROPERTY_ACTION_LEVEL = 2272;
+    public static final int TLC_CONFIG_PROPERTY_ACTION_LEVEL_SQUARE_A_SUB_V = 2273;
+    public static final int TLC_CONFIG_PROPERTY_ACTION_LEVEL_ANGLE_A_SUB_V = 2274;
     public static final int TLC_CONFIG_OP_ARITY_INCONSISTENT = 2242;
     public static final int TLC_CONFIG_NO_STATE_TYPE = 2243;
     public static final int TLC_CANT_HANDLE_REAL_NUMBERS = 2244;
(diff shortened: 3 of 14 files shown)


lemmy:
@craft095 @hwayne @muenchnerkindl If possible, please review the following TLC change. The main purpose of this update is to revise the documentation to improve the user experience; the accompanying Java changes are secondary and can largely be ignored.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -1296,6 +1296,31 @@ else if (parameters.length == 2) {
...
+            		+ "not what is intended. For this reason, TLC does not support checking action-level "
+            		+ "formulas as properties.\n"
+            		+ "If your intention is to check that the action formula %1% holds on every step of every "
+            		+ "behavior, you should wrap it with the \"always\" operator (□), which asserts that %1% holds "
+            		+ "in all steps of all behaviors. Additionally, the formula %1% is stuttering sensitive, which "
+            		+ "TLA does not allow. To make it stuttering insensitive, you must wrap it in [%1%]_v, where "
+            		+ "v is an appropriate state-level predicate. Thus, the property should likely be written as "
+            		+ "□[%1%]_v.");
+            break;
+        case EC.TLC_CONFIG_PROPERTY_ACTION_LEVEL:
```
craft095: Why do we need two separate error messages? I thought that if `A` were an action, then TLC would reject `[]A` regardless of whether it is stutter-sensitive.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -1296,6 +1296,31 @@ else if (parameters.length == 2) {
...
         	break;
+        case EC.TLC_CONFIG_PROPERTY_ACTION_LEVEL_AND_STUTTERING_SENSITIVE:
+            b.append("The formula %1% is an action-level formula (i.e., it contains no temporal operators), "
+            		+ "but it was used as a PROPERTY, where a temporal formula is normally required. "
+            		+ "Interpreting an action-level formula as a PROPERTY amounts to asserting that it holds "
+            		+ "only for the first step (i.e., the initial action) of every behavior, which is typically "
+            		+ "not what is intended. For this reason, TLC does not support checking action-level "
+            		+ "formulas as properties.\n"
+            		+ "If your intention is to check that the action formula %1% holds on every step of every "
+            		+ "behavior, you should wrap it with the \"always\" operator (□), which asserts that %1% holds "
```
craft095: Could we use ASCII version for `always` or at least mention it? I.e. `[]` instead of `□`. Otherwise, new users might find it difficult to follow this advice.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -1296,6 +1296,31 @@ else if (parameters.length == 2) {
                     + "\nbut TLC can only handle behaviors of length up to 65535 states. The last\n"
                     + "state in the behavior is:\n%1%");
         	break;
+        case EC.TLC_CONFIG_PROPERTY_ACTION_LEVEL_AND_STUTTERING_SENSITIVE:
+            b.append("The formula %1% is an action-level formula (i.e., it contains no temporal operators), "
+            		+ "but it was used as a PROPERTY, where a temporal formula is normally required. "
```
muenchnerkindl: Why "normally"? To me this sounds like in certain cases other kinds of formulas might be acceptable. (Remember that any state formula is also a temporal formula.) I suggest simply removing "normally".

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -1296,6 +1296,31 @@ else if (parameters.length == 2) {
...
+            		+ "Interpreting an action-level formula as a PROPERTY amounts to asserting that it holds "
+            		+ "only for the first step (i.e., the initial action) of every behavior, which is typically "
+            		+ "not what is intended. For this reason, TLC does not support checking action-level "
+            		+ "formulas as properties.\n"
+            		+ "If your intention is to check that the action formula %1% holds on every step of every "
+            		+ "behavior, you should wrap it with the \"always\" operator (□), which asserts that %1% holds "
+            		+ "in all steps of all behaviors.");
+            break;
```
muenchnerkindl, line 1321: Same comment here.

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java
```diff
@@ -1691,8 +1691,23 @@ private final void processConfigProps(String name, ExprNode pred, Context c, Lis
         {
             this.impliedTemporalVec.addElement(new Action(Specs.addSubsts(pred, subs), c));
             this.impliedTemporalNameVec.addElement(name);
+        } else if (level == 2)
+        {
+			if (pred instanceof OpApplNode
+					&& BuiltInOPs.getOpCode(((OpApplNode) pred).getOperator().getName()) == OPCODE_sa) {
+        		// [A]_v is a stuttering insensitive action.
```
muenchnerkindl: This is of course just a comment, but I don't understand what you mean. [A]_v is true of any stuttering step, but may be true or false of non-stuttering steps, so it is stuttering sensitive.

lemmy: @muenchnerkindl Are you suggesting to remove the special handling of `[A]_v` and instead emit the same warning as below?

muenchnerkindl: Indeed, I wonder if the distinction is really useful.

hwayne: *Specifying Systems* uses the term "invariant under stuttering" to mean a formula who's validity is unaffected by stuttering steps (pg 90), and goes on to say that `[][A]_v` is always invariant under stuttering. That's how I read "stuttering insensitive".

I think we don't need special handling for `A` vs `[A]_v`, we just need to tell people to use `[][A]_v` in all cases. Something like

>  For this reason, TLC does not support checking action-level formulas as properties.\n
>
> If you need to check that action A holds in every step of the behavior, you need to write it as `[][A]_vars`. See [LINK] for more info.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
hwayne: I think writing the ASCII is preferable to the unicode, or at least having both `(□/[])`

hwayne:
Is it possible to extend this to state predicates, too? Right now using an action-level formula `(x' > x)` as a PROPERTY just gives a confusing error, but using a state predicate `(x > 0)` as a PROPERTY is an active footgun: it still model-checks, but only looks at the predicate for the initial state. This might be a separate PR though!

lemmy:
> Is it possible to extend this to state predicates, too? Right now using an action-level formula `(x' > x)` as a PROPERTY just gives a confusing error, but using a state predicate `(x > 0)` as a PROPERTY is an active footgun: it still model-checks, but only looks at the predicate for the initial state. This might be a separate PR though!

This was already done in an earlier commit (4b84a3b24adcb5eb613ea2b9e6b068f0c99274eb) :-)

https://github.com/tlaplus/tlaplus/blob/f910c54bb1da33fd415933b0a04f94bca36ed3a2/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java#L918-L927

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java
lemmy: My rationale for treating `[A]_v` separately is essentially the same “foot-gun” that @hwayne mentions regarding state-level properties. The key difference is that TLC simply does **not** support checking `[A]_v` on the initial steps of a behavior. This is a mere technical limitation: there is no meaningful use case for evaluating a *state-change* predicate on the very first action, whereas evaluating a plain state predicate *is* meaningful when checking refinement.

Formulas such as `x' > x` are not valid TLA properties because they are **not stuttering-insensitive**, as per definition in *Specifying Systems*. Because of this, I provided a dedicated error message to help users understand the requirement of stuttering insensitivity.

If you feel the error messages above are too dense and we want a more concise one like the one proposed by @hwayne, the longer explanation could go into an explain.tlapl.us article.

muenchnerkindl: Not sure if I understand your argument about this being a technical limitation. As you say, there is no meaningful use case for checking if formulas such as ``[A]_v`` or ``x' > x`` hold for the first two states of a behavior. Neither of these is a TLA property (= temporal formula).

lemmy: We agree that both constructs are action-level. However, `[A]_v` is stuttering-invariant: inserting any number of stuttering steps between the initial state and its successors does not affect its validity. This is not true for action-level formulas like `x' > x`. Because of this, we don’t want to allow users to check formulas such as `x' > x` at all, whereas checking `[A]_v` could be allowed—though there is no practical use case for doing so.

muenchnerkindl: How is that? ``[A]_v`` might be false for the first two states of the behavior but inserting a stuttering step makes it true.

lemmy: @muenchnerkindl Thanks for catching my mistake!

Future reader: See third paragraph of p. 90 in Specifying Systems.

lemmy:
1. For `<<...>>_v`, the user is advised to rewrite it as `[]<><<...>>_v`
2. For `[...]_v`, the user is advised to rewrite it as `[][...]_v`
3. For any other action-level formula A, the user is advised to rewrite it as `[][A]_v`


1. "Error: The formula ActionLivePropE at line 32, col 5 to line 32, col 17 of module ActionLevelProp is an action-level formula (i.e., it contains no temporal operators). Only temporal-level (or state-level) formulas are allowed under PROPERTY or PROPERTIES. To check that infinitely many ActionLivePropE steps occur, define a temporal property by applying the "always eventually" temporal operator combination ([]<>) to the formula at line 32, col 5 to line 32, col 17 of module ActionLevelProp (compare page 91 of Specifying Systems at https://lamport.azurewebsites.net/tla/book.html)."

2. "Error: The formula ActionLivePropD at line 29, col 5 to line 29, col 15 of module ActionLevelProp is an action-level formula (i.e., it contains no temporal operators). Only temporal-level (or state-level) formulas are allowed under PROPERTY or PROPERTIES. To check that action ActionLivePropD holds in *every* step of the behavior, define a temporal property by applying the "always" temporal operator ([]) to the formula at line 29, col 5 to line 29, col 15 of module ActionLevelProp (compare page 90 of Specifying Systems at https://lamport.azurewebsites.net/tla/book.html)."

3. "Error: The formula ActionLivePropA at line 20, col 5 to line 20, col 10 of module ActionLevelProp is an action-level formula (i.e., it contains no temporal operators). Only temporal-level (or state-level) formulas are allowed under PROPERTY or PROPERTIES. To check that action ActionLivePropA holds in *every* step of the behavior, define a temporal property: Let A be the formula at line 20, col 5 to line 20, col 10 of module ActionLevelProp; rewrite A as [][A]_v, where v is a state function (typically the tuple of variables) (compare page 90 of Specifying Systems at https://lamport.azurewebsites.net/tla/book.html)."

PR #1267, opened 2025-11-26
Make MP formatting thread-safe

younes-io:
- Replace shared `SimpleDateFormat/DecimalFormat` with thread-safe alternatives (`DateTimeFormatter` + `ThreadLocal<DecimalFormat>`), preventing races in multi-threaded TLC output
- Preserve existing formatting (pattern `yyyy-MM-dd HH:mm:ss`, grouping commas) & switching to `Instant` and system default zone for timestamps

I ran the tests locally with `ant -f tlatools/org.lamport.tlatools/customBuild.xml test-set -Stest.testcases="tlc2/output/MPTest.java"` and they passed

image

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
@@ -6,8 +6,9 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.io.Writer;
 import java.text.DecimalFormat;
-import java.text.SimpleDateFormat;
-import java.util.Date;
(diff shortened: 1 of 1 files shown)


ahelwer:
What do existing races in thread output look like?

younes-io:
I have not reproduced the output, but I can write a java class to try to reproduce this.

lemmy (approved):
LGTM overall. That said, it would be helpful to understand the scenario that led to the interleaved output, so we can verify whether this is purely a stdout/stderr ordering issue or if there might be a deeper underlying problem.

PR #1276, opened 2026-01-10
Fix debugger expression evaluation for scoped identifiers

Qian-Cheng-nju:
## Problem

When evaluating expressions in the TLC debugger (e.g., hovering over a variable
or using the debug console), scoped identifiers like operator parameters, LET
bindings, and quantified variables fail with "identifier X is undefined".

## Root Cause

`TLCDebuggerExpression.process()` generates a temporary module with new
`FormalParamNode` instances. However, `Context.lookup()` uses object identity
(`==`) rather than name matching, so lookups fail because the new nodes differ
from the original `SymbolNode` objects in the stack frame's context.

## Solution

Build a new evaluation context using name-based lookup, consistent with the
existing approach in `TLCSourceBreakpoint.matchesExpression()`.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCStackFrame.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/debug/TLCStackFrame.java
@@ -556,7 +556,16 @@ public class TLCStackFrame extends StackFrame {
 		Variable variable;
 		try {
 			variable = tool.eval(() -> {
-				return getVariable(tool.eval(odn.getBody(), getContext(), getS(), getT(), EvalControl.Clear),
+				// Create evaluation context with parameter bindings from the current context.
+				// This mirrors how TLCSourceBreakpoint.matchesExpression() handles parameter binding.
+				// For generated debug expressions (from TLCDebuggerExpression), the parameters
+				// represent scoped identifiers (operator params, LET bindings, quantified vars)
+				// that need to be resolved from the stack frame's context.
+				Context evalCtxt = Context.Empty;
+				for (FormalParamNode p : odn.getParams()) {
+					evalCtxt = evalCtxt.cons(p, getContext().lookup(sn -> sn.getName().equals(p.getName())));
+				}
+				return getVariable(tool.eval(odn.getBody(), evalCtxt, getS(), getT(), EvalControl.Clear),
 						odn.getName());
 			});
 		} catch (TLCRuntimeException | EvalException | FingerprintException e) {


lemmy:
@Qian-Cheng-nju Thanks for the PR! We should add a test case for this. Could you please provide a minimal example (spec) that reproduces the issue?

lemmy:
Hovering already works for certain LET/IN constructs and quantifiers.

Screenshot 2026-01-12 at 7 10 19 AM

https://github.com/tlaplus/Examples/blob/3673694ba1809ae072139e0b3ef0231c3e5d2f62/specifications/ewd998/EWD998ChanID.tla#L128-L148

Qian-Cheng-nju:
## Minimal Example to Reproduce the Issue

### Test.tla

```tla
---- MODULE Test ----
EXTENDS Naturals
VARIABLE v

Op(i) == i + 1

Init == v = 0
Next == v' = Op(v)
Spec == Init /\ [][Next]_v
====
```

### Test.cfg

```
SPECIFICATION Spec
```

### Steps to Reproduce

1. Set a breakpoint at line `Op(i) == i + 1`
2. Start the TLC debugger
3. When the breakpoint is hit, open the **Debug Console**
4. Type expressions and observe the results

### Result

Screenshot 2026-01-13 at 17 06 02

As shown in the screenshot:
- `v` → `0` (state variable resolves correctly)
- `1 + 1` → `2` (constant expression evaluates correctly)
- `i` → `i` (should return `0`, the current value of the operator parameter)

State variables and constant expressions work, but **scoped identifiers** (operator parameters, LET bindings, quantified variables) fail to resolve.

---

### Additional Finding

While investigating this issue, I noticed that `TLCStackFrame.getWatch(String name)` calls `TLCDebuggerExpression.process()` **without passing the `location` parameter**:

```java
// Line 545 in TLCStackFrame.java
odn = TLCDebuggerExpression.process(tool.getSpecProcessor(), module, name);
```

This means `getScopedIdentifiers()` is never called, so the generated operator has no parameters. Should we also fix this to pass the location?

```java
Location location = node != null ? node.getLocation() : Location.nullLoc;
odn = TLCDebuggerExpression.process(tool.getSpecProcessor(), module, location, name);
```

lemmy:
@Qian-Cheng-nju PR https://github.com/tlaplus/tlaplus/pull/1278 changes the TLA+ Debugger so that it correctly resolves scoped identifiers, such as `i` in the example above.

image

Qian-Cheng-nju:
Thanks for PR #1278! After further investigation, I found that: The `getWatch()` method we modified doesn't appear to be called.                                                                      

Your fix in #1278 solves the issue we encountered. Closing this PR as it's no longer needed.

PR #1278, opened 2026-01-13
Fix debugger expression evaluation for scoped identifiers.

lemmy:
Addresses issue reported in Github PR #1276
https://github.com/tlaplus/tlaplus/pull/1276

[Bug][TLC][Debugger]


lemmy:
The build failure is unrelated to this PR and is instead caused by https://github.com/tlaplus/Examples/pull/196

PR #1284, opened 2026-01-22
Fix equality checks on polymorphic sequences inside functions

edwardcwang:
Prevent errors like the below during liveness checking of specs using polymorphic sequences.

Attempted to check equality of string "U_OF_UNIT" with non-string

Works with Apalache 0.52.1.

Testcase:

```tla
\* tla2tools.jar -config PolymorphicSeq.cfg -workers 8 PolymorphicSeq.tla
\* apalache-mc check --debug --write-intermediate PolymorphicSeq.tla

------------------------------- MODULE PolymorphicSeq -------------------------------
EXTENDS Integers, Sequences, Variants

\* Bug reproduction: "Attempted to check equality of integer 1 with non-integer: "U_OF_UNIT""
\* The bug requires mixing UNIT and integer variants with extendDomain

\* Message types - need both UNIT and integer variants to trigger the bug
FirstVariantUnit == Variant("FirstVariant", UNIT)

\* @typeAlias: myVariant = FirstVariant(UNIT) | SecondVariant(Int);
TYPES == TRUE

\* Critical component - the bug requires extendDomain
\* @type: ((a -> b), a, b) => (a -> b);
extendDomain(f, key, value) == [x \in (DOMAIN f) \cup {key} |-> IF x = key THEN value ELSE f[x]]

VARIABLES
  \* Polymorphic queue
  \* @type: Int -> Seq($myVariant);
  queue

Init == queue = extendDomain([ k \in {} |-> <<>> ], 100, <<>>)

\* Mix of UNIT and integer variants
Next ==
  \E msgChoice \in 1..2:
    LET msg == CASE msgChoice = 1 -> FirstVariantUnit
                 [] msgChoice = 2 -> Variant("SecondVariant", 1)
    IN
    queue' = [queue EXCEPT ![100] = Append(@, msg)]

Spec == Init /\ [][Next]_queue

\* Force equality check with UNIT variants
MyInvariant ==
  \A k \in DOMAIN queue:
    \A i \in 1..Len(queue[k]):
      LET msg == queue[k][i] IN
      \/ msg = FirstVariantUnit
      \/ VariantTag(msg) = "SecondVariant"

```

TLC's `.cfg` file:

```
SPECIFICATION Spec
INVARIANT
  MyInvariant
```

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/value/impl/RecordValue.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/value/impl/RecordValue.java
@@ -354,6 +354,15 @@ public class RecordValue extends Value implements FunctionValue {
       }
       // Then, compare values iff domains are equal.
       for (int i = 0; i < len; i++) {
+        // If neither value is a ModelValue, ensure they have the same type
+        // to prevent Assert.fail() in downstream equals() methods
+        // This can happen e.g. with polymorphic records during liveness checking.
+        if (!(this.values[i] instanceof ModelValue || rcd.values[i] instanceof ModelValue)) {
+          if (!this.values[i].getClass().equals(rcd.values[i].getClass())) {
+            return false;
+          }
+        }
+
         if (!(this.values[i].equals(rcd.values[i]))) {
         	  return false;
         }


tlatools/org.lamport.tlatools/src/tlc2/value/impl/RecordValue.java
```diff
@@ -354,6 +354,15 @@ public final boolean equals(Object obj) {
       }
       // Then, compare values iff domains are equal.
       for (int i = 0; i < len; i++) {
+        // If neither value is a ModelValue, ensure they have the same type
+        // to prevent Assert.fail() in downstream equals() methods
+        // This can happen e.g. with polymorphic records during liveness checking.
+        if (!(this.values[i] instanceof ModelValue || rcd.values[i] instanceof ModelValue)) {
+          if (!this.values[i].getClass().equals(rcd.values[i].getClass())) {
+            return false;
```
Calvin-L, line 362: Although this patch fixes your particular spec, it won't work in general. There are two problems.

First: equal values can be represented by different classes. The same TLA+ function can be represented as either a `FcnRcdValue` or a `FcnLambdaValue`. So this might return `false` when the values are actually equal!

Second, and more subtly: TLA+ values are not like values in ordinary programming languages. In TLA+, equality can be ambiguous! The axioms of the logic do not tell us whether `1 = <<0, 0>>` is true or false. In TLC at run-time, the left side will be an `IntValue` and the right side will be a `TupleValue`---but we have to be careful not to return a definitive true or false answer when asked if they are equal.

----

So what's the right thing to do?

One fix would be to change the definitions in Apalache's [`Variants` module](https://github.com/apalache-mc/apalache/blob/main/src/tla/Variants.tla) to make them more TLC-friendly. For instance, this definition of `Variant` should work with TLC (I've used something like it in my own work):

```tla
EXTENDS TLC \* for :>
Variant(__tagName, __value) ==
    __tagName :> __value
```

Because now two different variants will have distinct domains, and TLC can therefore prove them distinct without ever inspecting the actual values.

If we want to make a fix on the TLC side, there's a somewhat general limitation of TLC where it throws exceptions comparing certain values that _are_ provably distinct. See e.g. the tuple case at the bottom of https://github.com/tlaplus/tlaplus/issues/776#issuecomment-1374321278:

> it cannot prove that these two tuples are different:
>     `<<..., 10, ..., "step2", ...>>`
>     `<<..., "10", ..., "step3", ...>>`

A general fix for the tuple case (and the record case in this issue) would look something like

```
Throwable exception = null
for each (a, b) in values to compare:
    try:
        if !a.equals(b):
             # it doesn't matter if some previous checks were ambiguous---
             # the two structures are DEFINITELY provably distinct
            return false
    catch e:
        # equality is ambiguous---but we'll keep going in case some later
        # checks prove the values are actually distinct
        if exception == null:
            exception = e
if exception != null:
    throw exception
return true
```

...but it's a lot of added complexity (and maybe run-time overhead!) to fix a relatively infrequent case.

We would also have to be careful about which exceptions we catch, since we don't (yet) have a dedicated exception type for "ambiguous equality".

lemmy: I agree that Apalache's [Variants module](https://github.com/apalache-mc/apalache/blob/main/src/tla/Variants.tla) should be made more TLC-friendly.

/cc @konnov

edwardcwang: @Calvin-L @lemmy Thank you so much for looking into it! As a user of TLA+/TLC/Apalache, it would be great if we could at least have a user experience where the same spec should work out of the box with both TLC and Apalache.

konnov: We could change the Variants module, for sure. But first I would like to understand why TLC is failing on that. I thought TLC is ok with values of any type?

konnov: Actually, looking at the current definition of `Variant`:

```tla
Variant(__tagName, __value) ==
    \* default untyped implementation
    [ tag |-> __tagName, value |-> __value ]
```

Is not it the standard pattern that appears in Paxos, Raft, etc.?

Calvin-L: TL;DR: It's a common pattern, it works in TLAPS, and it _usually_ works in TLC, but it is not guaranteed to work in TLC for all possible values.

It's been a while since I looked at those specs, but I believe they avoid this limitation because they use model values, or non-model-values that TLC considers comparable (like records with different sets of fields).

In some cases (which I don't have links to, sorry) Apalache's pattern works entirely by accident, based on TLC's arbitrary strategy for sorting strings. One creative and ugly "fix" for the `PolymorphicSeq` spec above is to add this definition immediately after the `EXTENDS` clause and before anything else:

```tla
tag == "tag"
```

Such definitions can incidentally change how TLC sorts strings. In this case, TLC will assign an earlier sort order to the string "tag", so when it compares records it will compare their tag fields first, and will never actually compare their values if the tags are different. (This behavior is _not_ guaranteed and _very_ brittle and _nobody should use it_, but it works in this example.)

konnov: @Calvin-L Many thanks for a clear explanation! As far as I understand your proposed solution of using `__tagName :> __value` works since TLC would find out that the domains `{"FirstVariant"}` and `{"SecondVariant"}` are different and would not go into further value comparison.

So most likely this approach should also work without requiring us to pull TLC as a dependency:

```tla
Variant(__tagName, __value) ==
    \* default untyped implementation
    [ x \in { __tagName } |-> __value ]
```

konnov: https://github.com/apalache-mc/apalache/pull/3255 should fix it

lemmy:
Closing this given https://github.com/apalache-mc/apalache/pull/3255

PR #1285, opened 2026-01-23
Support replaying a trace with `-loadTrace` that was previously created with `-dumptrace`.

lemmy:
Support -loadTrace command-line parameter to replay previously dumped traces.

This change introduces a new `-loadTrace <format> <filename>` command-line parameter that enables TLC to replay execution traces that were previously saved using the `-dumpTrace <format> <filename>`command.

The implementation adds trace loading functionality by registering a dynamically generated constraint that compares each state during model checking against the corresponding state from the loaded trace file. Both TLC's internal binary format and JSON format are supported, providing flexibility for different use cases and integration scenarios.

Supported formats:
- tlc:  TLC's internal binary format (fast, compact, portable)
- json: Human-readable JSON format (portable, debuggable)

Usage examples:
```bash
    tlc -dumpTrace tlc error.tlc Spec.tla
    tlc -loadTrace tlc error.tlc Spec.tla
    tlc -dumpTrace json error.json Spec.tla
    tlc -loadTrace json error.json Spec.tla
```

[Feature][TLC]

Depends on: https://github.com/tlaplus/tlaplus/pull/1288

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_TLCTrace.tla
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_TLCTrace.tla
@@ -33,6 +33,10 @@ _TLCTrace ==
 LOCAL _TLCTraceFileDeserialized ==
     _TLCTraceDeserialize(_TLCTraceFile)
 
+\* This operator has a Java module override (tlc2.module._TLCTrace#tlcState).
+LOCAL _TLCState(level) ==
+	Trace[level]
+
 LOCAL _TLCTraceConstraint ==
     LET level == TLCGet("level")
         dump  == _TLCTraceFileDeserialized
@@ -43,6 +47,6 @@ LOCAL _TLCTraceConstraint ==
 	\* is intentionally vacuously satisfied.
 	\* Since the names of the spec’s variables are not known, Trace[level] is
 	\* used as a generic reference to the variables of the current state.
-    IN level \in DOMAIN dump => Trace[level] = dump[level]
+    IN level \in DOMAIN dump => _TLCState(level) = dump[level]
 
 =============================================================================
--- a/tlatools/org.lamport.tlatools/src/tlc2/module/_TLCTrace.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/module/_TLCTrace.java
@@ -28,13 +28,20 @@ package tlc2.module;
 import java.io.File;
 import java.io.IOException;
 
+import tla2sany.semantic.ExprOrOpArgNode;
+import tlc2.overrides.Evaluation;
 import tlc2.overrides.TLAPlusOperator;
+import tlc2.tool.TLCState;
+import tlc2.tool.coverage.CostModel;
+import tlc2.tool.impl.Tool;
+import tlc2.util.Context;
 import tlc2.value.IValue;
 import tlc2.value.ValueInputStream;
 import tlc2.value.ValueOutputStream;
 import tlc2.value.impl.BoolValue;
 import tlc2.value.impl.RecordValue;
 import tlc2.value.impl.StringValue;
+import tlc2.value.impl.Value;
 import util.UniqueString;
 
 public class _TLCTrace {
@@ -75,5 +82,16 @@ public class _TLCTrace {
 		} finally {
 			vis.close();
 		}
+	}
+
+	@Evaluation(definition = "_TLCState", module = "_TLCTrace", warn = false, silent = true, minLevel = 1)
+	public static Value tlcState(final Tool tool, final ExprOrOpArgNode[] args, final Context c,
+			final TLCState s0, final TLCState s1, final int control, final CostModel cm) {
+		// Overrides the TLA+ definition: _TLCState(level) == Trace[level]
+		// The use of TLCExt!Trace is inefficient because it internally reconstructs all
+		// states starting from the initial state. As a result, its cost grows linearly
+		// with the length of the trace. This override directly returns the state at the
+		// current level without reconstruction.
+		return new RecordValue(s0);
 	}
 }


Qian-Cheng-nju:
Hi @lemmy,                                                                                                                                             
                                                                                                                                                         
Thanks for this feature! I noticed that the implementation uses a constraint to compare states against the loaded trace:                               
                                                                                                                                                         
IN level \in DOMAIN dump => Trace[level] = dump[level]                                                                                                 
                                                                                                                                                         
This requires an **exact match** of all variables.                                                                                                     
                                                                                                                                                         
I'm curious: have you considered supporting traces where **only a subset of variables are recorded**? This would enable a "constrained exploration" mode — where TLC explores freely for unrecorded variables while following the trace for recorded ones.                                                 
                                                                                                                                                         
If this were supported, it could significantly simplify **trace validation** workflows. Instead of writing a dedicated Trace Spec to map system logs to TLA+ states, we could simply have the system output traces in a specified format (recording only observable variables), and let TLC replay them directly against the original spec.                                                                                                                    

Is this something you'd consider, or are there fundamental limitations that make partial-variable traces impractical?

lemmy:
In practice, we have encountered several kinds of *partial logs* when performing trace validation:

1. **Incomplete variable valuations**
   The log does not contain values for all variables.

2. **Imprecise variable information**
   The log records only ranges (rather than exact values) for some variables.

3. **Incomplete action information**
   The log lacks action names entirely, or includes action names but omits action parameters.

4. **Mismatched [grain of atomicity](https://arxiv.org/pdf/2404.16075) between the log and the specification**

   * **4a)** Logs that omit some states of the behavior (which is effectively equivalent to missing values for all variables in those states).
   * **4b)** Logs that include *lower-level* states that must be skipped during validation.
   * **4c)** Logs that contain *extra* states that must be merged (\cdot) with predecessor or successor states.

5. **Causality violations**
   The log is not in causal order.

6. **Combinations of the above**
   Real logs often exhibit several of these issues simultaneously.

Designing a generic `-loadtrace` constraint capable of handling all of these cases is not feasible. However, if you typically encounter *correctly ordered logs that match the specification’s grain of atomicity and only omit some variable values*, then the infrastructure introduced in this PR does make it straightforward to add support for an additional `loadtrace` format that applies those logs.

Related, https://github.com/tlaplus/Examples/blob/master/specifications/ewd998/EWD998ChanTrace.tla uses the constraint-based approach to trace validation. Notably, even for this moderately complex system, the resulting trace-validation specification is non-trivial.

tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_TLCTrace.tla
```diff
@@ -3,18 +3,46 @@ LOCAL INSTANCE TLC
...
+    LET level == TLCGet("level")
+        dump  == _TLCTraceFileDeserialized
+	\* For liveness properties, TLC trace dumps stop at the state *before* the
+	\* lasso is closed. When replaying such a trace, TLC may request a state with
+	\*   level = Len(dump) + 1,
+	\* which does not exist in the dump. In that case, the constraint
+	\* is intentionally vacuously satisfied.
+	\* Since the names of the spec’s variables are not known, Trace[level] is
+	\* used as a generic reference to the variables of the current state.
+    IN level \in DOMAIN dump => Trace[level] = dump[level]
```
lemmy: The use of TLCExt!Trace is inefficient because it internally reconstructs all states starting from the initial state. As a result, its cost grows linearly with the length of the trace.

This can be addressed with a Java override (`@Evaluation`).

lemmy:
> [...] if you typically encounter _correctly ordered logs that match the specification’s grain of atomicity and only omit some variable values_, then the infrastructure introduced in this PR does make it straightforward to add support for an additional `loadtrace` format that applies those logs.

The "log" defines the set of variables:

```diff
diff --git a/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_TLCTrace.tla b/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_TLCTrace.tla
index f14f95016..215ddd5f6 100644
--- a/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_TLCTrace.tla
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_TLCTrace.tla
@@ -18,7 +18,9 @@ CONSTANT _TLCTraceFile

 LOCAL _TLCTrace0(verbose) ==
     IF CounterExample.state = {} \/ ("console" \in DOMAIN CounterExample /\ CounterExample["console"] = FALSE) THEN TRUE ELSE
-        /\ _TLCTraceSerialize(ToTrace(CounterExample), _TLCTraceFile)
+        /\ LET trace == ToTrace(CounterExample)
+               vars  == UNION { DOMAIN trace[i] : i \in DOMAIN trace }
+           IN _TLCTraceSerialize([trace |-> trace, vars |-> vars], _TLCTraceFile)
         /\ IF verbose THEN PrintT("CounterExample written: " \o _TLCTraceFile) ELSE TRUE

 LOCAL _TLCTraceSilent ==
@@ -40,6 +42,8 @@ LOCAL _TLCState(level) ==
 LOCAL _TLCTraceConstraint ==
     LET level == TLCGet("level")
         dump  == _TLCTraceFileDeserialized
+        trace == dump["trace"]
+        vars  == dump["vars"]
        \* For liveness properties, TLC trace dumps stop at the state *before* the
        \* lasso is closed. When replaying such a trace, TLC may request a state with
        \*   level = Len(dump) + 1,
@@ -47,6 +51,6 @@ LOCAL _TLCTraceConstraint ==
        \* is intentionally vacuously satisfied.
        \* Since the names of the spec’s variables are not known, Trace[level] is
        \* used as a generic reference to the variables of the current state.
-    IN level \in DOMAIN dump => _TLCState(level) = dump[level]
+    IN level \in DOMAIN trace => \A v \in vars: _TLCState(level)[v] = trace[level][v]

 =============================================================================
```

tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_TLCTrace.tla
lemmy: https://github.com/tlaplus/tlaplus/pull/1285/changes/e8aabe58f015f3d87db6e2b088d39bce95de57f7

lemmy:
Follow-up at https://github.com/tlaplus/tlaplus/pull/1289

Qian-Cheng-nju:
Got it. Thanks for the detailed explanation, this is very helpful. The traces in [Specula](https://github.com/specula-org/Specula) (our tool) usually have good properties and strictly defined formats, so I think I can build support for our specific format on top of this PR.

I also saw you CC'd me on #1289 — thanks! Thanks for addressing the subset of variables case I asked about—this is exactly what I was looking for!

lemmy:
> Got it. Thanks for the detailed explanation, this is very helpful. The traces in [Specula](https://github.com/specula-org/Specula) (our tool) usually have good properties and strictly defined formats, so I think I can build support for our specific format on top of this PR.
> 
> I also saw you CC'd me on #1289 — thanks! Thanks for addressing the subset of variables case I asked about—this is exactly what I was looking for!

Could you confirm whether the (binary) tlc format works for your use case (Specula)?

Qian-Cheng-nju:
Thanks for asking!

No, Specula does not use TLC's binary format — we use NDJSON (newline-delimited JSON), a text-based format.

The reason we chose a human-readable format is that new version Specula incorporates an AI-assisted workflow for debugging and fixing trace validation failures. When validation fails, AI agents need to inspect and understand the trace content to diagnose issues and suggest fixes to the spec. A readable text format (JSON/NDJSON) makes this feasible, whereas reconstructing traces from the binary format would add significant complexity. Additionally, text-based formats are easier to build tooling around (parsing, filtering, etc.).

Given this context, I'd love to know if the "subset of variables" support (in PR #1289) will also work with the JSON format introduced in PR #1285? If not (or if it currently only targets binary), we'd be happy to explore implementing the support for our specific format ourselves based on the infrastructure.

Thanks again for your work on these features!

PR #1288, opened 2026-01-26
Prevent PostCondition from executing multiple times with concurrent errors

lemmy:
When multiple workers find counterexamples concurrently, only the first worker should execute PostCondition. This ensures only one counterexample is serialized with `-dumpTrace`, matching the counterexample printed to stdout.

Technically, `POSTCONDITION` is a constant-level expression and evaluating it multiple times should not matter because it is idempotent. However, it gets used to trigger side effects like serializing a counterexample, which is not idempotent. The fact that `TLCExt!CounterExample` exists arguably implies that conceptually `POSTCONDITION` is temporal-level.

[Bug][TLC]

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/Worker.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/Worker.java
@@ -611,10 +611,15 @@ public final class Worker extends IdThread implements IWorker, INextStateFunctor
 	
 	private boolean doNextSetErr(TLCState curState, TLCState succState, boolean keep, int ec, String param) throws IOException, WorkerException {
 		synchronized (this.tlc) {
+			final boolean wasAlreadyDone = this.tlc.done;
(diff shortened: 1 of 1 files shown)


lemmy:
This change does not violate any guarantees about `POSTCONDITION` that are explicitly stated anywhere. However, it is still a behavioral change, and it goes against how `POSTCONDITION` has been used in the wild. For example, the [SimKnuthYao](https://github.com/tlaplus/Examples/blob/master/specifications/KnuthYao/SimKnuthYao.tla) specification expects `POSTCONDITION` to be evaluated independently for each simulation worker.
Therefore, rather than preventing `POSTCONDITION` from being evaluated once per (simulation and BFS) worker, it would be less invasive to restrict `-dumptrace` so that it does not override the counterexample written to stdout when more than one counterexample exists.

lemmy:
Related https://github.com/tlaplus/tlaplus/pull/1218

PR #1293, opened 2026-02-02
fix: Allow simultaneous use of -loadTrace and -dumpTrace options

Qian-Cheng-nju:
## Problem

When using `-loadTrace` and `-dumpTrace` together, TLC fails:

```bash
java -jar tla2tools.jar -loadTrace json trace.json -dumpTrace json output.json <spec>.tla
# Error: output.json (No such file or directory)
```

**Root cause:** Both options share the same constant (`_JsonTraceFile`), so `-dumpTrace` overwrites `-loadTrace`'s value, causing the constraint to read from the non-existent output file.

## Solution

Separate the constants:
- `_JsonTraceInputFile` / `_TLCTraceInputFile` — for `-loadTrace`
- `_JsonTraceOutputFile` / `_TLCTraceOutputFile` — for `-dumpTrace`

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_JsonTrace.tla
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_JsonTrace.tla
@@ -4,14 +4,15 @@ LOCAL INSTANCE TLCExt
 LOCAL INSTANCE Json
 LOCAL INSTANCE Sequences
 
-CONSTANT _JsonTraceFile
+CONSTANT _JsonTraceOutputFile  \* Used by -dumpTrace json
+CONSTANT _JsonTraceInputFile   \* Used by -loadTrace json
 
 _JsonTrace ==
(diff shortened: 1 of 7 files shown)


lemmy:
@Qian-Cheng-nju I’m not sure I understand the use case here. Loading a trace and then dumping it again would simply reproduce the same trace without any observable change.

Qian-Cheng-nju:
@lemmy  The purpose is not to reproduce the same trace, but to observe additional information during replay. We have an existing trace (the execution history), and we want to observe or compute things that weren't captured in the original trace (debugging).             
                                                                                                                                                         
For example, we might want to evaluate derived expressions like x + y or Len(log), filter the view to only show specific variables like {currentTerm, state}, or check custom invariants at each state. -loadTrace replays the trace, ALIAS defines what to observe, and -dumpTrace outputs the observations.

If my understanding is incorrect in any way, please feel free to point it out. Thank you!

lemmy:
https://github.com/tlaplus/tlaplus/pull/1297

PR #1297, opened 2026-02-03
Allow simultaneous use of `-loadTrace` and `-dumpTrace` options.

lemmy:
Enable a trace replay workflow in TLC. Users can now load an existing trace with `-loadTrace`, replay it in TLC while applying a new ALIAS, and dump the resulting derived trace in the same run.

This allows TLC to act as a single-step trace post-processing or conversion tool by supporting `-loadTrace` and `-dumpTrace` together.

The constant name `_JsonTraceFile` remains unchanged to preserve backward compatibility.

Part of Github PR #1293
https://github.com/tlaplus/tlaplus/pull/1293

[Feature][TLC]

### DIFF
--- a/tlatools/org.lamport.tlatools/test/tlc2/debug/Debug05SimTest.java
+++ b/tlatools/org.lamport.tlatools/test/tlc2/debug/Debug05SimTest.java
@@ -99,8 +99,8 @@ public class Debug05SimTest extends TLCDebuggerTestCase {
 			ea.setFrameId(stackFrame.getId());
 
 			String traceFile = "Debug05SimTest.json";
-			ea.setExpression(
-					"LET J == INSTANCE _JsonTrace WITH _JsonTraceFile <- \"" + traceFile + "\" IN J!_JsonTrace");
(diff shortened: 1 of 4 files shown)


tlatools/org.lamport.tlatools/test/tlc2/debug/Debug05SimTest.java
```diff
@@ -99,8 +99,8 @@ public void testSpec() throws Exception {
 			ea.setFrameId(stackFrame.getId());
 
 			String traceFile = "Debug05SimTest.json";
-			ea.setExpression(
-					"LET J == INSTANCE _JsonTrace WITH _JsonTraceFile <- \"" + traceFile + "\" IN J!_JsonTrace");
+			ea.setExpression("LET J == INSTANCE _JsonTrace WITH _JsonTraceFile <- \"" + traceFile
+					+ "\", _JsonTraceInputFile <- \"" + traceFile + "\" IN J!_JsonTrace");
```
lemmy, line 103: @Qian-Cheng-nju This duplication and the assumption that users would not combine `-dumptrace` and `-loadtrace` (for the TLC format, such a combination does not make sense), was the reason why I originally avoided introducing two separate constants.  That said, your use case involving the JSON module is strong.

PR #1301, opened 2026-02-09
Remove RMI dependency from FPSet for CheerpJ compatibility

FedericoPonzi:
FPSet extended UnicastRemoteObject which triggers native RMI library loading on construction. CheerpJ (Java-in-browser via WebAssembly) does not support native RMI sockets.

Change `FPSet` to implement `Serializable` instead of extending `UnicastRemoteObject`. The `unexportObject` method becomes a no-op. This only affects distributed TLC which is on path of [deprecation](https://github.com/tlaplus/tlaplus/issues/718) anyways.

More info:
`FPSet` extended UnicastRemoteObject, which is an RMI class. On construction,  UnicastRemoteObject automatically calls `System.loadLibrary("rmi") `to load the native RMI library and exports the object for remote access over the network.
CheerpJ runs Java in WebAssembly in the browser — it has no native RMI library and no network sockets. So every time TLC created an `FPSet`, it crashed with `UnsatisfiedLinkError: no rmi in java.library.path`.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/fp/FPSet.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/fp/FPSet.java
@@ -6,9 +6,7 @@
 package tlc2.tool.fp;
 
 import java.io.IOException;
-import java.rmi.NoSuchObjectException;
 import java.rmi.RemoteException;
-import java.rmi.server.UnicastRemoteObject;
 
 import tlc2.tool.TLCTrace;
 import tlc2.tool.distributed.fp.DistributedFPSet;
@@ -23,7 +21,7 @@ import tlc2.util.LongVec;
  * guarantee that their methods are thread-safe.
  */
 @SuppressWarnings("serial")
-public abstract class FPSet extends UnicastRemoteObject implements FPSetRMI
+public abstract class FPSet implements FPSetRMI
 {
 	/**
 	 * Size of a Java long in bytes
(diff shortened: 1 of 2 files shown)


tlatools/org.lamport.tlatools/src/tlc2/tool/fp/FPSet.java
```diff
@@ -23,7 +22,7 @@
  * guarantee that their methods are thread-safe.
  */
 @SuppressWarnings("serial")
-public abstract class FPSet extends UnicastRemoteObject implements FPSetRMI
+public abstract class FPSet implements FPSetRMI, Serializable
```
Calvin-L: Why add `implements Serializable` here?

FedericoPonzi: It's to minimize disruption as UnicastRemoteObject implements Serializable, see: https://docs.oracle.com/javase/8/docs/api/java/rmi/server/UnicastRemoteObject.html. If you think it can be removed I'm also fine to remove it.

Calvin-L: My intuition says `Serializable` can (and should) be removed: `Serializable` is meant for things sent over the network, and the FPSet itself is never sent in its entirety. `FPSet` implements `Remote` (via `FPSetRMI`), which should be enough.

PR #1295, opened 2026-02-02
Add comprehensive corpus test for XMLExporter to validate TLA+ module exports.

lemmy:
Several tests are failing, indicating potential bugs in the XMLExporter. A thorough analysis is required.

[Tests][XML]

Issue #1072, opened 2024-11-13
Proposal: TLC/SANY quiet mode


hwayne:
When I run a large model, the TLC output always starts like this:

```
TLC2 Version 2.20 of Day Month 20?? (rev: 639382e)
Running breadth-first search Model-Checking with fp 99 and seed 154168001562069844 with 24 workers on 24 cores with 14517MB heap and 64MB offheap memory [pid: 19524] (Windows 11 10.0 amd64, Oracle Corporation 19.0.2 x86_64, MSBDiskFPSet, DiskStateQueue).
Starting SANY...
Parsing file [path to file]
Parsing file [path to file]
Parsing file [path to file]
Parsing file [path to file]
Parsing file [path to file]
Parsing file [path to file]
Semantic processing of module Naturals
Semantic processing of module Sequences
Semantic processing of module FiniteSets
Semantic processing of module TLC
Semantic processing of module Integers
Semantic processing of module Reals
SANY finished.
Starting... (2024-11-13 10:11:49)
Computing initial states...
Computed 2 initial states...
Computed 4 initial states...
Computed 8 initial states...
Computed 16 initial states...
Computed 32 initial states...
Computed 64 initial states...
Computed 128 initial states...
Computed 256 initial states...
Finished computing initial states: 484 distinct states generated at 2024-11-13 10:11:50.
```

It'd be convenient if there was a flag to suppress this initial input and only show success or error traces.

lemmy:
The `fp`, `seed`, ... should probably be shows as part of success or error trace if this gets implemented.

lemmy:
Related because also control over output: https://github.com/tlaplus/tlaplus/issues/1186

lemmy:
@craft095 This appears partially related to your work on https://github.com/tlaplus/tlaplus/pull/1344

PR #1303, opened 2026-02-10
XML Exporter: silence PlusCal logs

ahelwer:
Add global static method to modify PlusCal output
Currently, some PlusCal log messages (like "Labels added.") are output to stdout when parsing a module containing PlusCal in preparation for exporting it as XML. Since the XML is also output to stdout, this causes the XML to be invalid. This is a quick hack fix to expose the ability to silence PlusCal logs by routing them to a null output stream.

Related: #1072, #1186, #1295

### DIFF
--- a/tlatools/org.lamport.tlatools/src/pcal/PcalDebug.java
+++ b/tlatools/org.lamport.tlatools/src/pcal/PcalDebug.java
@@ -1,5 +1,6 @@
 package pcal;
 
+import java.io.PrintStream;
 import java.util.Vector;
 
 import pcal.exception.UnrecoverableException;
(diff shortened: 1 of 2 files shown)


lemmy:
Why does invoking XMLExporter trigger the PlusCal translator?

ahelwer:
SANY always runs PlusCal to check whether the PlusCal input has changed relative to the output, and print a warning if so. It does not re-generate the translation, just checks it.

lemmy:
Running the translator as part of SANY makes sense for most use cases. However, XMLExporter is primarily intended as a programmatic tool for building IDE integrations and other downstream tooling. In these scenarios, invoking the translator is not always desirable.

Regardless of the specific use case, XMLExporter should not have a dependency on (i.e., should not import) the translator.

ahelwer:
So would you prefer (1) injecting SANY's output infrastructure into PlusCal's translator so its logging is subject to the same log level filtering as SANY or (2) adding a SANY flag to skip PlusCal validation

lemmy:
Both approaches are reasonable, but in the context of the XMLExporter, I prefer approach (2).

ahelwer:
Superseded by #1305

PR #1305, opened 2026-02-11
XML Exporter: skip linting & PlusCal validation in SANY

ahelwer:
Expose a new SanySettings class for better programmatic control of SANY

Previously, programs running SANY had two options: they could either accept SANY's default settings, or call it with string command line settings. This work adds an additional SanySettings class which programmatic users can specify to customize SANY's runtime behavior at a finer grain than previously. This was motivated by a desire for the XML Exporter to skip PlusCal validation.

Additional minor change(s):
 - Removed the invariant doParsing global static setting
 - Throw `IllegalArgumentException` in `tla2sany.output` class constructors if `PrintStream` arguments are null

[SANY][Feature]

Supersedes #1303 

Fixes issues with XML Exporter found in #1295

Somewhat related to the goal of #891, although the global static variables in question are all private so there was no contagion to the rest of the codebase.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
@@ -62,10 +62,6 @@ public class SANY {
   public static String version = 
     "Version 2.2 created " + modDate ; 
 
-  private static boolean doParsing          = true;  
-    // true <=> parsing should be done
-    //          currently there is no way to turn this off
-
   private static boolean doSemanticAnalysis = true;  
     // true <=> semantic analysis should be done
     //          default is true; turned OFF by -S switch
@@ -88,9 +84,16 @@ public class SANY {
     //          return value
 
   /**
-   * Use {@link SANY#frontEndMain(SpecObj, String, SanyOutput)} instead for
-   * greater control of output. This method will print all human-readable log
-   * messages to the provided syserr parameter. It cannot be null.
+   * The default setting of whether to validate the root module's PlusCal
+   * translation and warn if it is out of sync.
+   */
+  private static final boolean doValidatePCalTranslationDefault = true;
+
+  /**
+   * Use {@link SANY#frontEndMain(SpecObj, String, SanyOutput, SanySettings)}
+   * instead for greater control of output and settings. This method will
+   * print all human-readable log messages to the provided syserr parameter.
+   * The syserr parameter cannot be null.
    */
   @Deprecated
   public static final int frontEndMain(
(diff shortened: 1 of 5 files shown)


tlatools/org.lamport.tlatools/src/tla2sany/drivers/SanySettings.java
```diff
@@ -0,0 +1,131 @@
...
+public final class SanySettings {
+
+  /**
+   * By default, SANY always returns 0 regardless of whether any stage of the
+   * parsing process succeeded or failed. To maintain backwards compatibility
+   * this setting was introduced, which if set to true makes SANY return a
+   * nonzero error code on failure. External users of this class will always
+   * have this setting set to true.
+   */
+  final boolean StrictErrorCodes;
```
lemmy: Please keep member variables lowercase to be consistent with most of the other code.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -139,21 +131,39 @@ public static final int frontEndMain(
...
-      if (doSemanticAnalysis) 
-            {frontEndSemanticAnalysis(spec, out, doLevelChecking);} ;
-
-      // **** Semantic analysis and level checking
-      if (doLinting) 
-            {frontEndLinting(spec, out); } ;
-
+      if (settings.DoSemanticAnalysis) {
+        frontEndSemanticAnalysis(spec, out, settings.DoLevelChecking);
+        if (settings.DoLevelChecking && settings.DoLinting) {
```
lemmy: Why is level checking a prerequisite for linting?

ahelwer: Do you think it should be? Probably some lints would use level information.

lemmy: I don’t believe the current linter requires level checking. In any case, behavior changes like this belong in their own commit.

ahelwer:
Got a bit too far ahead of myself refactoring the SANY CLI arg parsing code so that has been removed/simplified to avoid reproducing the conditions that led to the SANY output refactoring error.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -88,9 +84,16 @@ public class SANY {
...
+   * The default setting of whether to validate the root module's PlusCal
+   * translation and warn if it is out of sync.
+   */
+  private static final boolean doValidatePCalTranslationDefault = true;
+
+  /**
+   * Use {@link SANY#frontEndMain(SpecObj, String, SanyOutput, SanySettings)}
+   * instead for greater control of output and settings. This method will
+   * print all human-readable log messages to the provided syserr parameter.
+   * The syserr parameter cannot be null.
```
lemmy, line 96: Throw IllegalArgumentExp in SimpleSanyOutput ctor?

ahelwer:
It occurs to me that since we are creating a new version of the `SANY.frontEndMain()` method and deprecating the existing ones, and this is probably the single function that programmatic SANY users will call, we should take the opportunity to rename it to something more descriptive. @Calvin-L @FedericoPonzi @lemmy thoughts? How about something simple, like just `SANY.parse()`?

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -88,9 +84,16 @@ public class SANY {
...
    */
   @Deprecated
   public static final int frontEndMain(
```
lemmy, line 99: One of the key lessons learned from the legacy codebase is that too many components were implemented as static. If this is intended to be a new API, it would be better to avoid making it static and instead design it with greater flexibility and testability in mind.

ahelwer: Certainly, we have [this non-static API](https://github.com/tlaplus/tlaplus/blob/6fde5a9ea3967191118ce18b27878a9b4b24f77f/tlatools/org.lamport.tlatools/test/tla2sany/api/Frontend.java) prototyped in the tests directory, although in this case we have the opportunity to do something more incremental and just rename the new method instead of keeping its name as `SANY.frontEndMain()`.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -138,20 +160,21 @@ public static final int frontEndMain(
...
       // **** Semantic analysis and level checking
-      if (doSemanticAnalysis) 
-            {frontEndSemanticAnalysis(spec, out, doLevelChecking);} ;
+      if (settings.doSemanticAnalysis) 
+            {frontEndSemanticAnalysis(spec, out, settings.doLevelChecking);} ;
 
-      // **** Semantic analysis and level checking
-      if (doLinting) 
+      // **** Linting
+      if (settings.doLinting)
```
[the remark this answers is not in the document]
ahelwer: Will be fixed in later commit

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -138,20 +160,21 @@ public static final int frontEndMain(
   public static final int frontEndMain(
                              SpecObj spec, 
                              String fileName, 
-                             SanyOutput out) throws FrontEndException {
+                             SanyOutput out,
+                             SanySettings settings) throws FrontEndException {
```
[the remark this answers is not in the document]
ahelwer: Fixed by updating method documentation

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: Absent additional naming input I have called this method `SANY.parse()`

PR #1307, opened 2026-02-12
XML Exporter: expose proof level

ahelwer:
TLAPM needs the the level of various proof steps. These changes expose that info in the parse tree & export it in the XML. This PR is split into two commits, the first of which modifies SANY's parser code (along with tests) and the second of which (much simpler) modifies the XML export code and its format. The commit messages are reproduced here:

> SANY: expose proof level info in TreeNode
    TLAPM needs proof level info exposed in SANY's XML AST export. To that end, these changes modify the TreeNode interface to include a getProofLevel() method which returns the proof level for N_Proof nodes. The syntax parser is modified to attach the proof level to the SyntaxTreeNode class when appropriate. Tests were added to check this new functionality.

> XML Exporter: expose proof level
    The proof level will now be appended to non-leaf proofs

As always, ignore most of the auto-generated changes in `TLAplusParserGenerator.java` (moving `jj_`-prefixed methods around, etc.). The `tla+.jj` file is the real source of these changes, although it never hurts to peer through the resulting changes in the actual Java code. The correspondence between `tla+.jj` and `TLAplusParser.java` is checked by the CI.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/xml/sany.xsd
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/xml/sany.xsd
@@ -619,6 +619,7 @@
           <xs:complexType>
             <xs:sequence>
               <xs:group ref="node"/>
+              <xs:element name="proofLevel" type="xs:integer"/>
               <xs:sequence maxOccurs="unbounded">
                 <xs:group ref="step"/>
               </xs:sequence>
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/NonLeafProofNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/NonLeafProofNode.java
@@ -9,6 +9,7 @@ import org.w3c.dom.Element;
 
 import tla2sany.explorer.ExploreNode;
 import tla2sany.explorer.ExplorerVisitor;
+import tla2sany.st.SyntaxTreeConstants;
 import tla2sany.st.TreeNode;
 import tla2sany.utilities.Strings;
 import tla2sany.utilities.Vector;
@@ -225,6 +226,8 @@ public class NonLeafProofNode extends ProofNode {
   protected Element getLevelElement(Document doc, SymbolContext context, BiPredicate<SemanticNode, SemanticNode> filter) {
     Element e = doc.createElement("steps");
 
+    assert(this.getTreeNode().isKind(SyntaxTreeConstants.N_Proof));
+    e.appendChild(appendText(doc, "proofLevel", Integer.toString(this.getTreeNode().getProofLevel())));
     for (int i=0; i< steps.length; i++) {
       e.appendChild(steps[i].export(doc,context, filter));
     }


tlatools/org.lamport.tlatools/src/tla2sany/xml/sany.xsd
```diff
@@ -619,6 +619,7 @@
           <xs:complexType>
             <xs:sequence>
               <xs:group ref="node"/>
+              <xs:element name="proofLevel" type="xs:integer"/>
```
lemmy, line 622: Is this optional? Does this need `minOccurs="0"` or whatever syntax marks it optional?

tlatools/org.lamport.tlatools/src/tla2sany/parser/SyntaxTreeNode.java
```diff
@@ -148,6 +148,16 @@ public class SyntaxTreeNode implements TreeNode, SyntaxTreeConstants,
     * post-comment.                                                        *
     ***********************************************************************/
 
+  /**
+   * The proof level of this {@link SyntaxTreeNode} instance, if applicable.
+   */
+  int proofLevel = -1;
```
lemmy: Unless javacc doesn't support it, this should be private and a setLevel be used above.

tlatools/org.lamport.tlatools/src/tla2sany/st/TreeNode.java
```diff
@@ -25,4 +25,12 @@ public interface TreeNode {
...
   public TreeNode getParent();
   public void setParent();
+
+  /**
+   * The proof level of this node, if applicable. Only defined for the
+   * {@link SyntaxTreeConstants#N_Proof} node kind; -1 otherwise.
+   *
+   * @return The proof level of the node; -1 if not applicable.
+   */
+  public int getProofLevel();
```
lemmy, line 35: Since `getProofLevel` is not applicable, returns `-1` for most `TreeNode` instances, and is only used within `ProofTests`, it should not be added directly to `TreeNode`. Instead, `ProofTests` should cast the instance to `SyntaxTreeNode` when calling `getProofLevel`.

tlatools/org.lamport.tlatools/src/tla2sany/xml/sany.xsd
ahelwer: No, it should be present on every list of proof steps.

tlatools/org.lamport.tlatools/src/tla2sany/st/TreeNode.java
ahelwer: It is also used by the XML export code in the proof class in the second commit; the relationship between the TreeNode and SyntaxTreeNode classes is a bit weird and lost to history, but it seems like TreeNode is just a way to free consumers of SyntaxTreeNode from having to look at all the implementation methods of SyntaxTreeNode. In that case does it still make sense to remove this method from TreeNode and require a downcast?

Issue #1302, opened 2026-02-10
TLC Value pepresentation bug in `FcnLambdaValue` causes invalid `EXCEPT` semantics, violating soundness and completeness


lemmy:
`FcnLambdaValue#toFcnRcd` not only mutates `FcnLambdaValue#fcnRcd` (which is intended), but also unintentionally mutates `FcnLambdaValue#excepts` as a side effect. The issue surfaced during liveness checking in simulation mode, but has since also been reproduced when checking safety properties in model-checking mode, to the point that TLC can result in both missed property violations and spurious (bogus) violations, making it a soundness and completeness problem.

The fix ensures that `toFcnRcd` [creates and operates on a copy of the excepts state](https://github.com/tlaplus/tlaplus/pull/1304/changes/5f6d2af986def0d012e1f63d00af4f4ece5dda1e) before converting the `FcnLambdaValue`, preventing unintended mutation. The fix is validated by [several unit tests](https://github.com/tlaplus/tlaplus/pull/1316/changes/605a32415f386f86f9a39147c1b1dd3ab71daa00), as well as [two](https://github.com/tlaplus/tlaplus/pull/1316/changes/931d7c1d0f82d7d7b404e745537f7893d85167de) [integration](https://github.com/tlaplus/tlaplus/pull/1316/changes/f219625a07bd5104e6b23684001a32d3f902179b) tests based on minimal reproducible specifications. These tests confirm that the mutation bug is fixed. Moreover, the integration tests serve as [small reproducible examples](https://github.com/tlaplus/tlaplus/pull/1316/changes/f219625a07bd5104e6b23684001a32d3f902179b) that allow users to assess whether they may have been affected by the bug.

We advise all users to reconfirm their TLC model checking results by verifying their specifications again with the [fixed version of TLC](https://github.com/tlaplus/tlaplus/releases/tag/v1.8.0).

While the severity is high, it may not be the highest: in most of the cases inspected, TLC crashes with a `ClassCastException` or triggers a bogus assertion error, which at least makes the failure visible. Moreover, I was unable to reproduce the issue in a specification with only a single variable. That said, one unit test demonstrates that the issue can silently produce incorrect fingerprint values for `FcnLambdaValue` instances, which is particularly concerning. The bug appears to have existed since the early days of TLC—it has been confirmed below in TLC 2.03 (as of 2011-03-10).

[...]

PR #1316, opened 2026-02-17
Soundness & Completeness bugs caused by`FcnLambdaValue#toFcnRcd` conversion.

lemmy:
To determine whether issue https://github.com/tlaplus/tlaplus/issues/1302 should be classified with the highest severity:
- [x] Reproduce the crash when model checking safety properties
- [x] Find a formula (input) where TLC completes without crashing but returns an incorrect result
- [x] Combine the previous two into single test

### DIFF
--- a/tlatools/org.lamport.tlatools/test-model/Github1302c.tla
+++ b/tlatools/org.lamport.tlatools/test-model/Github1302c.tla
@@ -1,0 +1,32 @@
+---- MODULE Github1302c ----
+EXTENDS TLC, TLCExt
+
+\* Non-Reducible domain so the outer function stays a lazy FcnLambdaValue.
+S == [{1} -> {1, 2}]
+I == [s \in S |-> 0]
+
+r1 == [i \in {1} |-> 1]
+r2 == [i \in {1} |-> 2]
+
+fcn == r1 :> 10
+
+VARIABLES x, y
+
+Init ==
+    /\ x = [I EXCEPT ![r1] = fcn] \* (<<1>> :> (<<1>> :> 10) @@ <<2>> :> 0)
+    /\ y = [x EXCEPT ![r2] = 99]  \* (<<1>> :> (<<1>> :> 10) @@ <<2>> :> 99)
+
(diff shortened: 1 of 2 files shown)


lemmy:
Alerting @Calvin-L to help assess the severity of issue #1302 in light of the provided test cases. Issue #1302 can result in both missed property violations and spurious (bogus) violations, making it a soundness and completeness problem. 

While the severity is high, it may not be the highest: in most of the cases inspected, TLC crashes with a `ClassCastException` or triggers a bogus assertion error, which at least makes the failure visible. Moreover, I was unable to reproduce the issue in a specification with only a single variable. That said, one unit test demonstrates that the issue can silently produce incorrect fingerprint values for `FcnLambdaValue` instances, which is particularly concerning. The bug appears to have existed since the early days of TLC—it has been confirmed below in TLC 2.03 (as of 2011-03-10).

An announcement similar to https://discuss.tlapl.us/msg05355.html is warranted.



---------


```shell
markus@avocado [10:26:47] [~/src/TLA/tla/tlatools/org.lamport.tlatools] [mku-gh1302]
-> % java -cp ~/TLC/tla2tools_1-3-1.jar tlc2.TLC Github1302b.tla
TLC2 Version 2.03 of 10 March 2011
Running in Model-Checking mode.
Parsing file Github1302b.tla
Semantic processing of module Github1302b
Starting... (2026-02-18 10:26:49)
Computing initial states...
Finished computing initial states: 1 distinct state generated.
Model checking completed. No error has been found.
  Estimates of the probability that TLC did not check all reachable states
  because two distinct states had the same fingerprint:
  calculated (optimistic):  val = 5.4E-20
  based on the actual fingerprints:  val = 1.1E-19
2 states generated, 1 distinct states found, 0 states left on queue.
The depth of the complete state graph search is 1.
Finished. (2026-02-18 10:26:49)
```

lemmy:
Merging this issue is because it only adds tests for a closed issue. The communication, similar to https://github.com/tlaplus/tlaplus/issues/798#issuecomment-1477101847, still needs to be done.

PR #1318, opened 2026-02-20
XML Exporter: don't dump exception if parsing fails

ahelwer:
Before these changes, the XML Exporter would dump an XMLExportingException right to the console if module parsing failed. These changes ensure that does not happened, so users simply see the usual human-friendly SANY parsing failure messages. It makes use of SANY returning a nonzero exit code on failure. These changes also fixed two code paths which did not appropriately set the SANY error level.

[SANY][Bugfix][Feature]

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
@@ -259,10 +259,12 @@ public class SANY {
       catch (ParseException e) 
       {
           // get here if either the TLAPlusParser.parse() threw a ParseException or spec.ParseErrors was not empty
+          spec.errorLevel = 2;
           throw new ParseException();
       }
       catch (Exception e) 
       {
+          spec.errorLevel = 2;
           out.log(LogLevel.ERROR, "\nFatal errors while parsing TLA+ spec in file %s\n", spec.getFileName());
           out.log(LogLevel.ERROR, e.toString());
           out.log(LogLevel.ERROR, spec.parseErrors.toString());
--- a/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
@@ -106,8 +106,7 @@ public class XMLExporter {
 
   public static int run(String... args) throws XMLExportingException {
     try {
-      moduleToXML(args);
-      return 0;
+      return moduleToXML(args);
     } catch (IllegalArgumentException e) {
       ToolIO.err.println("ERROR: " + e.getMessage());
       printUsage(ToolIO.err);
@@ -115,7 +114,7 @@ public class XMLExporter {
     }
   }
 
-  static void moduleToXML(String... args) throws XMLExportingException {
+  static int moduleToXML(String... args) throws XMLExportingException {
 
     if (args.length < 1) throw new IllegalArgumentException("at least one .tla file must be given");
     LinkedList<String> pathsLs = new LinkedList<>();
@@ -157,7 +156,7 @@ public class XMLExporter {
 
     if (args[args.length - 1].equals("-help")) {
         printUsage(ToolIO.out);
-        return;
+        return 0;
     }
 
     String tla_name = args[lastarg++];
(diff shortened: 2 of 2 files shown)


tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
```diff
@@ -177,15 +176,14 @@ static void moduleToXML(String... args) throws XMLExportingException {
...
-          throw new XMLExportingException("spec " + spec.getName() + " is malformed - does not have a root module", null);
+        if (spec.getErrorLevel() != 0) {
+          return spec.getErrorLevel();
+        }
       } catch (FrontEndException fe) {
         // For debugging
         fe.printStackTrace();
         ToolIO.err.println(fe);
-        return;
+        return 1;
```
lemmy, line 186: Why 1 and not 42?

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -259,10 +259,12 @@ public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validateP
       catch (ParseException e) 
       {
           // get here if either the TLAPlusParser.parse() threw a ParseException or spec.ParseErrors was not empty
+          spec.errorLevel = 2;
           throw new ParseException();
       }
       catch (Exception e) 
       {
+          spec.errorLevel = 2;
```
lemmy, line 267: Why 2 and not 123?

tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
ahelwer: In this case there is no particular reason behind the choice, other than that it is a nonzero value.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: This has an actual reason! You can see here that 2 is the error code for an error during the syntax parsing phase; however, some exceptions being thrown can bypass this being set:

https://github.com/tlaplus/tlaplus/blob/932a781c97f35f8c5cb294f159cc89ab3396b60e/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java#L254

lemmy: Consider turning 2 into a constant with a descriptive name.

tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
```diff
@@ -106,16 +106,15 @@ public static void main(String... args) throws XMLExportingException {
...
+      return moduleToXML(args);
     } catch (IllegalArgumentException e) {
       ToolIO.err.println("ERROR: " + e.getMessage());
       printUsage(ToolIO.err);
       return 1;
     }
   }
 
-  static void moduleToXML(String... args) throws XMLExportingException {
+  static int moduleToXML(String... args) throws XMLExportingException {
```
lemmy, line 117: `moduleToXML` still throws `XMLExportingException`. Rather than catching SANY’s exception inside `moduleToXML`, consider handling SANY’s exceptions in `run`, alongside the existing `IllegalArgumentException` handling.

ahelwer: My thinking here is that it is probably okay if the user sees an XMLExportingException, which would seem to indicate an actual bug in the XML Exporter which they could report. However, for the fairly common case where the user tries to XML-export a spec that doesn't parse, they'll just see the usual human-friendly SANY output instead of an XMLExportingException that doesn't give them any more information. If we want to catch the XMLExportingException in the `run()` method and print out its inner message instead of the stack trace that would also be a decent approach, though.

ahelwer: Oh I see, you are talking about using exceptions instead of returning an error code. I will check to see whether catching the FrontEndException even makes sense there; I suspect it cannot really be caught but will see.

lemmy: `moduleToXML` is currently invoked from main, but it should also remain callable programmatically. We should preserve that flexibility. Any conversion of exceptions to error codes should be handled at a higher level, outside of `moduleToXML`. Perhaps, it makes sense to wrap SANY's exception in `XMLExportingException`. We want to avoid ending up in a situation like SANY, where the functionality cannot be invoked programmatically.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: Done in PR #1319

ahelwer:
Closing this to revisit after #1319 is merged; might involve modularizing XML Exporter class

PR #1319, opened 2026-02-20
SANY: enumerate possible exit codes

ahelwer:
This work creates a new class called SanyExitCode which enumerates all the possible exit codes that can be returned by SANY, on the command line and from its main parsing methods. It modifies SANYExitException to hold this enumerated exit code instead of an int, and modifies all relevant places in the SANY frontend to use the enumerated exit codes instead of integer constants like -1, 0, 2, and 4. These changes also set the SpecObj.errorLevel field in two code paths where it likely unintentionally remained unset.

As suggested in PR #1318.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
@@ -264,10 +264,12 @@ public class SANY {
       catch (ParseException e) 
       {
           // get here if either the TLAPlusParser.parse() threw a ParseException or spec.ParseErrors was not empty
+          spec.errorLevel = SanyExitCode.SYNTAX_PARSING_FAILURE.code();
           throw new ParseException();
       }
       catch (Exception e) 
       {
+          spec.errorLevel = SanyExitCode.SYNTAX_PARSING_FAILURE.code();
           out.log(LogLevel.ERROR, "\nFatal errors while parsing TLA+ spec in file %s\n", spec.getFileName());
           out.log(LogLevel.ERROR, e.toString());
           out.log(LogLevel.ERROR, spec.parseErrors.toString());


tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -251,18 +252,19 @@ public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validateP
...
-              spec.errorLevel = 2;
+              spec.errorLevel = SanyExitCode.SYNTAX_PARSING_FAILURE.code();
               throw new ParseException(); 
           }
       }
-      // TODO
       catch (ParseException e) 
       {
           // get here if either the TLAPlusParser.parse() threw a ParseException or spec.ParseErrors was not empty
+          spec.errorLevel = SanyExitCode.SYNTAX_PARSING_FAILURE.code();
```
lemmy, line 267: This is new?

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -251,18 +252,19 @@ public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validateP
...
           throw new ParseException();
       }
       catch (Exception e) 
       {
+          spec.errorLevel = SanyExitCode.SYNTAX_PARSING_FAILURE.code();
```
lemmy, line 272: This is new?

lemmy:
> These changes also set the SpecObj.errorLevel field in two code paths where it likely unintentionally remained unset.

Could these two changes please be moved into a separate git commit?

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: This is the change cribbed from #1318. It's a bit of a weird case. Ultimately if you trace what happens when the ParseException is re-thrown right below this, it ends up being caught in the `SANY.parse()` method and `-1` is returned from that method. We could probably avoid making this change, although it makes more sense that `SpecObj.errorLevel` is set to a nonzero level when an exception like this occurs. The user might check `SpecObj.errorLevel` expecting it to be a general-purpose indicator of error, but here an error occurs and `SpecObj.errorLevel` remains 0 (without this change).

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: same as above

ahelwer:
Split the additional `spec.errorLevel` settings into a separate commit in this PR.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -143,20 +143,22 @@ public static final int frontEndMain(
...
-   * {@link SpecObj#getErrorLevel()} to see what level of problems, if any,
-   * occurred during the parsing phases. Any value other than 0 is fatal, and
+   * The spec object must be queried by the caller using {@link SpecObj#getErrorLevel()}
+   * to see what level of problems, if any, occurred during the parsing
+   * phases. Any value other than {@link SanyExitCode#OK} (0) is fatal, and
    * the caller should not use the spec object any more. To see the actual
    * list of failures tagged with an {@link tla2sany.semantic.ErrorCode},
    * call {@link SpecObj#getParseErrors()} and {@link SpecObj#getSemanticErrors()}.
+   * A {@link SanyExitCode} is also returned by this method. The spec object
+   * should only be used if it has value {@link SanyExitCode#OK}.
```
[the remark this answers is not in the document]
ahelwer: The error handling for this method is unfortunately quite complicated. I wrote a longer comment explaining it.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -178,20 +180,19 @@ public static final int parse(
 
     }
     catch (ParseException pe) {
-      return -1;
+      return SanyExitCode.ERROR;
     }
     catch (SemanticException se) {
-      return -1;
+      return SanyExitCode.ERROR;
     }
```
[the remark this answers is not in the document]
ahelwer: True, although unfortunately it must be kept this way for backwards compatibility reasons. Although this behavior really is so egregious that it might be worth breaking backwards compatibility to make error handling easier for callers of this method.

Issue #715, opened 2022-02-15
Variable cannot be used as a property


craft095:
Property **Prop** is not recognized as valid:

```
---- MODULE TLC_M0 ----
VARIABLE x
Init == x = TRUE
Next == x' = ~x
Spec == Init /\ [][Next]_x
Prop == x
====
```

Error: 
```
The operator x, which equals line 2, col 10 to line 2, col 10 of module TLC_M0,
cannot be used as a property
```

At the same time, if I replace `x` with `x = TRUE`, error disappears

```
---- MODULE TLC_M0 ----
VARIABLE x
Init == x = TRUE
Next == x' = ~x
Spec == Init /\ [][Next]_x
Prop == x = TRUE
====
```

zitro3:
I tried to change the Prop as temporal formula in the Property section of the ToolBox. I was not able to see the error (I am using 1.8 version of the ToolBox) and I see the "Invariant x is violated". If I change it to "Prop == [](x = TRUE)" I get "Invariant Prop is violated". I am not sure if this helps.

```tla
------------------------------- MODULE TLC_M0 -------------------------------
VARIABLE x
Init == x = TRUE
Next == x' = ~x
Spec == Init /\ [][Next]_x
Prop == [](x)

=============================================================================
\* Modification History
\* Last modified Fri Nov 17 15:43:34 PST 2023 by zitro
\* Created Fri Nov 17 15:41:04 PST 2023 by zitro
```

PR #1323, opened 2026-02-21
Allow variables as `PROPERTY` in Config

lemmy:
`SpecProcessor#processConfigProps` previously rejected state-level variable references (for example, `Prop == x` where x is a VARIABLE) with the error:

> "cannot be used as a property"

One might argue that TLC should generally reject state-level formulas when they appear as properties. However, a state-level formula is simply a degenerate action-level formula, and TLA permits such degenerate forms wherever the corresponding non-degenerate form is allowed.

Moreover, users may intentionally specify that a constant or state-level formula should hold only in the initial state. Interpreted this way, the formula expresses precisely that meaning.

In addition, a recent TLC enhancement now emits a warning to alert users to the subtle semantic implications of using state-level formulas as properties, helping prevent unintentional misuse while preserving expressive flexibility.

Addresses Github issue #715
https://github.com/tlaplus/tlaplus/issues/715

[Bug][TLC]


lemmy:
@craft095 Please review.

PR #1324, opened 2026-02-22
XmlExporter: enumerate built-in operator names in schema file

ahelwer:
The first commit in this PR also runs the exporter on all tlaplus/examples specs, which is a good test to see whether schema modifications introduce errors. The XML output is, of course, sent to /dev/null. stderr output will be maintained, though.

Also added a bunch of documentation about built-in operators to the schema file.

Ref #1313

Issue #1044, opened 2024-10-18
Refactoring goal: improve stability of liveness checker


ahelwer:
As discussed near the end of the last community meeting, the liveness checker is quite unstable in the sense that minor changes to apparently unrelated parts of the codebase can cause it to find alternative liveness counterexamples. This is not a soundness issue, but many liveness tests hardcode an expected counterexample so it makes it difficult to confidently change other parts of the codebase when these have a chance of breaking existing liveness checker tests. Since we want to unlink parts of the codebase in this way as much as possible, we should endeavor to improve the stability of the liveness checker.

Here is an example, which I unfortunately was able to reproduce at the time but now can no longer do so; if [this](https://github.com/tlaplus/tlaplus/blob/b5512af1c114048e8e633ead1797f4c2419eeea5/tlatools/org.lamport.tlatools/src/tla2sany/semantic/BuiltInOperators.java#L88) line of code is deleted, then this test (used to, at one time) fails:
```
ant -f customBuild.xml info compile compile-test dist test-set -Dtest.testcases=tlc2/tool/liveness/CodePlexBug08EWD840FL1Test.java
```
The test logs from the CI run are attached here: [testresults-ubuntu-latest.zip](https://github.com/user-attachments/files/17437917/testresults-ubuntu-latest.zip)

Now for some reason I cannot reproduce this failure. It is mysterious as to why; I was consistently able to reproduce it locally after the CI failed (see https://github.com/tlaplus/tlaplus/issues/1008#issuecomment-2369178596). But anyway this is just more evidence that this is a problem.

@lemmy I know you have been bitten by this phenomenon before and you know the liveness checker better than anybody. Do you know how we might be able to attack this problem?

lemmy:
There are at least two sources of non-determinism in TLC:

1) **String Internalization**: TLC internalizes strings in a non-deterministic manner. For instance, adding a definition in your specification can change the order in which strings are stored in TLC’s internal table. This affects the order in which TLC generates sets of states, potentially leading to different counterexamples during safety and liveness checks.

2) **Worker Thread Scheduling**: When multiple worker threads are involved, they compete to explore states in parallel. This introduces non-determinism in how states are processed, meaning that the order in which TLC finds counterexamples may vary depending on the thread execution schedule. This also affects the order in which TLC detects strongly connected components during liveness checks.

ahelwer:
[...]

PR #1331, opened 2026-02-25
SANY: remove unused built-in operators

### LINKED ISSUE #1008: Level parameters are not set for built-in dot infix operator (record access operator)

ahelwer:
Several operators in the tla2sany.semantic.BuiltInOperator table are not actually used:
 - . which is a duplicate of $RcdSelect
 - \neg which is a duplicate of \lnot
 - \times which is a duplicate of $CartesianProd
 - $TemporalWhile which is not used at all

These changes remove the unused operators from the table. Their inclusion resulted in some spurious code review comments in #1324. Last time I tried to do this it resulted in some spurious failures like #1044 where the specific counterexample found by the liveness checker somehow was affected by the number of operators in this table; we will see what happens now.

[SANY][Refactoring]

Closes #1008


ahelwer:
`tlc2.tool.liveness.CodePlexBug08EWD840FL1Test` indeed still fails from making this change, because it finds a different liveness counterexample than the one hardcoded. I can run this test from the command line using `java tlc2.TLC -deadlock -fp 0 test-model/CodePlexBug08/EWD840MC1` to reproduce the 10-step error trace (the `-fp 0` is critical to avoiding nondeterminism).

ahelwer:
The test failure occurs any time three or more entries are removed from the built-in operator table; the specific entries do not matter (as long as their absence does not cause a parsing failure). Thus we can infer that liveness checking is sensitive to the length of this table. Possibly this has to do with interning string constants in `UniqueString`.

lemmy:
This behavior can almost certainly be attributed to string interning. Instead of modifying the table, consider leaving it intact and simply adding comments to mark the obsolete entries.

ahelwer:
Claude hypothesizes it is due to this:
- The `Context` class loads these built-in operators into a fresh hashtable before processing a module; the `Context` class then adds various other operator definitions into this hashtable
- The hashtable resizes when a certain load factor is exceeded; this resize changes the `Hashtable.elements()` enumeration order
- By removing three elements from the built-in operators table, this resize could be made to not happen when it otherwise would have; thus the final iteration order changes
- This changed iteration order affects which index a spec variable is assigned
- Ultimately this changes the order of successor states in `SetOfStates`, thus changing the BFS exploration order
- (Not sure I follow this final step, I thought liveness checking is supposed to use the full state graph and not just the successor graph) this means a different liveness counterexample will be found

ahelwer:
Moving this to draft until I have time to trace what is actually causing the instability

PR #1335, opened 2026-02-27
BufferedDataInputStream#readString silently returns a NUL-padded partial string

lemmy:
Fix double-counting bug in BufferedDataInputStream#readString that silently returned corrupt data instead of throwing EOFException on truncated streams.

The inner loop updated off and n as it copied bytes. But three lines after the loop recomputed the same delta and applied it a second time. When readString(n) hit EOF after reading k bytes (where k >= n/2), the double subtraction made n go negative, causing the outer loop to exit instead of re-entering and throwing EOFException at the len < 0 check. The method silently returned a NUL-padded partial string.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/util/BufferedDataInputStream.java
+++ b/tlatools/org.lamport.tlatools/src/util/BufferedDataInputStream.java
@@ -68,6 +68,7 @@ public final class BufferedDataInputStream extends FilterInputStream implements 
 
     private void initFields() {
         this.buff = new byte[8192];
+        this.len = -1; // as per Object invariants above.
         this.curr = 0;
         this.temp = new byte[8];
     }
--- a/tlatools/org.lamport.tlatools/test/util/BufferedDataInputStreamTest.java
+++ b/tlatools/org.lamport.tlatools/test/util/BufferedDataInputStreamTest.java
@@ -32,8 +32,11 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.EOFException;
 import java.io.IOException;
+import java.io.InputStream;
 
 import org.junit.Test;
+
+import tlc2.output.EC;
 
 public class BufferedDataInputStreamTest {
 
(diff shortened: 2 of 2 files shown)


tlatools/org.lamport.tlatools/src/util/BufferedDataInputStream.java
```diff
@@ -267,7 +267,6 @@ public final String readString(int n) throws IOException {
       int off = 0;
       while (n > 0) {
 	if (this.len < 0) throw new EOFException();
```
craft095, line 270: If `this.len` is initially zero, this loop will spin forever. Even if it is impossible now, it might be in the future. To fix it's enough to change 
`if (this.len < 0) throw new EOFException();`
to
`if (this.len <= 0) throw new EOFException();`

lemmy: Not sure if it is true, but `this.len # 0` is supposedly an invariant

https://github.com/tlaplus/tlaplus/blob/572093f445a89d79ae2ceaa1c8650ea81b99a30b/tlatools/org.lamport.tlatools/src/util/BufferedDataInputStream.java#L35

lemmy: It's *not* true because the ctor and `initFields` do don't initialize `len` to e.g. `-1`.  Thus, it is default-initialized to `0`.

https://github.com/tlaplus/tlaplus/blob/572093f445a89d79ae2ceaa1c8650ea81b99a30b/tlatools/org.lamport.tlatools/src/util/BufferedDataInputStream.java#L27-L30

https://github.com/tlaplus/tlaplus/blob/572093f445a89d79ae2ceaa1c8650ea81b99a30b/tlatools/org.lamport.tlatools/src/util/BufferedDataInputStream.java#L49-L55

https://github.com/tlaplus/tlaplus/blob/572093f445a89d79ae2ceaa1c8650ea81b99a30b/tlatools/org.lamport.tlatools/src/util/BufferedDataInputStream.java#L69-L73

Issue #725, opened 2022-04-21
Undefined variable with WF in submodules (and SelectSeq)


villiros:
This spec

```tla
---- MODULE wf_seq_test ----
EXTENDS Sequences, Naturals

VARIABLES outerRes, outerSeq

---- MODULE inner ----
    VARIABLES res, myseq
    
    Step ==
        /\ res = 0
        /\ res' = 1
        /\ myseq' = SelectSeq(myseq, LAMBDA a: a /= res')
    
    Fairness == WF_<<res, myseq>>(Step)
====

Svc == INSTANCE inner WITH res <- outerRes, myseq <- outerSeq

Init ==
    /\ outerSeq = <<1, 2, 3>>
    /\ outerRes = 0

Step ==
    \/ Svc!Step
    \/ ~ENABLED Svc!Step /\ UNCHANGED <<outerRes, outerSeq>>

Spec ==
    /\ Init
    /\ [][Step]_<<outerRes, outerSeq>>
    \* This crashes
    /\ Svc!Fairness
    \* This works
    \* /\ WF_<<outerRes, outerSeq>>(Svc!Step)

SpecRunsToEnd ==
    /\ <>[](~ENABLED Svc!Step)

====
```

Crashes in the inner `Fairness` function with:

```
Attempted to apply the operator overridden by the Java method
public static tlc2.value.impl.Value tlc2.module.Sequences.SelectSeq(tlc2.value.impl.Value,tlc2.value.impl.Value),
but it produced the following error:

In evaluation, the identifier outerRes is either undefined or not an operator.
line 17, col 35 to line 17, col 42 of module wf_seq_test
```

Same happens if the inner module is in a separate file, I believe.

It appears to be something to do with the instance + WF + `SelectSeq` (or maybe `LAMBDA`?) when using a primed variable.

lemmy:
With the `Svc!Fairness`, TLC's `OpLambdaValue` corresponding to the `LAMBDA ...` in TLA+, refers to the `res` variable.  With `WF_<<outerRes, outerSeq>>(Svc!Step)`, the `OpLambdaValue` points to `outerRes`, for which TLC has a value.

lemmy:
[...]

PR #1342, opened 2026-03-09
Add tests for ENABLED soundness bug with INSTANCE substitution and Java

lemmy:
operator overrides.

When a Java-overridden operator (e.g., SelectSeq) receives a LAMBDA that references a primed variable through an INSTANCE substitution, TLC's ENABLED evaluation can silently return the wrong boolean. The root cause is that Context.branch in the OPCODE_enabled handler places a BRANCH node above the INSTANCE substitution bindings. With cutoff=true, getPrimedVar cannot traverse past the BRANCH, so it returns the inner module's SymbolNode instead of resolving through the substitution to the outer variable. TLCStateFun then binds values under inner variable names. If an inner variable name collides with the outer variable name that a different inner variable maps to, TLCStateFun.lookup (which compares by UniqueString) returns the wrong value to the LAMBDA.

Related to Github issue #725
https://github.com/tlaplus/tlaplus/issues/725

[Test][TLC]


tlatools/org.lamport.tlatools/src/tlc2/value/impl/OpLambdaValue.java
```diff
@@ -113,8 +123,12 @@ public final Value eval(Value[] args, int control) {
       for (int i = 0; i < alen; i++) {
         c1 = c1.cons(formals[i], args[i]);
       }
-      return (Value) this.tool.eval(this.opDef.getBody(), c1, this.state, this.pstate,
-          control);
+      // Restore the Enabled flag from the creation context. Java module overrides
+      // call this method with EvalControl.Clear, which drops the Enabled flag and
+      // causes primed-variable lookups to use cutoff=false — leading to incorrect
+      // INSTANCE substitution resolution. See Github issue #725.
+	  return (Value) this.tool.eval(this.opDef.getBody(), c1, this.state, this.pstate,
```
lemmy: @ Reviewers: `EvalControl.isEnabled(this.control) ? EvalControl.setEnabled(control) : control)` is the crucial part.

lemmy:
Merging this PR to get the change vetted by https://github.com/tlaplus/ValidationTestSuite.  Expect on 3/14 as per https://github.com/tlaplus/ValidationTestSuite/blob/deed81471363cb08ac21e949e8a870f49334ed41/.github/workflows/main.yml#L7

PR #1343, opened 2026-03-10
Enhance CI workflows to support Windows environment. 

lemmy:
Updated both main.yml and pr.yml to include 'windows-latest' in the operating system matrix. Adjusted build steps to conditionally run based on the operating system.

[Build][CI]

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/Simulator.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/Simulator.java
@@ -19,6 +19,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TimerTask;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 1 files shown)


ahelwer:
#968 fixes grammar sync check on windows (or more properly, makes git work invariantly to line endings)

lemmy:
https://github.com/tlaplus/tlaplus/actions/runs/22915671127/job/66502469387 shows many syntax corpus tests failing.

lemmy:
@ahelwer This is what an AI thinks causes some Unicode-related test failures on Windows:

Unicode ≜ not representable in Windows Cp1252 encoding (1 test)
Affected: GetScopedIdentifiersTests (all 18 test cases)

The test creates TLA+ spec strings containing Unicode characters like ≜ (U+225C), ∀ (U+2200), and ⟨ (U+27E8):

GetScopedIdentifiersTests.java
Lines 100-100
```java
      new TestCase("op ≜", ""),
```

On Windows, the default file encoding is Cp1252, which cannot represent these characters. The parser sees ? instead of ≜, causing Lexical error ... Encountered: "?" after : "?" in every test case. This is the same class of issue as the tla2sany parser tests (which you asked to skip).

ahelwer:
This is quite odd, as I see the files are being read using the `Files.readString()` method in `SyntaxCorpusFileParser.java`, which is a `java.nio` method so uses UTF-8 encoding by default. The `TLAplusParser` is also created with UTF-8 encoding specified by default. Will have to see whether my Windows partition still boots to debug this.

lemmy:
> This is quite odd, as I see the files are being read using the `Files.readString()` method in `SyntaxCorpusFileParser.java`, which is a `java.nio` method so uses UTF-8 encoding by default. The `TLAplusParser` is also created with UTF-8 encoding specified by default. Will have to see whether my Windows partition still boots to debug this.

Claude-4.6-opus-high fixed the test failures in https://github.com/tlaplus/tlaplus/pull/1343/changes/7177b28ef9f9a1b5019477f7ca8450128513fdc6.  Please review this change.

ahelwer:
That looks great! I was worrying about getting the stuff from files into strings in the right encoding but never considered that getting it out of strings in the right encoding was also important. Something to watch out for in the future.

PR #1344, opened 2026-03-11
Add -suppressMessages and -warningsAsErrors CLI flags to SANY

craft095:
Add -suppressMessages and -warningsAsErrors CLI flags to SANY (phase 1 of #1186).
    
Support of the fine-grained warning control in SANY:

-suppressMessages comma-separated integer codes; silences those messages entirely (not
  printed, not a failure).
-warningsAsErrors comma-separated integer warning codes; when any such warning is
  triggered SANY exits with the semantic-analysis failure code.

Both flags validate each supplied code against ErrorCode.fromStandardValue() and reject
unknown codes with a descriptive error. -warningsAsErrors additionally rejects
non-warning-level codes.

This also
- resolves the long-standing TODO comment in frontEndSemanticAnalysis() to split
  errors and warnings into appropriate log levels.
- fixes a bug where linting warnings were accumulated in spec.semanticErrors
  during frontEndLinting() but never printed to the SanyOutput, so they were
  invisible to callers using the API path.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
@@ -5,13 +5,17 @@ package tla2sany.drivers;
 
 import java.io.PrintStream;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import tla2sany.explorer.Explorer;
 import tla2sany.explorer.ExplorerQuitException;
 import tla2sany.linter.Linter;
 import tla2sany.modanalyzer.ParseUnit;
 import tla2sany.modanalyzer.SpecObj;
+import tla2sany.semantic.ErrorCode;
 import tla2sany.output.LogLevel;
 import tla2sany.output.OutErrSanyOutput;
 import tla2sany.output.SanyOutput;
@@ -90,6 +94,16 @@ public class SANY {
   private static final boolean doValidatePCalTranslationDefault = true;
 
   /**
+   * Set of message codes to silence by default
+   */
+  private static final Set<ErrorCode> defaultSuppressedCodes = Set.of();
+
+  /**
+   * Set of message codes to elevate to errors by default
+   */
+  private static final Set<ErrorCode> defaultMessagesAsErrorCodes = Set.of();
+
+  /**
    * Use {@link SANY#frontEndMain(SpecObj, String, SanyOutput, SanySettings)}
    * instead for greater control of output and settings. This method will
    * print all human-readable log messages to the provided syserr parameter.
@@ -117,7 +131,9 @@ public class SANY {
         doSemanticAnalysis,
         doLevelChecking,
         doLinting,
-        doValidatePCalTranslationDefault
+        doValidatePCalTranslationDefault,
+        defaultSuppressedCodes,
+        defaultMessagesAsErrorCodes
       );
     return parse(spec, fileName, out, settings).code();
   }
@@ -173,15 +189,24 @@ public class SANY {
       frontEndInitialize();
     
       // **** Parsing 
-      frontEndParse(spec, out, settings.validatePCalTranslation);
+      frontEndParse(spec, out, settings);
     
       // **** Semantic analysis and level checking
       if (settings.doSemanticAnalysis) 
-            {frontEndSemanticAnalysis(spec, out, settings.doLevelChecking);} ;
+            {frontEndSemanticAnalysis(spec, out, settings);} ;
 
       // **** Linting
-      if (settings.doLinting) 
-            {frontEndLinting(spec, out); } ;
+      if (settings.doLinting) {
+        final int warnsBefore = spec.semanticErrors.getWarningDetails().size();
+        frontEndLinting(spec, out);
+        // Print linting warnings, respecting suppressed/elevated codes.
+        final List<ErrorDetails> allWarnings = spec.semanticErrors.getWarningDetails();
+        final List<ErrorDetails> lintingWarnings = allWarnings.subList(warnsBefore, allWarnings.size());
+        final boolean errorElevated = reportMessages(out, filterVisibleMessages(lintingWarnings, settings), settings);
+        if (errorElevated) {
+          spec.errorLevel = SanyExitCode.SEMANTIC_ANALYSIS_OR_LEVEL_CHECKING_FAILURE.code();
+        }
+      }
 
     }
     catch (ParseException pe) {
@@ -200,6 +225,43 @@ public class SANY {
         : SanyExitCode.OK;
   }
 
+  /**
+   * Checks if a message should be treated as an error.
+   */
+  private static boolean isMessageElevated(ErrorDetails message, SanySettings settings) {
+    return settings.messagesAsErrorCodes.contains(message.getCode());
+  }
+
+  /**
+   * Filters the visible messages based on the settings.
+   */
+  private static List<ErrorDetails> filterVisibleMessages(List<ErrorDetails> messages, SanySettings settings) {
+    return messages.stream()
+        .filter(w -> !settings.suppressedCodes.contains(w.getCode()))
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Reports the messages to the output. Returns true if any elevated messages were encountered.
+   */
+  private static boolean reportMessages(SanyOutput out, List<ErrorDetails> messages, SanySettings settings) {
+    boolean errorElevated = false;
+    // Print non-elevated messages first.
+    for (ErrorDetails message : messages) {
+      if (!isMessageElevated(message, settings)) {
+        out.log(LogLevel.WARNING, "%s\n\n\n", message);
+      }
+    }
+    // Print elevated messages as errors.
+    for (ErrorDetails message : messages) {
+      if (isMessageElevated(message, settings)) {
+        out.log(LogLevel.ERROR, "Warning treated as error: %s\n\n\n", message);
+        errorElevated = true;
+      }
+    }
+    return errorElevated;
+  }
+
   /** 
    * Initialize the all parts of the FrontEnd to handle a new specification.
    */
@@ -208,7 +270,7 @@ public class SANY {
   } // frontEndInitialize
 
   /**
-   * Use {@link SANY#frontEndParse(SpecObj, SanyOutput)} instead for greater
+   * Use {@link SANY#frontEndParse(SpecObj, SanyOutput, SanySettings)} instead for greater
    * control of output.
    */
   @Deprecated
@@ -224,7 +286,7 @@ public class SANY {
   }
 
   /**
-   * Use {@link SANY#frontEndParse(SpecObj, SanyOutput, boolean)} instead for greater
+   * Use {@link SANY#frontEndParse(SpecObj, SanyOutput, SanySettings)} instead for greater
    * control of output.
    */
   @Deprecated
@@ -233,7 +295,29 @@ public class SANY {
     frontEndParse(spec, new SimpleSanyOutput(sysErr, LogLevel.INFO), validatePCalTranslation);
   }
 
-  public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validatePCalTranslation) 
+  /**
+   * Use {@link SANY#frontEndParse(SpecObj, SanyOutput, SanySettings)} instead.
+   */
+  @Deprecated
+  public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validatePCalTranslation)
+  throws ParseException {
+    SanySettings settings = new SanySettings(
+      doStrictErrorCodes,
+      doSemanticAnalysis,
+      doLevelChecking,
+      doLinting,
+      validatePCalTranslation,
+      defaultSuppressedCodes,
+      defaultMessagesAsErrorCodes
+    );
+
+    frontEndParse(spec, out, settings);
+  }
+
+  /**
+   * Parses the specification.
+   */
+  public static void frontEndParse(SpecObj spec, SanyOutput out, SanySettings settings)
   throws ParseException {
       /***********************************************************************
        * Modified on 12 May 2008 by LL to remove "throws AbortException",     *
@@ -243,13 +327,16 @@ public class SANY {
       try 
       {
           // Actual parsing method called from inside loadSpec()
-          spec.loadSpec(spec.getFileName(), spec.parseErrors, validatePCalTranslation, out);
+          spec.loadSpec(spec.getFileName(), spec.parseErrors, settings.validatePCalTranslation, out);
 
           List<ErrorDetails> warnings = spec.parseErrors.getWarningDetails();
-          if (!warnings.isEmpty()) {
-            out.log(LogLevel.WARNING, "Warnings (%d) during syntax parsing of %s:\n\n", warnings.size(), spec.getFileName());
-            for (ErrorDetails warning : warnings) {
-              out.log(LogLevel.WARNING, "%s\n\n\n", warning);
+          // Filter and display warnings, respecting suppressed/elevated codes.
+          final List<ErrorDetails> visibleWarnings = filterVisibleMessages(warnings, settings);
+          if (!visibleWarnings.isEmpty()) {
+            out.log(LogLevel.WARNING, "Warnings (%d) during syntax parsing of %s:\n\n", visibleWarnings.size(), spec.getFileName());
+            final boolean errorElevated = reportMessages(out, visibleWarnings, settings);
+            if (errorElevated) {
+              spec.errorLevel = SanyExitCode.SYNTAX_PARSING_FAILURE.code();
             }
           }
 
@@ -279,7 +366,7 @@ public class SANY {
   } //
 
   /**
-   * Use {@link SANY#frontEndSemanticAnalysis(SpecObj, SanyOutput, boolean)}
+   * Use {@link SANY#frontEndSemanticAnalysis(SpecObj, SanyOutput, SanySettings)}
    * for greater control over output.
    */
   @Deprecated
@@ -290,9 +377,32 @@ public class SANY {
     frontEndSemanticAnalysis(spec, new SimpleSanyOutput(syserr, LogLevel.INFO), levelCheck);
   }
 
+  /**
+   * Use {@link SANY#frontEndSemanticAnalysis(SpecObj, SanyOutput, SanySettings)} instead.
+   */
+  @Deprecated
   public static void frontEndSemanticAnalysis(SpecObj spec,
                                               SanyOutput out,
                                               boolean levelCheck) 
+  throws SemanticException {
+    SanySettings settings = new SanySettings(
+      doStrictErrorCodes,
+      doSemanticAnalysis,
+      levelCheck,
+      doLinting,
+      doValidatePCalTranslationDefault,
+      defaultSuppressedCodes,
+      defaultMessagesAsErrorCodes
+    );
+    frontEndSemanticAnalysis(spec, out, settings);
+  }
+
+  /**
+   * Performs semantic analysis on the specification.
+   */
+  public static void frontEndSemanticAnalysis(SpecObj spec,
+                                              SanyOutput out,
+                                              SanySettings settings)
   throws SemanticException {
     String      moduleStringName;
     TreeNode    syntaxTreeRoot;
@@ -314,7 +424,7 @@ public class SANY {
         // if semantic analysis has not already been done on this module
         if (externalModuleTable.getContext( UniqueString.uniqueStringOf( moduleStringName)) == null ) {
           parseUnit = (ParseUnit)spec.parseUnitContext.get(moduleStringName);;
-      
+
           // get reference to the syntax tree for the module
           syntaxTreeRoot = parseUnit.getParseTree();
 
@@ -351,7 +461,7 @@ public class SANY {
                                   moduleNode);
   
           // Level check if semantic analysis succeeded and levelCheck is true.
-          if (moduleNode != null && semanticErrors.isSuccess() && levelCheck ) {
+          if (moduleNode != null && semanticErrors.isSuccess() && settings.doLevelChecking ) {
             moduleNode.levelCheck(semanticErrors);
           }
 
@@ -362,8 +472,25 @@ public class SANY {
           }
 
           if (semanticErrors.getNumMessages() > 0) {
-            // TODO: split warnings & errors out into appropriate log level
-            out.log(LogLevel.ERROR, "Semantic errors:\n\n%s", semanticErrors);
+            // Print warnings, respecting suppressed/elevated codes
+            final List<ErrorDetails> allWarnings = semanticErrors.getWarningDetails();
+            final List<ErrorDetails> visibleWarnings = filterVisibleMessages(allWarnings, settings);
+            if (!visibleWarnings.isEmpty()) {
+              out.log(LogLevel.WARNING, "*** Warnings: %d\n", visibleWarnings.size());
+              final boolean errorElevated = reportMessages(out, visibleWarnings, settings);
+              if (errorElevated) {
+                spec.errorLevel = SanyExitCode.SEMANTIC_ANALYSIS_OR_LEVEL_CHECKING_FAILURE.code();
+              }
+            }
+
+            // Print errors at ERROR level.
+            final List<ErrorDetails> errors = semanticErrors.getErrorDetails();
+            if (!errors.isEmpty()) {
+              out.log(LogLevel.ERROR, "Semantic errors:\n\n*** Errors: %d\n", errors.size());
+              for (ErrorDetails error : errors) {
+                out.log(LogLevel.ERROR, "%s\n\n", error);
+              }
+            }
 
             // indicate fatal error during semantic analysis or level-checking
             if ( semanticErrors.getNumErrors() > 0 ) {
@@ -428,6 +555,17 @@ public class SANY {
           "'2' Error during parsing.\n" +
           "'4' Error during semantic analysis or level-checking.", true));
       variant.add(new UsageGenerator.Argument(
+          "-suppressMessages", "codes",
+          "Suppress specific messages; comma-separated list of message codes.\n" +
+          "Suppressed messages are not printed and do not cause SANY to fail.\n" + 
+          "Errors cannot be suppressed.\n " +
+          "Message codes can be found in https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/src/tla2sany/semantic/ErrorCode.java#L38", true));
+      variant.add(new UsageGenerator.Argument(
+          "-messagesAsErrors", "codes",
+          "Treat specific message codes as errors; comma-separated list of message codes.\n" +
+          "When triggered, SANY exits with failure.\n" +
+          "Message codes can be found in https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/src/tla2sany/semantic/ErrorCode.java#L38", true));
+      variant.add(new UsageGenerator.Argument(
           "-d", "Opens the semantic graph explorer prompt. The prompt accepts the following commands:\n" +
           "'cst' Prints out the concrete syntax tree.\n" +
           "'dot' Emits the semantic graph to a ModuleName.dot file in the DOT graph" +
@@ -470,6 +608,12 @@ public class SANY {
       printUsage();
       throw new SANYExitException(SanyExitCode.ERROR, "No arguments provided");
     }
+    // set of message codes to silence; populated by -suppressMessages CLI flag
+    final Set<ErrorCode> suppressedCodes = new HashSet<>();
+
+    // set of message codes to elevate to errors; populated by -messagesAsErrors CLI flag
+    final Set<ErrorCode> messagesAsErrorCodes = new HashSet<>();
+
     int i;
     // Parse and process the command line switches, which are
     // distinguished by the fact that they begin with a '-' character.
@@ -490,6 +634,52 @@ public class SANY {
       }
       else if (args[i].toLowerCase().equals("-error-codes"))
            doStrictErrorCodes = true;
+      else if (args[i].toLowerCase().equals("-suppressmessages")) {
+           i++;
+           if (i >= args.length) {
+             ToolIO.out.println("Error: -suppressMessages requires a comma-separated list of codes.");
+             throw new SANYExitException(SanyExitCode.ERROR, "-suppressMessages requires an argument");
+           }
+           for (String token : args[i].split(",")) {
+             token = token.trim();
+             try {
+               final int code = Integer.parseInt(token);
+               final ErrorCode messageCode = ErrorCode.fromStandardValue(code); // validate; throws if unknown
+               if (messageCode.getSeverityLevel() == ErrorCode.ErrorLevel.ERROR) {
+                 ToolIO.out.println("Error: -suppressMessages: code " + token + " is an error and cannot be suppressed.");
+                 throw new SANYExitException(SanyExitCode.ERROR, "-suppressMessages: unsuppressable code: " + token);
+               }
+               suppressedCodes.add(messageCode);
+             } catch (NumberFormatException nfe) {
+               ToolIO.out.println("Error: -suppressMessages: not an integer: " + token);
+               throw new SANYExitException(SanyExitCode.ERROR, "-suppressMessages: not an integer: " + token);
+             } catch (IllegalArgumentException iae) {
+               ToolIO.out.println("Error: -suppressMessages: unknown message code: " + token);
+               throw new SANYExitException(SanyExitCode.ERROR, "-suppressMessages: unknown code: " + token);
+             }
+           }
+      }
+      else if (args[i].toLowerCase().equals("-messagesaserrors")) {
+           i++;
+           if (i >= args.length) {
+             ToolIO.out.println("Error: -messagesAsErrors requires a comma-separated list of codes.");
+             throw new SANYExitException(SanyExitCode.ERROR, "-messagesAsErrors requires an argument");
+           }
+           for (String token : args[i].split(",")) {
+             token = token.trim();
+             try {
+               final int code = Integer.parseInt(token);
+               final ErrorCode errorCode = ErrorCode.fromStandardValue(code); // validate; throws if unknown
+               messagesAsErrorCodes.add(errorCode);
+             } catch (NumberFormatException nfe) {
+               ToolIO.out.println("Error: -messagesAsErrors: not an integer: " + token);
+               throw new SANYExitException(SanyExitCode.ERROR, "-messagesAsErrors: not an integer: " + token);
+             } catch (IllegalArgumentException iae) {
+               ToolIO.out.println("Error: -messagesAsErrors: unknown message code: " + token);
+               throw new SANYExitException(SanyExitCode.ERROR, "-messagesAsErrors: unknown code: " + token);
+             }
+           }
+      }
       else if (args[i].toLowerCase().equals("-help")) {
            printUsage();
            return;
@@ -504,6 +694,17 @@ public class SANY {
       ToolIO.out.println("At least 1 filename is required.");
       ToolIO.out.println("Try 'SANY -help' for more information.");
       throw new SANYExitException(SanyExitCode.ERROR, "No filename provided");
+    }
+
+    // Check that suppressMessages and messagesAsErrors do not intersect and report an error if they do.
+    Set<ErrorCode> intersection = new HashSet<>(suppressedCodes);
+    intersection.retainAll(messagesAsErrorCodes);
+    if (!intersection.isEmpty()) {
+        ToolIO.out.println("Error: The following codes were set to both -suppressMessages and -messagesAsErrors: " + intersection);
+        throw new SANYExitException(
+            SanyExitCode.ERROR,
+            "Options -suppressMessages and -messagesAsErrors overlap for codes: " + intersection
+        );
     }
 
     final SanyOutput out = new OutErrSanyOutput(
@@ -538,7 +739,9 @@ public class SANY {
                   SANY.doSemanticAnalysis,
                   SANY.doLevelChecking,
                   SANY.doLinting,
-                  SANY.doValidatePCalTranslationDefault
+                  SANY.doValidatePCalTranslationDefault,
+                  suppressedCodes,
+                  messagesAsErrorCodes
               );
               final SanyExitCode result = parse(spec, args[i], out, settings);
               if (result != SanyExitCode.OK) {
--- a/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SanySettings.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SanySettings.java
@@ -22,6 +22,10 @@
  ******************************************************************************/
 package tla2sany.drivers;
 
+import java.util.Set;
+
+import tla2sany.semantic.ErrorCode;
+
 /**
  * Encapsulates various settings controlling SANY execution. This is intended
  * to replace global static settings derived from parsing command line args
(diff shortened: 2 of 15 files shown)


tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -427,6 +522,15 @@ private static void printUsage()
           "-error-codes", "Return a descriptive exit code in case of error.\n\n" +
           "'2' Error during parsing.\n" +
           "'4' Error during semantic analysis or level-checking.", true));
+      variant.add(new UsageGenerator.Argument(
+          "-suppressMessages", "codes",
+          "suppress specific warning/message codes; comma-separated list of\n" +
+          "integer codes; suppressed messages are not printed and do not\n" +
```
lemmy: Nit: "integer" - I don't believe SANY or TLC have negative codes.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -490,6 +598,52 @@ else if (args[i].equals("-STAT") || args[i].equals("-stat")) {
...
+             } catch (NumberFormatException nfe) {
+               ToolIO.out.println("Error: -suppressMessages: not an integer: " + token);
+               throw new SANYExitException(SanyExitCode.ERROR, "-suppressMessages: not an integer: " + token);
+             } catch (IllegalArgumentException iae) {
+               ToolIO.out.println("Error: -suppressMessages: unknown message code: " + token);
+               throw new SANYExitException(SanyExitCode.ERROR, "-suppressMessages: unknown code: " + token);
+             }
+           }
+      }
+      else if (args[i].toLowerCase().equals("-warningsaserrors")) {
```
lemmy: Currently, there are two levels: warning and error. However, it is conceivable that additional levels (such as info) may be introduced in the future. Instead of specifically elevating warnings to errors, we could instead allow arbitrary messages to be elevated to errors. More generally, this flag would provide a mechanism to make SANY (or TLC) terminate when a specified message is encountered.

tlatools/org.lamport.tlatools/test/util/TestPrintStream.java
```diff
@@ -60,10 +60,9 @@ public void assertContains(final String seq) {
 	}
 	
 	public void assertSubstring(String substring) {
-		for (String string : strings) {
-			if (string.contains(substring)) {
-				return;
-			}
+		String combinedString = String.join("\n", strings);
```
lemmy: I don’t believe this PR should alter an assertion. Furthermore, this change is asymmetric because it doesn’t affect assertRegex, assertNoSubstring.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -195,11 +213,54 @@ public static final SanyExitCode parse(
...
+    // If any warnings in warningsAsErrorCodes were encountered, treat as failure.
+    if (!settings.warningsAsErrorCodes.isEmpty()) {
+      final boolean parseWarningElevated = spec.parseErrors.getWarningDetails().stream()
+          .anyMatch(w -> isWarningElevated(w, settings));
+      final boolean semanticWarningElevated = spec.semanticErrors.getWarningDetails().stream()
+          .anyMatch(w -> isWarningElevated(w, settings));
+      if (parseWarningElevated || semanticWarningElevated) {
+        return SanyExitCode.SEMANTIC_ANALYSIS_OR_LEVEL_CHECKING_FAILURE;
+      }
+    }
```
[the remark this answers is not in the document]
lemmy: Do not allow the codes of both flags to intersect?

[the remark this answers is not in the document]
craft095: Agree, we should not allow such combinations.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Wouldn't it be too confusing for users to use "-messagesAsErrors" instead of "-warningsAsErrors"? The latter is self-explanatory and well aligned with e.g. "-Werror" in gcc.
For niche use cases we can use "-warningsAsErrors" with any messages if we want.

tlatools/org.lamport.tlatools/test/util/TestPrintStream.java
craft095: I had to split a [previously one message](https://github.com/tlaplus/tlaplus/pull/1344/changes#diff-bd541374410de9bd709fc8d254d465009c079454d964fa6af08e4db7a0724455L365) into several messages, and some tests failed because of that.
I can support the same for other assertions, or rework failing tests. It would increase PR though.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
lemmy: The other flag also mentions “messages.” Thus, I find “messagesAsErrors” and “suppressMessages” more consistent and therefore less confusing.

tlatools/org.lamport.tlatools/test/util/TestPrintStream.java
lemmy: I prefer to retain the assertion as it is, since other existing tests rely on it.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -83,6 +87,12 @@ public class SANY {
     // true <=> error level should be reported as the tools'
     //          return value
 
+  private static Set<Integer> suppressedCodes = new HashSet<>();
```
ahelwer: Both of these can be made final

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -173,15 +185,21 @@ public static final SanyExitCode parse(
       frontEndInitialize();
     
       // **** Parsing 
-      frontEndParse(spec, out, settings.validatePCalTranslation);
+      frontEndParse(spec, out, settings.validatePCalTranslation, settings);
```
ahelwer: If you're passing in a settings object to a new frontEndParse method, consider removing the settings.validatePCalTranslation parameter since it is part of the settings anyway.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -173,15 +185,21 @@ public static final SanyExitCode parse(
...
     
       // **** Semantic analysis and level checking
       if (settings.doSemanticAnalysis) 
-            {frontEndSemanticAnalysis(spec, out, settings.doLevelChecking);} ;
+            {frontEndSemanticAnalysis(spec, out, settings.doLevelChecking, settings);} ;
```
ahelwer: Same as above

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -173,15 +185,21 @@ public static final SanyExitCode parse(
...
 
       // **** Linting
-      if (settings.doLinting) 
-            {frontEndLinting(spec, out); } ;
+      if (settings.doLinting) {
+        final int warnsBefore = spec.semanticErrors.getWarningDetails().size();
+        frontEndLinting(spec, out);
+        // Print linting warnings, respecting suppressed/elevated codes.
+        final List<ErrorDetails> allWarnings = spec.semanticErrors.getWarningDetails();
+        final List<ErrorDetails> lintingWarnings = allWarnings.subList(warnsBefore, allWarnings.size());
```
ahelwer, line 204: Would it be worth it to add a getLintingDetails() method to Errors so you don't have to do this?

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: How about storing ErrorCode enums in here instead of integers, validating the user-provided error codes in the CLI arg parsing method? It would also slightly simplify subsequent equality checks.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -234,6 +295,11 @@ public static void frontEndParse(SpecObj spec, PrintStream sysErr, boolean valid
   }
 
   public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validatePCalTranslation) 
+  throws ParseException {
+    frontEndParse(spec, out, validatePCalTranslation, null);
```
ahelwer: Not the biggest fan of adding null here. Maybe a better frontEndParse API design would destructure SanySettings to pass in the warning HashMaps directly. Then you could just pass in empty HashMaps here.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -293,6 +359,14 @@ public static void frontEndSemanticAnalysis(SpecObj spec,
   public static void frontEndSemanticAnalysis(SpecObj spec,
                                               SanyOutput out,
                                               boolean levelCheck) 
+  throws SemanticException {
+    frontEndSemanticAnalysis(spec, out, levelCheck, null);
```
ahelwer: Same as above

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -362,8 +436,29 @@ public static void frontEndSemanticAnalysis(SpecObj spec,
           }
 
           if (semanticErrors.getNumMessages() > 0) {
-            // TODO: split warnings & errors out into appropriate log level
-            out.log(LogLevel.ERROR, "Semantic errors:\n\n%s", semanticErrors);
+            // Print errors at ERROR level.
+            // Note: each out.log() call goes through println() which appends
+            // a newline, so we use one fewer trailing \n than Errors.toString()
+            // to produce identical output.
+            final List<ErrorDetails> errors = semanticErrors.getErrorDetails();
```
ahelwer: I think it makes most sense to print warnings before errors, so the most pertinent problems are showed to users right above the console line. It also makes sense to delay printing elevated warnings-as-errors until they can be grouped with errors.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -470,6 +574,10 @@ public static void SANYmain0(String args[]) throws SANYExitException {
       printUsage();
       throw new SANYExitException(SanyExitCode.ERROR, "No arguments provided");
     }
+    // Reset per-run state
+    suppressedCodes = new HashSet<>();
```
ahelwer: Nit: add an empty HashSet as a private static final field called like defaultSuppressedCodes

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -470,6 +574,10 @@ public static void SANYmain0(String args[]) throws SANYExitException {
...
+    warningsAsErrorCodes = new HashSet<>();
```
ahelwer: Same as above

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -117,7 +127,9 @@ public static final int frontEndMain(
         doSemanticAnalysis,
         doLevelChecking,
         doLinting,
-        doValidatePCalTranslationDefault
+        doValidatePCalTranslationDefault,
+        suppressedCodes,
```
ahelwer: These should be set to a default value (empty HashSet) instead of introducing another static variable, I think.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: This comment conflicts with a comment further down, where I separately advocate destructuring SanySettings into the message filter HashMaps; choose one or both.

tlatools/org.lamport.tlatools/test/util/TestPrintStream.java
craft095: OK, I will retain the existing assert behavior and fix failing tests instead.

tlatools/org.lamport.tlatools/test-model/sany/FieldNameClashWarning.tla
```diff
@@ -0,0 +1,9 @@
+---- MODULE FieldNameClashWarning ----
```
ahelwer: You can use the existing W4802_Test.tla for this.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -117,7 +127,9 @@ public static final int frontEndMain(
...
+        warningsAsErrorCodes
```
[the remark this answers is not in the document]
craft095: If the previous recommendation to copy these sets in `SanySettings` constructor is implemented, this one seems to be excessive.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: I guess currently we don't separate lints from warnings so that can't be done.

craft095: Well, the proposed trick with `subList` is far from ideal. However,

1. Adding `getLintingDetails` would require change in the linting code that deserves its own PR.
2. `getLintingDetails` would not be well aligned with `getErrorDetails` and `getWarningDetails`. It would logically require adding `getParseDetails` and `getSemanticDetails`.
3. Cleaner solution would be deferring error reporting until all phases are done. Then this problem won't even appear. I.e. `frontEndParse`, `frontEndSemanticAnalysis` and `frontEndLinting` should only collect the errors and reporting should be performed later at one place.

ahelwer: I do agree with 3 as a cleaner solution. Would be very tricky to do while maintaining backwards compatibility but is desirable. Anyway agree this is beyond the scope of this PR.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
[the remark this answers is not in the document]
ahelwer: Java also has [unmodifiable sets](https://docs.oracle.com/en/java/javase/11/core/creating-immutable-lists-sets-and-maps.html)

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Fixed

tlatools/org.lamport.tlatools/test/tla2sany/drivers/WarningControlTest.java
```diff
@@ -0,0 +1,244 @@
...
+   * code {@link SanyExitCode#SEMANTIC_ANALYSIS_OR_LEVEL_CHECKING_FAILURE} and
+   * prints "Warning treated as error".
+   */
+  @Test
+  public void testCLIWarningsAsErrorsCausesFailure() {
+    final TestPrintStream out = new TestPrintStream();
+    ToolIO.out = out;
+    try {
+      SANY.SANYmain0(new String[]{"-warningsAsErrors", "4802", "-error-codes", SPEC_PATH});
+      Assert.fail("Expected SANYExitException for elevated warning");
```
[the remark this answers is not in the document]
craft095: Fixed

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -195,11 +213,54 @@ public static final SanyExitCode parse(
       throw new FrontEndException(e);
     }
 
+    // If any warnings in warningsAsErrorCodes were encountered, treat as failure.
+    if (!settings.warningsAsErrorCodes.isEmpty()) {
+      final boolean parseWarningElevated = spec.parseErrors.getWarningDetails().stream()
+          .anyMatch(w -> isWarningElevated(w, settings));
+      final boolean semanticWarningElevated = spec.semanticErrors.getWarningDetails().stream()
+          .anyMatch(w -> isWarningElevated(w, settings));
+      if (parseWarningElevated || semanticWarningElevated) {
```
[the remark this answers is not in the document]
craft095: Reworked and removed this code.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SanySettings.java
```diff
@@ -120,11 +162,15 @@ public SanySettings(
...
+      final boolean validatePCalTranslation,
+      final Set<Integer> suppressedCodes,
+      final Set<Integer> warningsAsErrorCodes) {
     this.doStrictErrorCodes = doStrictErrorCodes;
     this.doSemanticAnalysis = doSemanticAnalysis;
     this.doLevelChecking = doLevelChecking;
     this.doLinting = doLinting;
     this.validatePCalTranslation = validatePCalTranslation;
+    this.suppressedCodes = suppressedCodes;
+    this.warningsAsErrorCodes = warningsAsErrorCodes;
```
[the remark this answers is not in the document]
craft095: Fixed.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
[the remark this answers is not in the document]
craft095: Fixed

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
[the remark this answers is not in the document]
craft095: Fixed.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -246,11 +312,11 @@ public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validateP
...
-            out.log(LogLevel.WARNING, "Warnings (%d) during syntax parsing of %s:\n\n", warnings.size(), spec.getFileName());
-            for (ErrorDetails warning : warnings) {
-              out.log(LogLevel.WARNING, "%s\n\n\n", warning);
-            }
+          // Filter and display warnings, respecting suppressed/elevated codes.
+          final List<ErrorDetails> visibleWarnings = filterVisibleMessages(warnings, settings);
+          if (!visibleWarnings.isEmpty()) {
+            out.log(LogLevel.WARNING, "Warnings (%d) during syntax parsing of %s:\n\n", visibleWarnings.size(), spec.getFileName());
+            reportWarnings(out, visibleWarnings, settings);
           }
```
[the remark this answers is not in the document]
craft095: Clarified.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Fixed.

tlatools/org.lamport.tlatools/test/util/TestPrintStream.java
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Partially addressed: `reportMessages` prints warnings first and only then elevated errors.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Used `clear()` method instead.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Used `clear()` method instead.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: I didn't understand the suggestion. This is deprecated static method - where should I use empty HashSet?

tlatools/org.lamport.tlatools/test-model/sany/FieldNameClashWarning.tla
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Fixed in a different way - there is null there anymore.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: I suggest:

1. Create a private final static immutable empty HashSet instance called like defaultSuppressedCodes in the SANY class
2. Use defaultSuppressedCodes as the parameter to SanySettings here instead of referencing a variable. Basically if people use this deprecated function then they can't get error code filtering.

We want to reduce the number of static global variables, not add to them.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -83,6 +87,12 @@ public class SANY {
     // true <=> error level should be reported as the tools'
     //          return value
 
+  private static final Set<ErrorCode> suppressedCodes = new HashSet<>();
+    // set of message codes to silence; populated by -suppressMessages CLI flag
+
+  private static final Set<ErrorCode> messagesAsErrorCodes = new HashSet<>();
```
ahelwer: I do think it is possible to not add static global variables here. We want to avoid adding more; removing them is a goal of the project see #891

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -200,6 +221,46 @@ public static final SanyExitCode parse(
         : SanyExitCode.OK;
   }
 
+  /**
+   * Checks if a message should be treated as an error.
+   */
+  private static boolean isMessageElevated(ErrorDetails message, SanySettings settings) {
+    return settings != null && settings.messagesAsErrorCodes.contains(message.getCode());
```
ahelwer: If you don't use settings as a null parameter anywhere then this null check (and others near here) is unnecessary.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -233,7 +294,22 @@ public static void frontEndParse(SpecObj spec, PrintStream sysErr, boolean valid
...
-  public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validatePCalTranslation) 
+  public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validatePCalTranslation)
+  throws ParseException {
+    SanySettings settings = new SanySettings(
+      doStrictErrorCodes,
+      doSemanticAnalysis,
+      doLevelChecking,
+      doLinting,
+      validatePCalTranslation,
+      suppressedCodes,
```
ahelwer: Same as above, this should reference a constant static field that is just the empty set instead of a static global variable. If people want error filtering they have to use the new API directly.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -362,11 +449,32 @@ public static void frontEndSemanticAnalysis(SpecObj spec,
...
+                    .filter(w -> !isMessageElevated(w, settings))
+                    .count();
+            if (normalCount > 0) {
+              out.log(LogLevel.WARNING, "*** Warnings: %d\n", normalCount);
+            }
+            final boolean errorElevated = reportMessages(out, visibleWarnings, settings);
 
-            // indicate fatal error during semantic analysis or level-checking
-            if ( semanticErrors.getNumErrors() > 0 ) {
+            // indicate ferror during semantic analysis or level-checking
```
ahelwer: ferror?

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -490,6 +611,51 @@ else if (args[i].equals("-STAT") || args[i].equals("-stat")) {
...
+             ToolIO.out.println("Error: -suppressMessages requires a comma-separated list of codes.");
+             throw new SANYExitException(SanyExitCode.ERROR, "-suppressMessages requires an argument");
+           }
+           for (String token : args[i].split(",")) {
+             token = token.trim();
+             try {
+               final int code = Integer.parseInt(token);
+               suppressedCodes.add(ErrorCode.fromStandardValue(code)); // validate; throws if unknown
+             } catch (NumberFormatException nfe) {
+               ToolIO.out.println("Error: -suppressMessages: not an integer: " + token);
```
ahelwer, line 654: I think both printing an error message & throwing it here will cause the message to be printed twice, which is undesirable.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -490,6 +611,51 @@ else if (args[i].equals("-STAT") || args[i].equals("-stat")) {
...
+           i++;
+           if (i >= args.length) {
+             ToolIO.out.println("Error: -suppressMessages requires a comma-separated list of codes.");
+             throw new SANYExitException(SanyExitCode.ERROR, "-suppressMessages requires an argument");
+           }
+           for (String token : args[i].split(",")) {
+             token = token.trim();
+             try {
+               final int code = Integer.parseInt(token);
+               suppressedCodes.add(ErrorCode.fromStandardValue(code)); // validate; throws if unknown
```
ahelwer: I believe we only want to suppress warnings and not errors, right? All errors are fatal so it does not make sense to suppress them. Thus you can check the level of the parsed code here to ensure it is only a warning.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Throwing SANYExitException doesn't result in printing - it only sets exit code.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Fixed.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Fixed.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Ah, makes sense. Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -89,6 +93,12 @@ public class SANY {
    */
   private static final boolean doValidatePCalTranslationDefault = true;
 
+  private static final Set<ErrorCode> defaultSuppressedCodes = Set.of();
+    // set of message codes to silence by default
```
ahelwer: Nit: though it matches the (incorrect) surroundings, the correct way to document things in JavaDoc is with a comment above the field or method. Ideally we want the documentation to show up when hovering over a reference to the field in major Java IDEs.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -233,7 +291,22 @@ public static void frontEndParse(SpecObj spec, PrintStream sysErr, boolean valid
     frontEndParse(spec, new SimpleSanyOutput(sysErr, LogLevel.INFO), validatePCalTranslation);
   }
 
-  public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validatePCalTranslation) 
+  public static void frontEndParse(SpecObj spec, SanyOutput out, boolean validatePCalTranslation)
```
ahelwer, line 302: Add a `@Deprecated` method annotation here with a comment pointing to the new `frontEndParse(SpecObj, SanyOutput, SanySettings)` method.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -293,6 +369,22 @@ public static void frontEndSemanticAnalysis(SpecObj spec,
   public static void frontEndSemanticAnalysis(SpecObj spec,
```
ahelwer, line 384: Same as above, add a `@Deprecated` method annotation here.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: I'm not totally sure I understand the behavior here, do we print errors, then warnings, then call reportMessages which prints warnings and then errors again?

tlatools/org.lamport.tlatools/test/tla2sany/drivers/WarningControlTest.java
```diff
@@ -0,0 +1,286 @@
...
+import util.SimpleFilenameToStream;
+import util.TestPrintStream;
+import util.ToolIO;
+
+/**
+ * Tests for the {@code -suppressMessages} and {@code -messagesAsErrors} CLI
+ * flags and corresponding {@link SanySettings} API fields, as part of GitHub
+ * issue #1186.
+ *
+ * <p>The fixture spec {@code FieldNameClashWarning.tla} triggers warning 4802
```
ahelwer: Nit: you no longer use this file so the comment should be changed

tlatools/org.lamport.tlatools/test-model/sany/FieldNameClashWarning.tla
ahelwer: Since you no longer use this file, should it be deleted?

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Reordered message printing so that the order is: warnings, elevated warnings, errors.

tlatools/org.lamport.tlatools/test-model/sany/FieldNameClashWarning.tla
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done, but it required elevating the new frontEndParse to public method, even though it is not used outside SANY at the moment.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
craft095: Done.

tlatools/org.lamport.tlatools/test/tla2sany/drivers/WarningControlTest.java
craft095: Done

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SanySettings.java
```diff
@@ -69,6 +73,20 @@ public final class SanySettings {
...
+   * The set of error codes whose messages should be silenced (not printed).
+   * An empty set means all messages are shown. Messages with codes in this
+   * set are neither printed nor cause SANY to fail.
+   */
+  public final Set<ErrorCode> suppressedCodes;
+
+  /**
+   * The set of message codes that should be elevated to errors. An empty set
+   * means no messages are elevated. When a message with a code in this set is
+   * encountered, SANY will treat it as an error and fail accordingly.
```
lemmy: Do we somewhere formally state that an *error* causes SANY to fail?

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -506,6 +699,17 @@ else if (args[i].toLowerCase().equals("-help")) {
       throw new SANYExitException(SanyExitCode.ERROR, "No filename provided");
     }
 
+    // Check that suppressMessages and messagesAsErrors do not intersect and report an error if they do.
+    Set<ErrorCode> intersection = new HashSet<>(suppressedCodes);
+    intersection.retainAll(messagesAsErrorCodes);
+    if (!intersection.isEmpty()) {
+        ToolIO.out.println("Error: The following codes were set to both -suppressMessages and -messagesAsErrors: " + intersection);
+        throw new SANYExitException(
+            SanyExitCode.ERROR,
```
lemmy, line 705: Can users suppress this error message? :-p

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -490,6 +633,56 @@ else if (args[i].equals("-STAT") || args[i].equals("-stat")) {
...
+           if (i >= args.length) {
+             ToolIO.out.println("Error: -messagesAsErrors requires a comma-separated list of codes.");
+             throw new SANYExitException(SanyExitCode.ERROR, "-messagesAsErrors requires an argument");
+           }
+           for (String token : args[i].split(",")) {
+             token = token.trim();
+             try {
+               final int code = Integer.parseInt(token);
+               final ErrorCode errorCode = ErrorCode.fromStandardValue(code); // validate; throws if unknown
+               if (errorCode.getSeverityLevel() != ErrorCode.ErrorLevel.WARNING) {
```
lemmy: I suggest removing this check. There’s no issue with requesting messages that are configured to be elevated to errors (it’s effectively a no-op). More importantly, we may want to introduce additional severity levels in the future, and there’s no clear reason to prevent those from being elevated to errors as well.

tlatools/org.lamport.tlatools/test/tla2sany/drivers/WarningControlTest.java
```diff
@@ -0,0 +1,286 @@
...
+import util.ToolIO;
+
+/**
+ * Tests for the {@code -suppressMessages} and {@code -messagesAsErrors} CLI
+ * flags and corresponding {@link SanySettings} API fields, as part of GitHub
+ * issue #1186.
+ *
+ * <p>The fixture spec {@code W4802_Pre_Test.tla} triggers warning 4802
+ * ({@link ErrorCode#RECORD_CONSTRUCTOR_FIELD_NAME_CLASH}) reliably: it defines
+ * {@code Foo == TRUE} and then constructs {@code [Foo |-> 42]}, where the field
```
lemmy, line 48: Is suppressing messages the primary use case here, or just happens to be the test case? If so, I’d argue it’s better to encourage users to update their specification instead. There are clear signs that some TLA+ users aren’t aware of this subtlety. Additionally, suppressing lints doesn’t seem granular enough, since the same lint message can appear multiple times.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -427,6 +555,15 @@ private static void printUsage()
           "-error-codes", "Return a descriptive exit code in case of error.\n\n" +
           "'2' Error during parsing.\n" +
           "'4' Error during semantic analysis or level-checking.", true));
+      variant.add(new UsageGenerator.Argument(
+          "-suppressMessages", "codes",
+          "Suppress specific warning/message codes; comma-separated list of\n" +
+          "integer codes. Suppressed messages are not printed and do not\n" +
```
lemmy: Are negative codes actually possible here? If not, we should remove “integer” since it’s unnecessary.

More importantly, this message doesn’t help users understand where to find valid codes. We should update it to point them to a source of truth. If we don’t have a built-in (dynamic or static) list, the next best option is to direct users to the GitHub repository where the codes are defined.

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
```diff
@@ -362,11 +472,29 @@ public static void frontEndSemanticAnalysis(SpecObj spec,
           }
 
           if (semanticErrors.getNumMessages() > 0) {
-            // TODO: split warnings & errors out into appropriate log level
-            out.log(LogLevel.ERROR, "Semantic errors:\n\n%s", semanticErrors);
+            // Print warnings, respecting suppressed/elevated codes
+            final List<ErrorDetails> allWarnings = semanticErrors.getWarningDetails();
+            final List<ErrorDetails> visibleWarnings = filterVisibleMessages(allWarnings, settings);
+            // Count non-elevated warnings to show correct header
+            final long normalCount = visibleWarnings.stream()
```
lemmy: Can we align this block and the one at https://github.com/tlaplus/tlaplus/pull/1344/changes#diff-bd541374410de9bd709fc8d254d465009c079454d964fa6af08e4db7a0724455R334-R338

tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
ahelwer: Probably we should start printing out the codes with the warning & error messages, but perhaps that can go in a separate PR.

tlatools/org.lamport.tlatools/test/tla2sany/drivers/WarningControlTest.java
craft095: Not sure I understand what's the concern. This is a test case for the suppression/elevation logic. It doesn't matter what warning we're dealing with given that it is triggered reliably.

> Additionally, suppressing lints doesn’t seem granular enough, since the same lint message can appear multiple times.

Yes, it is unable to suppress warnings depending on the context. This level of control would require in-file annotations like pragmas in most C-compilers.

PR #1347, opened 2026-03-13
Integrate Formatter into tla2tools

FedericoPonzi:
This PR ports the formatter from the [tlaplus-formatter](https://github.com/tlaplus/tlaplus-formatter/) repo to make it available as part of tlatools.jar; alongside SANY, TLC and other tools. As part of the porting, I've dropped the web related parts.

I created separate commits to ease the review:
1. Initial porting from tlaplus-formatter repo
2. Refactorings required to remove extra dependencies
3. Changes to ant build to include the formatter.
4. Update tests to work on Windows 
5. Update CI to fetch dependencies and run the tests.

Ref: https://github.com/tlaplus/tlaplus-formatter/issues/22

### Usage

```bash
java -cp tla2tools.jar formatter.Main [OPTIONS] <FILE> [OUTFILE]

Options:
  -i, --indent <N>             Indentation spaces (default: 4)
  -w, --width <N>              Target line width (default: 80)
  -v, --verbosity <LEVEL>      ERROR, WARN, INFO, DEBUG
      --skip-ast-verification  Skip AST verification after formatting
  -h, --help                   Show help message and exit
```

---

### Refactorings implemented to remove extra dependencies:
| Dependency | Reason | Replacement |
|-----------|--------|-------------|
| `tla2tools` | Now in same project | Direct classpath resolution |
| `commons-io` | Only used `WriterOutputStream` | `java.io.ByteArrayOutputStream` in `SANYWrapper.java` |
| `commons-cli` | CLI argument parsing | Manual argument parsing in `Main.java` |
| `slf4j` + `logback` | Logging framework | `java.util.logging` in `Main.java`, `TLAPlusFormatter.java`, `TlaDocBuilder.java` |
| JUnit 5 | Test framework | JUnit 4 (already present in tlatools) |

----
### Testing & CI
- **170 formatter unit tests** run on Ubuntu, Windows, and macOS
- **21 TLAPS-dependent tests** (ProofConstructTest, PickStepConstructTest, ExtendsConstructTest) require tlapm + CommunityModules
- **333 semantic preservation tests** run against [tlaplus/examples](https://github.com/tlaplus/examples) (Linux only, in the examples-tests job). These are running the formatter on all spec twice, and ensure AST is preserved.
- Formatter tests are **isolated**: `TLA-Library` is only set for the formatter's `<junit>` block to avoid module shadowing (e.g., `BagsTest` which would break if tlapm's `Bags.tla` shadows the standard module).
- TLA-Library paths are passed as separate Ant properties (`formatter.tlapm.path`, `formatter.community.modules.path`) rather than a single combined value, because Windows `ant.bat` splits on `;` in command-line arguments, silently dropping the second path.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -135,23 +135,11 @@ jobs:
       ## removed to allow the test suite to run to completion.
       ## (The same sequence of steps is executed by pr.yml)
       ## 
-    - name: Checkout TLAPM
-      uses: actions/checkout@v4
-      with:
-        repository: tlaplus/tlapm
-        path: tlapm
-    - name: Download CommunityModules
-      shell: bash
-      run: |
-        curl -sLO https://github.com/tlaplus/CommunityModules/releases/latest/download/CommunityModules-deps.jar
-        unzip -o CommunityModules-deps.jar -d community-modules-tla
     - name: Run TLC tests
       shell: bash
       run: |
         ant -f tlatools/org.lamport.tlatools/customBuild.xml info compile compile-test test \
-          -Dtest.halt=${{ inputs.failfast || 'false' }} \
-          -Dformatter.tlapm.path="${{ github.workspace }}/tlapm/library" \
-          -Dformatter.community.modules.path="${{ github.workspace }}/community-modules-tla"
+          -Dtest.halt=${{ inputs.failfast || 'false' }}
     - name: Generate JUnitReport
       if: always()
       run: ant -f tlatools/org.lamport.tlatools/customBuild.xml test-report
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -51,22 +51,10 @@ jobs:
     ## 
     - name: Build tools and unit tests
       run: ant -f tlatools/org.lamport.tlatools/customBuild.xml info compile compile-test dist
-    - name: Checkout TLAPM
-      uses: actions/checkout@v4
-      with:
-        repository: tlaplus/tlapm
-        path: tlapm
-    - name: Download CommunityModules
-      shell: bash
-      run: |
-        curl -sLO https://github.com/tlaplus/CommunityModules/releases/latest/download/CommunityModules-deps.jar
-        unzip -o CommunityModules-deps.jar -d community-modules-tla
     - name: Run unit tests
       shell: bash
       run: |
-        ant -f tlatools/org.lamport.tlatools/customBuild.xml info test -Dtest.halt=true \
-          -Dformatter.tlapm.path="${{ github.workspace }}/tlapm/library" \
-          -Dformatter.community.modules.path="${{ github.workspace }}/community-modules-tla"
+        ant -f tlatools/org.lamport.tlatools/customBuild.xml info test -Dtest.halt=true
     - name: Generate JUnitReport
       if: always()
       run: ant -f tlatools/org.lamport.tlatools/customBuild.xml test-report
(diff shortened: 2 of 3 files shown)


FedericoPonzi:
The PR is ready now, I've updated the PR description with relevant information as well.

.github/workflows/main.yml
```diff
@@ -135,9 +135,23 @@ jobs:
       ## removed to allow the test suite to run to completion.
       ## (The same sequence of steps is executed by pr.yml)
       ## 
+    - name: Checkout TLAPM
```
lemmy: Why do this and PR.yml need TLAPS/TLAPM?

.github/workflows/pr.yml
```diff
@@ -51,9 +51,22 @@ jobs:
...
+    - name: Checkout TLAPM
+      uses: actions/checkout@v4
+      with:
+        repository: tlaplus/tlapm
+        path: tlapm
+    - name: Download CommunityModules
+      shell: bash
+      run: |
+        curl -sLO https://github.com/tlaplus/CommunityModules/releases/latest/download/CommunityModules-deps.jar
+        unzip -o CommunityModules-deps.jar -d community-modules-tla
```
lemmy: Does this have to be renamed?  tla2tools.jar automatically loads `Community-Modules-deps.jar`.

.github/workflows/pr.yml
```diff
@@ -51,9 +51,22 @@ jobs:
...
     - name: Run unit tests
       shell: bash
-      run: ant -f tlatools/org.lamport.tlatools/customBuild.xml info test -Dtest.halt=true
+      run: |
+        ant -f tlatools/org.lamport.tlatools/customBuild.xml info test -Dtest.halt=true \
+          -Dformatter.tlapm.path="${{ github.workspace }}/tlapm/library" \
```
lemmy: I now understand why this clones tlapm. Considering that TLAPS has only a few specifications, I suggest avoiding introducing this dependency.

lemmy (approved):
I'm a bit concerned about introducing dependencies on TLAPM/TLAPS and CommunityModules. Even if the GitHub Action only forks those repositories, upstream changes could break the formatter tests and potentially cause the tla2tools.jar build to fail.

tlatools/org.lamport.tlatools/src/formatter/constructs/impl/GenInfixOp.java
```diff
@@ -0,0 +1,18 @@
...
+public class GenInfixOp extends AbstractAppendImageConstruct {
+    @Override
+    public int getSupportedNodeKind() {
+        return NodeKind.GEN_PREFIX_OP.getId();
+    }
+
+    @Override
+    public String getName() {
+        return "N_GenInfixOp";
+    }
```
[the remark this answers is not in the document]
FedericoPonzi: we can followup on these comments in a separate pr

FedericoPonzi:
Simplified the CI by removing the direct dependencies. The formatter has some tests that EXTENDS TLAPS (which in turn needs CommunityModules). This is needed because SANY needs to pass on the whole spec to generate the AST.

The Examples repo already pulls TLAPS: https://github.com/tlaplus/Examples/blob/master/.github/scripts/linux-setup.sh 
I've updated the CI to run the proof related tests in the Examples Integration Test, as long as the TLAPS modules are still available. It's a bit of an hack but it avoid the hard dependency on TLAPS and CommunityModules. Let me know what you think.

lemmy:
@FedericoPonzi Do you see value in keeping the six commits, or may I squash them into one big commit? I'm leaning towards the latter in this case.

FedericoPonzi:
Up to you, squashing is fine by me.

lemmy:
Do we archive the old repo with a note pointing to this repo?

FedericoPonzi:
Sounds good. Later I will update the vscode extension as well.

Issue #763, opened 2022-09-22
TLC reports error with trace, but no message about what went wrong


Calvin-L:
## Description

I encountered this while working on a real spec today. It was due to a typo on my part, but was very difficult to diagnose because TLC did not output an error message. Here's a relatively small spec that triggers the bug:

LivenessNoOutput.tla
```tla
---- MODULE LivenessNoOutput ----

CONSTANT MV
VARIABLE x
Init == x = {}
Grow == x' = [val |-> 1]
Next == Grow \/ UNCHANGED x
Spec == Init /\ [][Next]_x /\ WF_<<x>>(Grow)
Live == []<>(x /= MV)

====
```

LivenessNoOutput.cfg
```
SPECIFICATION Spec
CONSTANT MV = MV
PROPERTY Live
```

I haven't spent much time looking into it, but it seems to be an exception thrown due to the subscript of the `WF_<<x>>(...)` operator. Evaluating the `WF` operator requires comparing the value of the subscript in one state to the value in the previous state, and they are not comparable. Unexpectedly, changing the spec to `Spec == ... /\ WF_x(Grow)` causes TLC to reveal the error message, although this is only possible for artificial single-variable specs like this one.

## Expected Behavior
An error message at least as good as:
```
Error: TLC threw an unexpected exception.
This was probably caused by an error in the spec or model.
See the User Output or TLC Console for clues to what happened.
The exception was a java.lang.RuntimeException
: Attempted to check equality of the set {} with the value:
[val |-> 1]
```

## Actual Behavior
```
TLC2 Version 2.18 of Day Month 20?? (rev: e4961d3)
Running breadth-first search Model-Checking with fp 58 and seed 8837408557918770782 with 1 worker on 8 cores with 7282MB heap and 64MB offheap memory [pid: 38780] (Mac OS X 12.5.1 x86_64, Oracle Corporation 11.0.15.1 x86_64, MSBDiskFPSet, DiskStateQueue).
Parsing file /Users/cloncari/src/tlaplus/inputs/LivenessNoOutput.tla
Semantic processing of module LivenessNoOutput
Starting... (2022-09-22 14:26:31)
Implied-temporal checking--satisfiability problem has 1 branches.
Computing initial states...
Finished computing initial states: 1 distinct state generated at 2022-09-22 14:26:31.
Error: The behavior up to this point is:
State 1: <Initial predicate>
x = {}

3 states generated, 2 distinct states found, 1 states left on queue.
[...]

PR #1349, opened 2026-03-14
TLC reports error with trace, but no message about what went wrong

lemmy:
Commit 1004befc introduced support for re-running liveness checking with CallStackTool to improve error diagnostics (the irony). However, CallStackTool can throw a FingerprintException because it carries source information that FastTool does not. The calling code in Worker did not catch FingerprintException, which caused TLC to report an error trace without the corresponding error message. Catch FingerprintException and report the appropriate error message.

Fixes Github issue #763
https://github.com/tlaplus/tlaplus/issues/763

[Bug][TLC]


lemmy:
.oO(TLC's exception handling is a mess)

Issue #1313, opened 2026-02-16
Feedback & updates to the XML output format from TLAPM work


ahelwer:
One of the goals of https://github.com/tlaplus/tlapm/issues/213 is to serve as a comprehensive adaptation of the SANY XML exporter, so as to identify pitfalls & difficulties that others might encounter when using it in their own projects. Now that the SANY-XML translation work is largely complete, I will record my experience here so that the XML output format can be improved.

Overall, the XML output format is very good. With the exception of some corner cases it is generally obvious what TLA+ construct a given XML node type is representing, and the meaning of its fields. I will group the comparatively small number of difficulties I encountered into two types: parsing difficulties, and semantic mapping difficulties.

### Parsing Difficulties

This section deals with difficulties encountered when turning the XML representation back into a set of typed classes within the consuming application's language. While it is certainly possible for scripts or applications that only care about specific things to keep the XML as nested string-key-value dictionaries, I believe any serious consumer application that transforms the entire parse tree will adopt this intermediate representation approach. This was the approach taken by the OCaml code in TLAPM, which encountered the following difficulties:

- Optional fields: in general, it is easier to parse an XML node with a set known number of fields (children) rather than an XML node with optional fields that may or may not be present at certain indices in its children. However, there are many places where boolean flags are represented by the presence or absence of a field rather than a field that is always present but has a single child containing a boolean value. [Here](https://github.com/tlaplus/tlaplus/blob/631a9efab86b5f43695d337e59ec6ff268262c84/tlatools/org.lamport.tlatools/src/tla2sany/xml/sany.xsd#L399) is one example. The recommendation here is that **when possible, optional fields should be avoided and have their optional aspects delegated to presence or absence of that field's children**.
- Dual representation: many TLA+ constructs have variable-length fields, for example the list of parameters accepted by a user-defined operator. These variable-length fields can usually have 0 elements. However, some places in the XML schema list this variable-length field *itself* as optional. In `UserDefinedOpKind`, a 0-parameter operator can be represented both by an XML node with a missing "params" field, and by an XML node with a "params" field with zero children (see [example](https://github.com/tlaplus/tlaplus/blob/631a9efab86b5f43695d337e59ec6ff268262c84/tlatools/org.lamport.tlatools/src/tla2sany/xml/sany.xsd#L502-L508)). The recommendation here is that **variable-length fields should never be optional**.
[...]

Issue #1175, opened 2025-04-25
Support in XMLExporter to restrict export to set of modules


lemmy:
SANY's `XMLExporter` currently exports the full transitive closure of all modules. While this behavior is useful in the general case, it is excessive for scenarios like https://github.com/tlaplus/vscode-tlaplus/pull/367, where the goal is only to export a specific module (e.g., the root module) or all modules contained within a single .tla file.

Proposed Improvement:
The XMLExporter should support command-line options to control which modules are exported, allowing users to specify whether they want just the root module, all modules in a file, or the full transitive closure.

lemmy:
This is my half-assed attempt to suppress all but the root module, but it didn't fully work:

```diff
diff --git a/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java b/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
index 716a9bb68..195730a07 100644
--- a/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java
@@ -58,6 +58,8 @@ public class XMLExporter {
         "-o", "Offline mode; skip XML schema validation step.", true));
     args.add(new UsageGenerator.Argument(
         "-t", "Terse; format XML output without tabs or newlines.", true));
+    args.add(new UsageGenerator.Argument(
+            "-r", "Only include the root module in the output.", false));
     args.add(new UsageGenerator.Argument(
         "-I", "Include; use given directory path to resolve module dependencies.", true));
     args.add(new UsageGenerator.Argument(
@@ -102,10 +104,11 @@ public class XMLExporter {
   static void moduleToXML(String... args) throws XMLExportingException {

     if (args.length < 1) throw new IllegalArgumentException("at least one .tla file must be given");
-    LinkedList pathsLs = new LinkedList();
+    LinkedList<String> pathsLs = new LinkedList<>();

     boolean offline_mode = false;
     boolean pretty_print = true;
+    boolean rootModuleOnly = false;
     int lastarg = -1; // lastarg will be incremented, initialize at -1
     for (int i = 0; i < args.length - 1; i++) {
       if ("-o".equals(args[i])) {
@@ -114,6 +117,9 @@ public class XMLExporter {
       } else if ("-t".equals(args[i])) {
         pretty_print = false;
         lastarg = i;
[...]

PR #1352, opened 2026-03-17
XML Exporter: UserDefinedOpKind original op reference

ahelwer:
These changes modify the UserDefinedOpKind export format in several ways:
1. Add originalOperator and originallyDefinedInModule reference fields
2. Move optional pre-comments field below fixed fields to facilitate ease of parsing XML children
3. Fix dual-representation issue with parameters field where an operator with no parameters could be represented both by a missing parameters field and a parameters field with zero children; now a parameterless operator will have a parameters field with zero children

[Refactor][Feature][XMLExporter]

Ref #1313

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
@@ -1503,25 +1503,28 @@ public class OpDefNode extends OpDefOrDeclNode
         ret = doc.createElement("UserDefinedOpKind");
         ret.appendChild(appendText(doc,"uniquename",getName().toString()));
         ret.appendChild(appendText(doc,"arity",Integer.toString(getArity())));
-		if (getPreComments().length > 0) {
-			ret.appendChild(appendCDATA(doc, "pre-comments", Arrays.stream(getPreComments()).filter(Objects::nonNull)
-					// Do not trim to preserve indentation of comments.
-					.map(String::stripTrailing).filter(s -> !s.isEmpty())
-					// Use "\n" rather than System.lineSeparator(): XML 1.0 §2.11
-					// normalizes line endings to LF, so \r\n would not survive a
-					// round-trip through a conformant XML parser.
-					.collect(Collectors.joining("\n"))));
-		}
+        ret.appendChild(appendElement(doc, "originalOperator", this.getSource().export(doc, context, filter)));
+        ret.appendChild(appendElement(doc, "originallyDefinedInModule", this.getOriginallyDefinedInModuleNode().export(doc, context, filter)));
         ret.appendChild(appendElement(doc,"body",body.export(doc,context, filter)));
+        final Element arguments = doc.createElement("params");
+        ret.appendChild(arguments);
         if (params != null) {
-          Element arguments = doc.createElement("params");
           for (int i=0; i<params.length; i++) {
             Element lp = doc.createElement("leibnizparam");
             lp.appendChild(params[i].export(doc,context, filter));
             if (isLeibnizArg != null && isLeibnizArg[i]) lp.appendChild(doc.createElement("leibniz"));
             arguments.appendChild(lp);
           }
-          ret.appendChild(arguments);
+        }
+
+        if (getPreComments().length > 0) {
+          ret.appendChild(appendCDATA(doc, "pre-comments", Arrays.stream(getPreComments()).filter(Objects::nonNull)
+              // Do not trim to preserve indentation of comments.
+              .map(String::stripTrailing).filter(s -> !s.isEmpty())
+              // Use "\n" rather than System.lineSeparator(): XML 1.0 §2.11
+              // normalizes line endings to LF, so \r\n would not survive a
+              // round-trip through a conformant XML parser.
+              .collect(Collectors.joining("\n"))));
         }
 
         if (inRecursive) {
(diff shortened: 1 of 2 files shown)


tlatools/org.lamport.tlatools/src/tla2sany/xml/sany.xsd
```diff
@@ -1036,14 +1036,40 @@
...
+              will reference its own UID here.
             </xs:documentation>
           </xs:annotation>
+          <xs:complexType>
+            <xs:sequence>
+              <xs:element ref="UserDefinedOpKindRef"/>
+            </xs:sequence>
+          </xs:complexType>
+        </xs:element>
+        <xs:element name="originallyDefinedInModule">
```
ahelwer, line 1059: I think it would be a bit cleaner for this to be a reference to the actual module in which the operator is defined instead of the original, but couldn't find a way to get that information in OpDefNode.java. Then users could get both the actual module in which this operator is inlined, and also the original module by dereferencing the originalOperator field and looking at its module field.

tlatools/org.lamport.tlatools/src/tla2sany/xml/sany.xsd
```diff
@@ -1036,14 +1036,40 @@
...
           </xs:annotation>
         </xs:element>
-        <xs:element minOccurs="0" ref="pre-comments">
+        <xs:element name="originalOperator">
           <xs:annotation>
             <xs:documentation>
-              Any comments occurring before the operator. For example,
-              (* Text *) op == expr will have "(* Text *)" here, or "Text" if
-              certain XML Exporter settings are used.
+              When one module imports another using EXTENDS or INSTANCE, the
```
lemmy, line 1042: Is this compatible with XMLExporter's "Restrict mode; include only declarations and definitions of the specified module, excluding extended or instantiated modules."?

ahelwer: I'm not sure; I'm a bit confused about how the `-r` flag works. I tried running the XML Exporter from the master branch on these two modules with and without the `-r` flag:
```tla
---- MODULE Test ----
INSTANCE Other WITH C <- 0
====
```
```tla
---- MODULE Other ----
CONSTANT C
op == C
====
```
The only difference is that a `ModuleNodeRef` is missing from the trailing list of modules after the `context` field, but the `Other` module is still actually present in the `context` table along with all of its operators:
```
java tla2sany.xml.XMLExporter Test.tla > base.xml
java tla2sany.xml.XMLExporter -r Test.tla > restricted.xml
diff base.xml restricted.xml
242,244d241
<   <ModuleNodeRef>
<     <UID>304</UID>
<   </ModuleNodeRef>
```

ahelwer: @lemmy do you know what the restrict mode's behavior should be here?

lemmy: I would expect `Other` to be excluded, similar to how the `r` flag excludes `Other` when `INSTANCE` is converted into `EXTENDS`.  https://github.com/tlaplus/tlaplus/issues/1175 and https://github.com/tlaplus/tlaplus/pull/1235 are apparently incomplete.

ahelwer: I suppose so. Well, in that case these changes do not affect the restrict behavior.

ahelwer:
@Calvin-L would also appreciate a review on this one!

tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
```diff
@@ -1503,25 +1503,28 @@ protected Element getSymbolElement(Document doc, SymbolContext context, BiPredic
...
-					// Use "\n" rather than System.lineSeparator(): XML 1.0 §2.11
-					// normalizes line endings to LF, so \r\n would not survive a
-					// round-trip through a conformant XML parser.
-					.collect(Collectors.joining("\n"))));
-		}
+        ret.appendChild(appendElement(doc, "originalOperator", this.getSource().export(doc, context, filter)));
+        ret.appendChild(appendElement(doc, "originallyDefinedInModule", this.getOriginallyDefinedInModuleNode().export(doc, context, filter)));
         ret.appendChild(appendElement(doc,"body",body.export(doc,context, filter)));
+        final Element arguments = doc.createElement("params");
+        ret.appendChild(arguments);
```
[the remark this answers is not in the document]
ahelwer: Regression tests are done by running every tlaplus/examples spec through the XML Exporter in the CI and validating the output against this schema.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/OpDefNode.java
```diff
@@ -1503,25 +1503,28 @@ protected Element getSymbolElement(Document doc, SymbolContext context, BiPredic
...
-			ret.appendChild(appendCDATA(doc, "pre-comments", Arrays.stream(getPreComments()).filter(Objects::nonNull)
-					// Do not trim to preserve indentation of comments.
-					.map(String::stripTrailing).filter(s -> !s.isEmpty())
-					// Use "\n" rather than System.lineSeparator(): XML 1.0 §2.11
-					// normalizes line endings to LF, so \r\n would not survive a
-					// round-trip through a conformant XML parser.
-					.collect(Collectors.joining("\n"))));
-		}
+        ret.appendChild(appendElement(doc, "originalOperator", this.getSource().export(doc, context, filter)));
+        ret.appendChild(appendElement(doc, "originallyDefinedInModule", this.getOriginallyDefinedInModuleNode().export(doc, context, filter)));
```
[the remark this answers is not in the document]
ahelwer: These changes are being made in response to XML Exporter needs, so if these two fields are insufficient to identify imported definitions then further changes can be made in the future.

PR #1356, opened 2026-03-19
SANY: Only lint after semantic analysis and level checking

ahelwer:
Fixes crash if linter is run after skipping semantic analysis

Before these changes, SANY would crash if semantic analysis was skipped but linting was run anyway. Now SANY only runs the linter if semantic analysis completed successfully. In addition, SANY requires level-checking to have been run before running the linter, in anticipation of possible future linter messages that might require use of level information; it also just makes sense to have the most "optional" analysis level only run after all the strictly required analysis levels have run.

[Enhancement][Bug][SANY]

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/drivers/SANY.java
@@ -192,22 +192,22 @@ public class SANY {
       frontEndParse(spec, out, settings);
     
       // **** Semantic analysis and level checking
-      if (settings.doSemanticAnalysis) 
-            {frontEndSemanticAnalysis(spec, out, settings);} ;
+      if (settings.doSemanticAnalysis) {
+        frontEndSemanticAnalysis(spec, out, settings);
 
-      // **** Linting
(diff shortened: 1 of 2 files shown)


ahelwer:
Worked for a while on more elaborate changes that tracked a sort of "lifecycle state" in the `SpecObj` class then realized I was just trying to write OCaml code in Java and it was going nowhere, so I think keeping it simple and just moving linting inside the `settings.doSemanticAnalysis` block is fine.

lemmy:
This PR broke our VSCode extension's `Parse Module Stream Merging Tests` test: https://github.com/tlaplus/vscode-tlaplus/actions/runs/23358784806/job/67959300591

ahelwer:
Odd, will check it out.

ahelwer:
Ahh, this is because of this line:

https://github.com/tlaplus/vscode-tlaplus/blob/96ad1f0b73619c56909928f4e1c2d7437060b8b3/tests/suite/commands/parseModule.test.ts#L85

It hardcodes the "Linting of module ParseError" message and now we don't run the linter if parsing fails.

PR #1359, opened 2026-03-22
SANY: Improve stability by preserving operator definition order

### LINKED ISSUE #1044: Refactoring goal: improve stability of liveness checker

ahelwer:
TLC now also prints state traces with variables listed in the order of their definition

SANY's Context class is used to hold names that are in scope when resolving references. For this purpose it was using a Hashtable, which does not preserve the iteration order of inserted elements. While the Context class did implement its own linked list system to supply this order, this system was not often used and so users retrieved the list of defined operators in essentially random order directly from the Hashtable. This had ramifications further up the stack, ultimately affecting the state exploration order of the safety & liveness checkers and thus the specific counterexamples they found. While not a bug as such, this made it annoying to make changes to SANY as they could cause completely unrelated unit test failures.

These changes swap out the Hashtable with the newer LinkedHashMap data-structure from the Java standard library, which preserves iteration according to insertion order.

[Refactor][SANY]

Closes #1044 

Facilitates #1331

Likely necessary for the TLAPM/SANY parser rebasing work; still evaluating different approaches to get the XML Exporter to output unit-level definitions in the right order.


ahelwer:
Happily these changes turned out to be the key to a very nice user-facing feature which has always annoyed me: TLC will now print error traces with the variables matching the order of declaration in the module, instead of a random order! As this breaks a number of existing unit tests, I changed the unit test function that checks trace equivalence to check equality under permutation of conjuncts. Some unit tests still have to be fixed.

tlatools/org.lamport.tlatools/test/tlc2/tool/CommonTestCase.java
```diff
@@ -108,24 +120,110 @@ protected void assertTraceWith(final List<Object> actual, final List<String> exp
...
+	 *
+	 * Only re-ordering of top-level conjuncts is supported. Future work could
+	 * possibly expand to analyzing the abstract syntax tree itself, or even
+	 * rehydrating the state in TLC to compare it that way and avoid all tricky
+	 * string mismatch issues. While from this point forward TLC should print
+	 * out states with variables in a stable order, this permutation function
+	 * was written to avoid the very tedious task of converting all existing
+	 * unit test traces to use the new stable ordering.
+	 */
+	private static int[] assertStateEqualUnderPermutation(
```
lemmy, line 149: The complete fixture (trace) update is a mechanical, one-time task: run the tests, capture the new stable TLC output, and rewrite the expected strings in each test file. This is exactly the kind of tedious-but-straightforward bulk edit that AI tooling handles reliably, and it avoids permanently complicating the test infrastructure to save a one-time effort. Moreover, the permutation-invariant approach does not assert the operator definition order this PR introduces.  If something breaks the ordering guarantee, the tests won't catch it.

ahelwer: Is declaration-based trace variable ordering a sufficiently interesting feature to justify that large of a diff?

lemmy: As far as I recall, users can control the ordering via an `ALIAS`. That said, having a stable default ordering is likely beneficial for those running TLC outside of IDEs that enforce a specific order.

In any case, this change only affects test code, so it doesn’t meaningfully impact our goal of minimizing diffs to preserve effective git bisect usage.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/ModuleNode.java
```diff
@@ -386,8 +386,8 @@ public final OpDeclNode[] getConstantDecls() {
 
     final Vector<OpDeclNode> contextVec = ctxt.getConstantDecls();
     constantDecls = new OpDeclNode[contextVec.size()];
-    for (int i = 0, j = constantDecls.length - 1; i < constantDecls.length; i++) {
-      constantDecls[j--] = contextVec.elementAt(i);
+    for (int i = 0; i < constantDecls.length; i++) {
+      constantDecls[i] = contextVec.elementAt(i);
```
Calvin-L, line 390: It looks like this code (and the code on line 405 below) previously reversed the list. Now it doesn't. Is that significant?

If it isn't, then perhaps you can further simplify this to something like
```java
constantDecls = ctxt.getConstantDecls().toArray(new OpDeclNode[0]);
```

.github/workflows/pr.yml
```diff
@@ -54,12 +54,12 @@ jobs:
     - name: Run unit tests
       shell: bash
       run: |
-        ant -f tlatools/org.lamport.tlatools/customBuild.xml info test -Dtest.halt=true
+        ant -f tlatools/org.lamport.tlatools/customBuild.xml info test
```
Calvin-L, line 57: Was this change intentional?

ahelwer: This is just a temporary thing to try to catch all the test failures in one go, as otherwise I think it stops after a single failure.

tlatools/org.lamport.tlatools/src/tla2sany/semantic/ModuleNode.java
ahelwer: It is not significant because the list it was reversing had a random order, coming from enumerating the hashtable. The context class did maintain a sort of reverse linked list through the hash table which other methods in the ModuleNode class use, so reversing is the right thing to do in those cases. But then this method aped them by reversing the list without getting the elements from the reverse linked list. Arguably this fix could have just changed these methods to use the existing reverse linked list instead of switching to a LinkedHashMap, but LinkedHashMap is nice.

PR #1360, opened 2026-03-23
Add -suppressMessages and -warningsAsErrors CLI flags to TLC (phase 2…

craft095:
… of #1186)).

- Add isKnownCode(int) static method with a KnownCodes holder class that lazily builds the set of all known EC codes via reflection.
- Parse -suppressMessages and -warningsAsErrors in TLC.handleParameters(); validate codes and their combinations.
- SpecProcessor.java, CheckImplFile.java: Replace deprecated SANY.frontEndMain/ frontEndParse/frontEndSemanticAnalysis calls with SANY.parse() + SanySettings, forwarding any SANY-range suppressed/warningsAsError codes from MP.

Github issue #1186

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/TLC.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/TLC.java
@@ -14,6 +14,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -23,6 +24,7 @@ import java.util.Map;
 import java.util.Random;
 import java.util.Set;
 import java.util.TimeZone;
+import java.util.function.Consumer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -507,7 +509,21 @@ public class TLC {
                     printErrorMsg("Error: -suppressMessages requires a comma-separated list of message codes.");
                     return false;
                 }
-                if (!parseAndCategorizeMessageCodes(args[index++], sanySuppressedCodes, tlcSuppressedCodes, true)) {
+                try {
+                    Arrays.stream(args[index++].split(","))
+                          .map(String::trim)
+                          .mapToInt(Integer::parseInt)
+                          .forEach(code -> classifyMessageCode(code,
+                              tlcSuppressedCodes::add,
+                              sanyCode -> {
+                                  if (sanyCode.getSeverityLevel() == ErrorCode.ErrorLevel.ERROR) {
+                                      throw new IllegalArgumentException(
+                                          "code " + code + " is an error and cannot be suppressed.");
+                                  }
+                                  sanySuppressedCodes.add(sanyCode);
+                              }));
+                } catch (IllegalArgumentException e) {
+                    printErrorMsg("Error: " + e.getMessage());
                     return false;
                 }
             } else if (args[index].equals("-messagesAsErrors"))
@@ -517,7 +533,14 @@ public class TLC {
                     printErrorMsg("Error: -messagesAsErrors requires a comma-separated list of message codes.");
                     return false;
                 }
-                if (!parseAndCategorizeMessageCodes(args[index++], sanyMessagesAsErrorCodes, tlcMessagesAsErrorCodes, false)) {
+                try {
+                    Arrays.stream(args[index++].split(","))
+                          .map(String::trim)
+                          .mapToInt(Integer::parseInt)
+                          .forEach(code -> classifyMessageCode(code,
+                              tlcMessagesAsErrorCodes::add, sanyMessagesAsErrorCodes::add));
+                } catch (IllegalArgumentException e) {
+                    printErrorMsg("Error: " + e.getMessage());
                     return false;
                 }
             } else if (args[index].equals("-gzip"))
@@ -1605,54 +1628,20 @@ public class TLC {
     }
 
     /**
-     * Parses and cotegorizes a comma-separated list of message codes into {@code tlcTarget} and 
-	 * {@code sanyTarget}.
-     * Each code must be known to TLC ({@link EC#isKnownCode}) or to SANY
-     * ({@link tla2sany.semantic.ErrorCode#fromStandardValue}).
-     * If SANY code is an error, it is rejected. TLC has no per-code level metadata.
+     * Validates a single numeric code as either a TLC or SANY code and passes
+     * it to the appropriate consumer.
      *
-     * @param arg The comma-separated list of message codes.
-     * @param sanyTarget The set to add SANY message codes to.
-     * @param tlcTarget The set to add TLC message codes to.
-     * @param suppressed Whether the message codes are suppressed.
-     * @return {@code true} on success
-     *         {@code false} on the first invalid token.
+     * @throws IllegalArgumentException if the code is unknown.
      */
-    private boolean parseAndCategorizeMessageCodes(
-			final String arg, 
-			final Set<ErrorCode> sanyTarget,
-			final Set<Integer> tlcTarget,
-			final boolean suppressed) {
-        for (String token : arg.split(",")) {
-            token = token.trim();
-            final int code;
-            try {
-                code = Integer.parseInt(token);
-            } catch (final NumberFormatException e) {
-                printErrorMsg("Error: expected a message code, got: " + token);
-                return false;
-            }
-            // Validate: must be a known TLC code or a known SANY code.
-            if (EC.isKnownCode(code)) {
-                tlcTarget.add(code);
-                continue;
-            }
-            try {
-                final tla2sany.semantic.ErrorCode sanyCode = tla2sany.semantic.ErrorCode.fromStandardValue(code);
-                // Check if the code is an error and if it is, and it is suppressed, print an error 
-                // message and return false. We do it for SANY codes only because TLC has no 
-				// per-code level metadata to distinguish between warnings and errors.
-                if (suppressed && sanyCode.getSeverityLevel() == tla2sany.semantic.ErrorCode.ErrorLevel.ERROR) {
-                    printErrorMsg("Error: code " + code + " is an error and cannot be suppressed.");
-                    return false;
-                }
-                sanyTarget.add(sanyCode);
-            } catch (final IllegalArgumentException ignored) {
-                printErrorMsg("Error: unknown message code: " + code);
-                return false;
-            }
+    private static void classifyMessageCode(int code,
+            Consumer<Integer> onTlc, Consumer<ErrorCode> onSany) {
+        if (EC.isKnownCode(code)) {
+            onTlc.accept(code);
+        } else if (ErrorCode.isStandardValue(code)) {
+            onSany.accept(ErrorCode.fromStandardValue(code));
+        } else {
+            throw new IllegalArgumentException("unknown message code: " + code);
         }
-        return true;
     }
 
     /**
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/ErrorCode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/ErrorCode.java
@@ -222,6 +222,14 @@ public enum ErrorCode {
       .collect(Collectors.toMap(code -> code.value, code -> code));
 
   /**
+   * Returns {@code true} if {@code value} corresponds to a known
+   * {@link ErrorCode} enum constant.
+   */
+  public static boolean isStandardValue(final int value) {
+    return codeMap.containsKey(value);
+  }
+
+  /**
    * Given the standardized error value, find and return the enum instance
    * to which it corresponds.
    *


tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -217,8 +218,14 @@ public static class Colors {
     private static final DateTimeFormatter SDF = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss")
     		.withZone(ZoneId.systemDefault()); //$NON-NLS-1$
 	private static final ThreadLocal<DecimalFormat> DF = ThreadLocal.withInitial(() -> new DecimalFormat("###,###.###"));
-	
-	private static final java.util.Set<Integer> SUPPRESSED = new HashSet<>();
+
+    // TLC-range message codes
+	private static final java.util.Set<Integer> TLC_SUPPRESSED = new HashSet<>();
```
ahelwer: Maybe just import java.util.Set? It is used in a bunch of places in this file.

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java
```diff
@@ -409,7 +411,9 @@ private final void processSpec(final Mode mode)
             // Only if something unexpected happens the
             // exception is thrown
             SanyOutput out = new SimpleSanyOutput(ps, LogLevel.INFO);
-            SANY.frontEndMain(specObj, this.rootFile, out);
+            final SanySettings settings = SanySettings.forExternalCaller(
+                MP.getSanySuppressedCodes(), MP.getSanyMessagesAsErrorCodes());
+            SANY.parse(specObj, this.rootFile, out, settings);
```
ahelwer: The return value here should be checked to see whether it is `tla2sany.drivers.SanyExitCode.OK`, which should encompass all error handling except for catching the `FrontEndException`.

tlatools/org.lamport.tlatools/src/tlc2/tool/CheckImplFile.java
```diff
@@ -99,14 +101,13 @@ public final boolean getTrace()
...
         SpecObj spec = new SpecObj(rfname, null);
         try
         {
             SanyOutput out = new SimpleSanyOutput(ToolIO.out, LogLevel.INFO);
-            SANY.frontEndInitialize();
-            SANY.frontEndParse(spec, out);
-            SANY.frontEndSemanticAnalysis(spec, out, true);
+            final SanySettings settings = SanySettings.forExternalCaller(
+                MP.getSanySuppressedCodes(), MP.getSanyMessagesAsErrorCodes());
+            SANY.parse(spec, rfname, out, settings);
```
ahelwer: Same as above

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
```diff
@@ -1539,7 +1604,56 @@ private void printErrorMsg(String msg)
...
+                return false;
+            }
+            // Validate: must be a known TLC code or a known SANY code.
+            if (EC.isKnownCode(code)) {
+				tlcTarget.add(code);
+				continue;
+			}
+            try {
+                final tla2sany.semantic.ErrorCode sanyCode = tla2sany.semantic.ErrorCode.fromStandardValue(code);
+				// Check if the code is an error and if it is, and it is suppressed, print an error
```
ahelwer: wonky indentation compared to surroundings (likely mix of tabs and spaces)

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java
craft095: I thought about it but what should we do if it is not OK? We handle all the errors below anyway.

ahelwer: I think the `specObj.errorCode` check can be replaced with this one.

tlatools/org.lamport.tlatools/src/tlc2/output/EC.java
```diff
@@ -468,4 +472,31 @@ public static boolean exitStatusToCrash(final int exitStatus) {
...
+     * NOTE: Ideally, we should refactor TLC and introduce an enumeration of all the codes. However,
+     * it would result in multiple changes all over the codebase.
+     */
+    final class KnownCodes {
+        static final Set<Integer> ALL;
+        static {
+            final HashSet<Integer> s = new HashSet<>();
+            for (final Field f : EC.class.getDeclaredFields()) {
+                if (f.getType() == int.class && Modifier.isStatic(f.getModifiers())) {
+                    try { s.add(f.getInt(null)); } catch (final IllegalAccessException ignored) {}
```
lemmy: Please call `ignored.printStackTrace()` in the catch block instead of silently swallowing the exception. Even if we believe the exception should never occur, logging it will provide useful insight if it ever does.

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
```diff
@@ -1115,7 +1160,27 @@ else if (index < args.length)
                 }
             }
         }
-		
+
+        // Conflict check: -nowarning cannot be combined with per-code controls.
+		final boolean hasSuppressedCodes = !sanySuppressedCodes.isEmpty() || !tlcSuppressedCodes.isEmpty();
+		final boolean hasMessagesAsErrorCodes = !sanyMessagesAsErrorCodes.isEmpty() || !tlcMessagesAsErrorCodes.isEmpty();
+        if (!TLCGlobals.warn && (hasSuppressedCodes || hasMessagesAsErrorCodes)) {
```
lemmy: Why is `-nowarning` incompatible with elevating messages to errors?

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
```diff
@@ -1539,7 +1604,56 @@ private void printErrorMsg(String msg)
...
+            // Validate: must be a known TLC code or a known SANY code.
+            if (EC.isKnownCode(code)) {
+                tlcTarget.add(code);
+                continue;
+            }
+            try {
+                final tla2sany.semantic.ErrorCode sanyCode = tla2sany.semantic.ErrorCode.fromStandardValue(code);
+                // Check if the code is an error and if it is, and it is suppressed, print an error 
+                // message and return false. It's a best effort to prevent the user from suppressing 
+                // errors because we cannot do the same for TLC codes.
```
lemmy: I do not understand this comment., which suggests that SANY supports suppressing errors.  What does SANY do if it encounters a suppressed error?

Assuming SANY supports suppressing errors and does something reasonable, why would we prevent a user from using it here?

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
```diff
@@ -1539,7 +1604,56 @@ private void printErrorMsg(String msg)
...
+     * If SANY code is an error, it is rejected. TLC has no per-code level metadata.
+     *
+     * @param arg The comma-separated list of message codes.
+     * @param sanyTarget The set to add SANY message codes to.
+     * @param tlcTarget The set to add TLC message codes to.
+     * @param suppressed Whether the message codes are suppressed.
+     * @return {@code true} on success; prints an error message and returns
+     *         {@code false} on the first invalid token.
+     */
+    private boolean parseMessageCodes(final String arg,
```
lemmy: This name is misleading, as the method does more than simply parse message codes. In reality, it converts strings into integers as a preliminary step before categorizing them into TLC and SANY codes. It also conditionally suppresses its own error message, which I believe should be removed.

I recommend refactoring this logic into three distinct steps at the call site. There’s no need to be overly clever here—performance is not a concern.

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
```diff
@@ -391,6 +392,25 @@ private boolean checkEnvironment() {
...
 
+
+	/**
+	 * Generic method to check for overlap between two sets of codes
+	 * @param <T> the type of the codes
+	 * @param suppressedCodes the set of suppressed codes
+	 * @param messagesAsErrorCodes the set of messages as error codes
+	 * @return true if there is no overlap, false otherwise
+	 */
+	private <T> boolean checkForMessageCodeOverlap(final Set<T> suppressedCodes, final Set<T> messagesAsErrorCodes) {
```
lemmy, line 405: static?

tlatools/org.lamport.tlatools/src/tlc2/tool/CheckImplFile.java
```diff
@@ -116,6 +118,10 @@ public final boolean getTrace()
         {
             Assert.fail(EC.TLC_PARSING_FAILED);
         }
+        // Abort if SANY reported elevated errors.
```
lemmy, line 121: This looks like a tangential behavioral change that’s not directly tied to the main objective of this PR. Would you mind moving it into a separate commit to help keep the review focused?

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -238,22 +245,110 @@ public static class Colors {
...
+        SANY_MESSAGES_AS_ERRORS.clear();
     }
 
     /**
      * Private constructor to avoid instantiations
      */
     private MP()
     {
-        warningHistory = new Set();
+        warningHistory = new HashSet<>();
```
lemmy: Replacing `util.Set` with `java.util.HashSet` appears to introduce a behavioral change that is not directly related to the primary goal of this PR. In particular, many methods in `util.Set` are synchronized, whereas `HashSet` is not. Even seemingly minor refactorings like this can carry subtle risks, so it would be better to avoid them unless they are strictly necessary to the change at hand.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -238,22 +245,110 @@ public static class Colors {
...
+    public static void addSanyMessagesAsErrors(java.util.Collection<ErrorCode> codes) {
+        SANY_MESSAGES_AS_ERRORS.addAll(codes);
+    }
+
+    /**
+     * Clears all per-code message controls ({@link #TLC_SUPPRESSED}, {@link #TLC_MESSAGES_AS_ERRORS},
+     * {@link #SANY_SUPPRESSED}, and {@link #SANY_MESSAGES_AS_ERRORS}).  Intended for use by tests 
+     * to restore MP to a clean state between test cases.
+     */
+    public static void resetMessageControl() {
```
lemmy: Consider moving `WarningControlTest` to the `tlc2.output` package so this doesn't have to be public.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -238,22 +245,110 @@ public static class Colors {
         final String suppressed = System.getProperty("tlc2.output.MP", "");
```
lemmy: Should this be removed entirely, or should it instead produce an error guiding users to the new functionality?

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
craft095: Clarified the comment.

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
craft095: It uses member function to report an error and therefore, cannot be static.

tlatools/org.lamport.tlatools/src/tlc2/tool/CheckImplFile.java
craft095: I reverted the change, but wanted to clarify: do you mean a separate commit or a separate PR? I would say that this change is logically connected to this PR.

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
craft095: I renamed the function and clarified comment. I don't want to expand its content at the call site - it is already a huge function.

> It also conditionally suppresses its own error message, which I believe should be removed.

Not sure I understand. It doesn't suppress its own message. It reports an error when user tries to suppress SANY errors. It does so because it is impossible to suppress them and it feels like it is useful for user to be informed about it.

tlatools/org.lamport.tlatools/src/tlc2/tool/CheckImplFile.java
lemmy: I meant commit, not PR. Pull requests are a GitHub workflow artifact and become largely irrelevant once they’re closed. Commits, on the other hand, are permanent and remain important—especially when bisecting history to track down regressions. In general, each commit ideally represents a single, logical change.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -241,14 +241,6 @@ public static class Colors {
     static
     {
         instance = new MP();
-        
-        final String suppressed = System.getProperty("tlc2.output.MP", "");
```
lemmy, line 238: The removal of the system property is a behavioral change and should be explicitly mentioned in the commit message.

tlatools/org.lamport.tlatools/src/tlc2/TLC.java
lemmy: The new name still doesn’t fully capture the method’s responsibilities. If we were to make it explicit, it would end up as something like parseAndCategorizeAndCheckMessageCodesAndConditionallyReportErrors, which suggests the naming direction may not scale well.

As an alternative, I’d suggest: https://github.com/tlaplus/tlaplus/commit/477565b18a115d650ea8a51dd0ef582d5dafb8e8

craft095: > The new name still doesn’t fully capture the method’s responsibilities.  it would end up as something like parseAndCategorizeAndCheckMessageCodesAndConditionallyReportErrors

Is it unusual/unexpected that a parsing function does some checks and reports errors?

> As an alternative, I’d suggest: https://github.com/tlaplus/tlaplus/commit/477565b18a115d650ea8a51dd0ef582d5dafb8e8

1. https://github.com/tlaplus/tlaplus/commit/477565b18a115d650ea8a51dd0ef582d5dafb8e8 duplicates virtually the same code twice. Avoiding such a duplication was the main point of `parseAndCategorize`. 
2. It makes already large function even bigger.

I believe both points are valid reasons to factorize the code. 
That said, if you insist, I will adopt your suggestion.

lemmy: `handleParameters` is large because it is self-contained—it doesn’t delegate logic to helper methods in TLC.java. https://github.com/tlaplus/tlaplus/commit/477565b18a115d650ea8a51dd0ef582d5dafb8e8
duplicates string parsing and integer conversion, which appears in many else if branches within `handleParameters`. Please adopt https://github.com/tlaplus/tlaplus/commit/477565b18a115d650ea8a51dd0ef582d5dafb8e8

lemmy:
Logically merged via https://github.com/tlaplus/tlaplus/pull/1374

Issue #206, opened 2018-11-06
Allow efficiently combining VIEW with SYMMETRY reduction


ctmarinas:
For the attached spec and configuration file, with symmetry reduction, TLC generates 244 distinct states. Adding a debug "cmd" variable but limiting the VIEW to exclude it leads to TLC generating 470 distinct states. Ideally, the distinct states should remain at 244 when both SYMMETRY and VIEW are used.
[CPU.zip](https://github.com/tlaplus/tlaplus/files/2552823/CPU.zip)

lemmy:
https://github.com/tlaplus/tlaplus/blob/f17cbf1a20c65302857e61431bc49c317e693417/tlatools/src/tlc2/tool/TLCStateMut.java#L154-L255

PR #1369, opened 2026-04-03
fix: compute permutation on VIEW for fingerprint

marco6:
This PR attempts to fix #1368 and #206 by computing the view upfront and then applying symmetry sets to it.

Tests are passing, and trying it with both the simple TLA+ spec in #1368 and with the original #206 one seems to report the expected number of states.

I wanted to add both specs to the test cases, but I am not sure how to check the expected number of states generated. Probably adding a new test case in Java? Is it overkill? Please give me some guidance here.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/TLCStateMut.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/TLCStateMut.java
@@ -166,110 +166,112 @@ public final class TLCStateMut extends TLCState implements Serializable {
    *
    * Since the values in this state can be shared by multiple threads
    * via the state queue. They have to be normalized before adding to
-   * the state queue.  We do that here.
+   * the state queue. We do that here.
    */
-	public final long fingerPrint() {
-		int sz = this.values.length;
+  public final long fingerPrint() {
+    int sz = this.values.length;
+    for (int i = 0; i < sz; i++) {
+      this.values[i].deepNormalize();
+    }
 
-		// TLC supports symmetry reduction. Symmetry reduction works by defining classes
-		// of symmetrically equivalent states for which TLC only checks a
-		// single representative of the equivalence class (orbit). E.g. in a two
-		// process mutual exclusion problem, the process ids are - most of the time -
-		// not relevant with regards to mutual exclusion: We don't care if process A or
-		// B is in the critical section as long as only a single process is in CS. Thus
-		// two states are symmetric that only differ in the process id variable value.
-		// Symmetry can also be observed graphically in the state graph s.t. subgraphs
-		// are isomorphic (Graph Isomorphism). Instead of enumerating the complete state
-		// graph, TLC enumerates one of the isomorphic subgraphs whose state correspond
-		// to the representatives. With respect to the corresponding Kripke structure M,
-		// the resulting Kripke M' is called the "quotient structure" (see "Exploiting
-		// Symmetry in Temporal Logic Model Checking" by Clarke et al).
-		// 
-		// The definition of equivalence classes (orbits) is provided manually by the
-		// user at startup by defining 1 to n symmetry sets. Thus TLC has to find
-		// representative at runtime only which happens below. Given any state s, TLC
-		// evaluates rep(s) to find the lexicographically smallest state ss = rep(s)
-		// with regards to the variable values. The state ss is then fingerprint instead
-		// of s.
-		//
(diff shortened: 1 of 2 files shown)


lemmy:
@marco6 Thanks for your effort on this—much appreciated.

Please note that fingerprinting, symmetry, and views are particularly sensitive areas. Because of this, we need a very high level of confidence in the correctness of any changes before they can be accepted.

There are many existing Java/JUnit tests under tlatools/org.lamport.tlatools/test/ that verify expected counts of generated and distinct states (as well as counterexamples, etc.). You can reuse this testing infrastructure to add targeted tests for this specific area of TLC.

nit: Please try to minimize unrelated code reformatting so the diffs remain small and easier to review.

Calvin-L:
Yes, thanks for opening this PR. :)

My intuition says that this is a good change, provided we can get enough confidence that it's safe.

That said, I'll note that reversing the order of permutation/view symmetry might not be a universal win. Here's a sneaky little spec where the number of distinct states actually _increases_ with this change:

```tla
---- MODULE GitHub1369ViewSymmetry ----

EXTENDS TLC

CONSTANT S
VARIABLE x

Init == x = 0
Next == x' \in S
Spec == Init /\ [][Next]_x

Symmetry == Permutations(S)
View == <<x, CHOOSE a \in S: TRUE>>

====
```

```
SPECIFICATION Spec

CONSTANT S = {a,b}

SYMMETRY Symmetry
VIEW View
```

On current master (becec3564816c48dc07b9ab64319c8f7e606376d):

```
5 states generated, 2 distinct states found, 0 states left on queue.
```

On this branch (70dacb804f22de5e740999fd9d8cf9533112253d):

```
7 states generated, 3 distinct states found, 0 states left on queue.
```

I'm guessing that
 1. This change is safe and is a win when the view is symmetric wrt. the permutations
 2. For all practical use cases, the view is symmetric wrt. the permutations

In my sneaky spec above, the view is _not_ symmetric wrt. the permutations because it uses `CHOOSE` to pick a distinguished element of `S`.

...but it warrants further thought. My guesses might be wrong, and off the top of my head I'm not even sure how to rigorously define "the view is symmetric wrt. the permutations".

marco6:
> we need a very high level of confidence in the correctness of any changes before they can be accepted

Sure! Some tests are failing, I will try to spend some time digging into that. Also I will try to add the example and the original #206 one.

Another way could be to introduce this as an "experimental" feature for some time? Like either controlling that through a flag or an env variable? WDYT?

@Calvin-L while I cannot say I have any type of *proof* that computing the view before the symmetry set is the absolute best, isn't your example actually showing an improvement over correctness? 

Reading that spec, I see that a view can take 3 possible values under the symmetry set:

```
<< 0, a >> <-- init state with any of the values in the set
<< a, a >> <--- case where they are equal
<< a, b >> <--- case where they are *not* equal
```

Dumping the states from the example above gives me only two though:

```
<< 0, a >>
<< a, a >>
```

Am I missing something? From my PoV checking only two states is wrong in this case.

marco6:
@Calvin-L to show what I mean, I think it is enough to add an invariant to your spec:

```TLA
---- MODULE test ----

EXTENDS TLC

CONSTANT S
VARIABLE x

Init == x = 0
Next == x' \in S
Spec == Init /\ [][Next]_x

Symmetry == Permutations(S)
View == <<x, CHOOSE a \in S: TRUE>>
AreAlwaysEqual == x = 0 \/ x = CHOOSE a \in S: TRUE

====
SPECIFICATION Spec

CONSTANT S = {a,b}

SYMMETRY Symmetry
VIEW View

INVARIANT AreAlwaysEqual
```

The invariant can't really be true as `S` is composed by two distinct values and there must be a case where the invariant is violated as a single value can't equal two distinct values at the same time.

TLC reports *no error* here with the symmetry set. With a build from this branch, I get:

```
Computing initial states...
Finished computing initial states: 1 distinct state generated at 2026-04-07 09:30:17.
Error: Invariant AreAlwaysEqual is violated.
Error: The behavior up to this point is:
State 1: <Initial predicate>
x = 0

State 2: <Next line 9, col 9 to line 9, col 16 of module test>
x = b

3 states generated, 3 distinct states found, 1 states left on queue.
The depth of the complete state graph search is 2.
Finished in 00s at (2026-04-07 09:30:17)
```

Which seems more correct? Consider that I say "more" correct as I suspect that the example is not fit for using a symmetry set at all and that even with this patch, removing the view doesn't trigger the invariant failure.

PR #1371, opened 2026-04-04
Add optional CLI wrapper for tla2tools.jar

suckrowPierre:
This PR adds a minimal CLI wrapper (`tla`) for invoking the TLA+ tools.

Currently, users invoke tools via Java class names, for example:
  java -cp tla2tools.jar tlc2.TLC

This change introduces an optional unified entrypoint:

  tla tlc Spec.tla
  tla sany Spec.tla
  tla repl

The script delegates directly to existing Java entry points in
tla2tools.jar and does not introduce any new functionality.

Key properties:
- no changes to core tools or build system
- no assumptions about repository layout or build artifacts
- requires users to provide tla2tools.jar (e.g., via TLA_JAR)
- placed under general/cli to remain separate from core code

This is intended as a small usability improvement and does not affect
existing workflows.

As a secondary benefit, having a stable CLI entrypoint also makes it easier
to package the tla2tools via external package managers (e.g., Homebrew),
without requiring changes to the core repository. 
A brew formulae for the toolbox does already exists but not for tla2tools.

### DIFF
--- a/general/cli/README.md
+++ b/general/cli/README.md
@@ -1,0 +1,44 @@
+# TLA+ CLI Wrappers
+
+This directory contains a lightweight command-line wrapper for the TLA+ tools
+provided in `tla2tools.jar`.
+
+## Overview
+
+The TLA+ tools are typically invoked via Java class names, for example:
+
+```bash
+  java -cp tla2tools.jar tlc2.TLC
+```
+
+This wrapper provides a more convenient interface:
+
+```bash
+tla <command> [args]
+```
+
+Available commands:
+- `tlc` Run the model checker
+- `sany` Parse a TLA+ specification
+- `repl` Start the TLA+ REPL
+- `pcal` Translate PlusCal to TLA+
+- `tla2tex` Export TLA+ to LaTeX
+
+Examples:
+```bash
+tla tlc Spec.tla
+tla sany Spec.tla
+```
+
+## Requirements
+
+- Java 11+
+- `tla2tools.jar`, available from the official releases of the
+  TLA+ project
+
+## Setup
+
+Download `tla2tools.jar` from the official releases and set the environment variable:
+
+```sh
+export TLA_JAR=/path/to/tla2tools.jar
--- a/general/cli/tla
+++ b/general/cli/tla
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+JAVA="${JAVA_HOME:+$JAVA_HOME/bin/}java"
+JAR="${TLA_JAR:-}"
+
+if [ -z "$JAR" ]; then
+  echo "Error: tla2tools.jar not found." >&2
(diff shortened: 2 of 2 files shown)


ahelwer:
Thanks! This is one of those things we only have a chance to do once so it's important to get it right. I've wanted to do it for a while (#783) although never got around to it. But if you'd like to drive the process for this it probably involves opening up a discussion on the mailing list and also joining the monthly community meeting to talk it over with people. @hwayne has some experience with [tlacli](https://github.com/hwayne/tlacli) which might inform the design here as well. Basically a lot of design-level work!

suckrowPierre:
Thanks, that makes sense. I agree that something official and standardized, along the lines of tlacli, would deserve proper design discussion and broader planning. I will definitely take the steps you proposed.

Sorry that I did not follow the path described in `Contributing.md`. I thought that what I had in mind for this PR was relatively small in scope: not to define a new CLI, but just to make the existing tooling a bit easier to invoke. Basically, a thin wrapper around what already exists, with tla2tools remaining the source of truth.

That said, if supporting a Homebrew formula is also part of the goal (like I had in mind), then I understand the concern that changing the CLI later could become awkward.

ahelwer:
No worries, I think this is a good initiative. We've just been dragged down by backwards compatibility stuff for current command-line tooling creating a not-so-great experience (for example: requiring the `-error-codes` flag to SANY to actually return a nonzero exit code on failure) that I think it is worth it to spend a lot of time thinking about what we want these commands to look like. It would also be fine to prototype release them with the message that these commands can change at any time, since sharp edges are only found with use. But yeah, looping in other people is a good idea.

lemmy:
Since TLCI is referenced in this PR, I’d like to highlight a broader design consideration: we should aim to move as much functionality as possible into tla2tools.jar, keeping any wrapper script minimal and lightweight.

At the same time, TLC.java is difficult to modify due to legacy constraints. However, there’s no inherent limitation preventing us from introducing a new entry point—such as TLC2.java (or a better-named alternative)—that the wrapper script can delegate to. This new class could define a cleaner, more flexible command-line interface without being bound by the constraints of the existing implementation.

ahelwer:
@lemmy so you're thinking the wrapper script should do very little CLI arg parsing/translation and such, and instead just be a thin wrapper around the new entrypoint so people don't have to run `java -cp tla2tools.jar new.entrypoint.name`? I think that seems like a good approach, as otherwise the CLI wrapper script seems likely to grow in size and then suddenly we are maintaining a bunch of bash code (and possibly fish code or whatever other shells we want to support).

ahelwer:
@suckrowPierre there is a virtual TLA+ community meeting next week on Tuesday if you would like to join and discuss this work! https://zoom-lfx.platform.linuxfoundation.org/meetings/tla?view=month

suckrowPierre:
@ahelwer ah sorry, I just saw your message now. I will try to join the next one.

Issue #860, opened 2024-01-10
Feature proposal: `REACHABLE` invariant type


ahelwer:
Lamport often says "be suspicious of success" and to that end it is good practice to occasionally run your spec with some invariant that *should* fail, to ensure that a particular state is actually reached (liveness properties have too many issues with vacuous truth to give high certainty of same). This practice is useful & widespread enough in spec development that perhaps we should consider encoding it as an actual model file element: people can add the `SENTINEL` or `SENTINELS` keyword to their .cfg file along with a list of state-level invariant formulas. Each of these formulas should be false in at least one state reachable from the set of initial states. This will run at the safety checking level, so will not have the same vacuous truth issues as liveness checking (or require specifying fairness and such). If safety checking terminates without all the `SENTINEL` invariants being invalidated at some point, TLC will print an error listing the non-invalidated invariants then exit with a nonzero error code.

Edit: Per discussion below I'm a great fan of using `REACHABLE` instead of `SENTINEL`, even though as Markus says it will still be satisfied even if there are some initial states from which it is not reachable. It's reachable I guess from some synthetic null state serving as the starting point of TLC.

lemmy:
Related: https://github.com/tlaplus/tlaplus/issues/827

Calvin-L:
I like this idea! Today I regularly do something similar to what Markus suggested, or try to use coverage information to learn if a state got reached. I would prefer to have TLC just spit out an answer.

Quick thoughts:

(1) "Each of these formulas should be false in at least one state" --> that seems backwards from how I would use this feature. I often want to check that a formulas is _true_ in at least one state so I can make sure certain states are reachable. (E.g. termination: I want to make sure that `\A actor \in Actors: pc[actor] = "Done"` is reachable.)

(Long aside: point (1) is also related to my longstanding personal allergy to double negatives. `SENTINEL ~P` reads to me like "please verify that in some behavior, not P is false", which is needlessly confusing. If you flip it around we could have `IN_SOME_BEHAVIOR P` which means the same thing without the inverted inversion. Or if you want to check that some property really is false in some behavior, `IN_SOME_BEHAVIOR ~P` reads like "please verify that in some behavior, P is false". Pleasantly free of double negatives.)

(2) Why `SENTINEL`? To me that word means either "sentry" or "marker for end-of-data". Maybe
 - `NOT_INVARIANT` to check that a property is false in some reachable state (i.e. it is not an invariant of the system)
[...]

PR #1377, opened 2026-04-08
_POSSIBLE invariants and action properties

lemmy:
https://github.com/tlaplus/tlaplus/issues/860


tlatools/org.lamport.tlatools/src/tlc2/tool/impl/ModelConfig.java
```diff
@@ -259,21 +267,14 @@ public final void parse()
...
-                    String old = (String) this.configTbl.put(PostCondition, tt.image);
-                    if (old.length() != 0)
+                    while ((tt = getNextToken(tmgr)).kind != TLAplusParserConstants.EOF)
                     {
-                        throw new ConfigFileException(EC.CFG_TWICE_KEYWORD, new String[] { String.valueOf(loc),
-                                PostCondition });
+                        if (this.configTbl.get(tt.image) != null)
+                            break;
+                        postConditions.addElement(tt.image);
                     }
```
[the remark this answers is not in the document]
lemmy: Not great, but consistent with ACTION_CONSTRAINT/ACTION_CONSTRAINTS, PROPERTY/PROPERTIES, INVARIANT/INVARIANTS, ...

tlatools/org.lamport.tlatools/test-model/TLCSet.tla
```diff
@@ -39,6 +46,10 @@ Next == /\ x' = x + 1
 
 Spec == Init /\ [][Next]_x
 
+PostCondition ==
+    /\ TLCGet("s:label") = "hello"
```
craft095, line 50: How does TLCGet work in a postcondition? Does it implicitly refer to 0 worker's map?

tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_Possible.tla
```diff
@@ -0,0 +1,45 @@
...
+
+_Counts == \* Java override for efficiency
+    LET workers == TLCGet("all:named")["s:_possible"]
+        n == Len(workers)
+        Fold[i \in 0..n] ==
+            IF i = 0 THEN <<>>
+            ELSE _Merge(Fold[i-1], workers[i])
+    IN Fold[n]
+
+_Check ==
```
craft095: Is it guaranteed that every "possible" is represented in `_Counts` even if it was never TRUE? If it's not the case, `_Check` can  `TRUE`.
For example, will we actually see zero counter for an action "possible" if spec has no transitions at all - only initial states?

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCDebuggerExpression.java
```diff
@@ -283,7 +283,7 @@ public static OpDefNode process(final SpecProcessor processor, final ModuleNode
 		return bpOp;
 	}
 	
-	private static ExternalModuleTable resolveDependencies(final SpecProcessor processor, final ExternalModuleTable emt,
+	public static ExternalModuleTable resolveDependencies(final SpecProcessor processor, final ExternalModuleTable emt,
```
craft095, line 286: As the method became public, it would be useful to document it - what it does and what it expects.

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java
```diff
@@ -1681,7 +1693,231 @@ private void processActionConstraints() {
...
+	        }
+
+	        if (this.periodic == null) {
+	            this.periodic = printOp.getBody();
+	        }
+
+	        final ModuleNode possibleModule = emt.getModuleNode("_Possible");
+	        if (possibleModule != null) {
+	            final AssumeNode[] possibleAssumes = possibleModule.getAssumptions();
+	            if (possibleAssumes.length > 0) {
```
craft095: Why do we need to explicitly pull _Possible's assumptions in?

tlatools/org.lamport.tlatools/test-model/PossibleFailTest.tla
```diff
@@ -0,0 +1,14 @@
+---- MODULE PossibleFailTest ----
```
craft095: It might make sense to split the test into two: one fails state-level possible and another one fails action-level possible.
It would also be useful to check if error message is useful if there are several possibles but only some of them fail.

tlatools/org.lamport.tlatools/src/tlc2/module/TLCGetSet.java
```diff
@@ -82,6 +82,12 @@ public class TLCGetSet implements ValueConstants {
 
 	public static final UniqueString BEHAVIOR = UniqueString.of("behavior");
 	public static final UniqueString ALL_VALUES = UniqueString.of("all");
+	public static final UniqueString ALL_NAMED_VALUES = UniqueString.of("all:named");
+
+	// String keys starting with this prefix are mapped to per-worker named
+	// registers (Map<UniqueString, IValue> in each IdThread), avoiding the
+	// array-sizing overhead of integer-indexed registers.
+	public static final String NAMED_REGISTER_PREFIX = "s:";
```
craft095, line 90: PR proposes to use "s:" prefix for named registers, and use "s:_" prefix for internal named registers. Would it be more straightforward to introduce (as a convention) namespaces to use e.g. "user:" for user's registers and "tlc:" for internal use?

tlatools/org.lamport.tlatools/test-model/TLCSet.tla
lemmy: This is not explicitly defined, but by default it refers to the 0-worker’s map unless another worker has reported a counterexample. In that case, it instead refers to the map of the worker that reported the counterexample.

This is also why `TLCGet("all")` and, now, `TLCGet("all:named")` exist—they allow users to aggregate the register across all workers.

tlatools/org.lamport.tlatools/src/tlc2/debug/TLCDebuggerExpression.java
lemmy: I agree, but I’ll postpone this for a later refactoring. The entire class needs to be refactored and moved out of the debug package.

tlatools/org.lamport.tlatools/src/tlc2/module/TLCGetSet.java
lemmy: `s:` denotes the “type,” and the underscore (`_`) signifies TLC’s internal namespace. Using `user:` would make it appear as if the existing, numbered registers are non-user registers.  Moreover, we've been using the `_` prefix elsewhere to denote our internal namespace.

tlatools/org.lamport.tlatools/src/tlc2/tool/impl/SpecProcessor.java
lemmy: `Finally, the {@code ASSUME TLCSet("s:_possible", <<>>)} from {@code _Possible.tla} is extracted and appended to {@code this.assumptions} so the per-worker named register is initialized before state exploration.`

...but I can think of a better way.

tlatools/org.lamport.tlatools/src/tla2sany/StandardModules/_Possible.tla
lemmy: Good catch, `_Check` has been removed.

tlatools/org.lamport.tlatools/test-model/TLCSet.tla
lemmy: WRT this PR, the Java module override `_Possible.java` always aggregates across all workers.

hwayne:
Oh my gosh

lemmy:
# TLC `_POSSIBLE`

`_POSSIBLE` is a TLC `.cfg` directive for reachability checks. It requires that each listed predicate be witnessed at least once during model checking.

> Status: experimental/provisional. The name and semantics may change.

## Syntax

```cfg
_POSSIBLE OpName
```

or:

```cfg
_POSSIBLE
  OpName1
  OpName2
```

Each `OpName` must name a nullary operator in the root module.

## State predicate

```tla
ReachedDone == pc = "Done"
```

```cfg
SPECIFICATION Spec

_POSSIBLE ReachedDone
```

TLC succeeds only if some reachable state satisfies `ReachedDone`.

## Action predicate

```tla
SendMsg == msg' # msg
```

```cfg
SPECIFICATION Spec

_POSSIBLE SendMsg
```

If the predicate contains primed variables, TLC treats it as an action predicate. TLC succeeds only if some checked transition satisfies it.

## Failure condition

At the end of model checking, TLC reports an error for each `_POSSIBLE` predicate with no witness.

Example:

```tla
ReachedBad == x = 42
```

```cfg
_POSSIBLE ReachedBad
```

If no reachable state has `x = 42`, TLC fails the run.

## Relation to existing TLC checks

| Directive             | Meaning                                                 |
| --------------------- | ------------------------------------------------------- |
| `INVARIANT P`         | `P` holds in every reachable state                      |
| `PROPERTY P`          | temporal formula checked over behaviors                 |
| `CONSTRAINT P`        | restricts reachable states                              |
| `ACTION_CONSTRAINT A` | restricts transitions                                   |
| `_POSSIBLE P`         | `P` holds in some reachable state or checked transition |

## Interaction with constraints

`_POSSIBLE` is evaluated over the state graph TLC actually constructs.

If a `CONSTRAINT` removes a state, a state predicate cannot be witnessed there.

If an `ACTION_CONSTRAINT` removes a transition, an action predicate cannot be witnessed there.

```cfg
CONSTRAINT x < 3
_POSSIBLE ReachedThree
```

```tla
ReachedThree == x = 3
```

`ReachedThree` is not witnessed because states with `x = 3` are outside the constrained state graph.

## Typical use

Use `_POSSIBLE` to assert that a model actually reaches a case of interest:

```tla
ReachedElection == pc = "Election"
ReachedCommit == pc = "Commit"
CrashRecovery == crashed /\ recovered
```

```cfg
SPECIFICATION Spec
INVARIANT TypeOK

_POSSIBLE
  ReachedElection
  ReachedCommit
  CrashRecovery
```

This is useful for detecting vacuous specs, overrestrictive constraints, disabled branches of `Next`, and unreachable control locations.

## Notes

`_POSSIBLE P` is existential over the TLC run. It does **not** mean:

```tla
[]<>P
```

or:

```tla
<>P
```

as a temporal property over every behavior.

It means TLC found at least one reachable state, or one checked transition, satisfying `P`.

PR #1378, opened 2026-04-18
fix: show config file errors directly without 'unexpected exception' wrapper

### LINKED ISSUE #1234: Make TLC's missing config file error message clearer and more concise
When TLC is run without a configuration file, it prints the following error:

```
The exception was a tlc2.tool.ConfigFileException
: TLC encountered the following error when trying to read the configuration file DieHard.cfg:
File not found.
```

This message is overly verbose and misleading:
* It suggests there was an "unexpected exception," when in fact the issue is simply that the config file is missing.
* The text implies a problem with the spec or model, which isn’t accurate in this case.

mvanhorn:
When TLC cannot find a configuration file, the error was wrapped in a misleading message:

```
The exception was a tlc2.tool.ConfigFileException
: TLC encountered the following error when trying to read the configuration file DieHard.cfg:
File not found.
```

A missing config file is a routine user error, not an unexpected exception. The wrapper adds noise and misdirects users into thinking something broke internally.

Added an early return in `ECGeneralMsg` (`MP.java`) for `ConfigFileException` so the user sees just the config file error directly, without the alarming wrapper text. Other exceptions still show the full diagnostic wrapper.

Closes #1234

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
@@ -13,7 +13,6 @@ import java.util.HashSet;
 
 import tla2sany.semantic.ErrorCode;
 import tlc2.TLCGlobals;
-import tlc2.tool.ConfigFileException;
 import tlc2.tool.TLCState;
 import tlc2.tool.TLCStateInfo;
 import tlc2.tool.liveness.LiveWorker;
@@ -1714,10 +1713,6 @@ public class MP
      * @return
      */
     public static String ECGeneralMsg(String cause, Throwable throwable) {
-        if (throwable instanceof ConfigFileException
-                && ((ConfigFileException) throwable).getErrorCode() == EC.CFG_ERROR_READING_FILE) {
-            return throwable.getMessage();
-        }
         String msg = "TLC threw an unexpected exception.";
         msg = msg
                 + "\nThis was probably caused by an error in the spec or model.";
--- a/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
@@ -13,6 +13,7 @@ import java.util.HashSet;
 
 import tla2sany.semantic.ErrorCode;
 import tlc2.TLCGlobals;
+import tlc2.tool.ConfigFileException;
 import tlc2.tool.TLCState;
 import tlc2.tool.TLCStateInfo;
 import tlc2.tool.liveness.LiveWorker;
@@ -1713,6 +1714,9 @@ public class MP
      * @return
      */
     public static String ECGeneralMsg(String cause, Throwable throwable) {
+        if (throwable instanceof ConfigFileException) {
+            return throwable.getMessage();
+        }
         String msg = "TLC threw an unexpected exception.";
         msg = msg
                 + "\nThis was probably caused by an error in the spec or model.";
--- a/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
@@ -1714,7 +1714,8 @@ public class MP
      * @return
      */
     public static String ECGeneralMsg(String cause, Throwable throwable) {
-        if (throwable instanceof ConfigFileException) {
+        if (throwable instanceof ConfigFileException
+                && ((ConfigFileException) throwable).getErrorCode() == EC.CFG_ERROR_READING_FILE) {
             return throwable.getMessage();
         }
         String msg = "TLC threw an unexpected exception.";
--- a/tlatools/org.lamport.tlatools/src/tlc2/tool/ConfigFileException.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/tool/ConfigFileException.java
@@ -11,13 +11,22 @@ import tlc2.output.MP;
  */
 public class ConfigFileException extends RuntimeException
 {
+    private final int errorCode;
+
     public ConfigFileException(int errorCode, String[] parameters)
     {
         super(MP.getMessage(errorCode, parameters));
+        this.errorCode = errorCode;
     }
 
     public ConfigFileException(int errorCode, String[] parameters, Exception cause)
     {
         super(MP.getMessage(errorCode, parameters), cause);
+        this.errorCode = errorCode;
+    }
+
+    public int getErrorCode()
+    {
+        return errorCode;
     }
 }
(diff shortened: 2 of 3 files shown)


tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -1713,6 +1714,9 @@ public static void printError(int errorCode, String cause, Throwable throwable)
      * @return
      */
     public static String ECGeneralMsg(String cause, Throwable throwable) {
+        if (throwable instanceof ConfigFileException) {
```
lemmy, line 1717: `CFE` is thrown in `ModelConfig#parse` for multiple reasons. The issue discussed in https://github.com/tlaplus/tlaplus/issues/1234 specifically concerns `CFG_ERROR_READING_FILE`. For other error codes, however, the message “This was probably caused by an error in the spec or model.” seems appropriate.

mvanhorn:
Good call, thanks. Pushed 1b5c2e0: exposed `ConfigFileException.getErrorCode()` and narrowed the early return in `ECGeneralMsg` to `EC.CFG_ERROR_READING_FILE`. All the other `CFG_*` codes thrown from `ModelConfig#parse` (missing ID, twice keyword, expected symbol, etc.) still get the 'probably an error in the spec or model' wrapper.

tlatools/org.lamport.tlatools/src/tlc2/tool/ConfigFileException.java
```diff
@@ -11,13 +11,22 @@
  */
 public class ConfigFileException extends RuntimeException
 {
+    private final int errorCode;
```
ahelwer, line 14: Add a javadoc link to where the error code domain is defined

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
```diff
@@ -1713,6 +1714,10 @@ public static void printError(int errorCode, String cause, Throwable throwable)
      * @return
      */
     public static String ECGeneralMsg(String cause, Throwable throwable) {
+        if (throwable instanceof ConfigFileException
```
ahelwer, line 1717: If the answer to #1379 is "yes" then this could be simplified to `if (throwable instanceof ConfigFileException cfe && cfe.getErrorCode() == EC.CFG_ERROR_READING_FILE)` but that might not be the case yet.

tlatools/org.lamport.tlatools/src/tlc2/tool/ConfigFileException.java
```diff
@@ -11,13 +11,22 @@
...
+        this.errorCode = errorCode;
     }
 
     public ConfigFileException(int errorCode, String[] parameters, Exception cause)
     {
         super(MP.getMessage(errorCode, parameters), cause);
+        this.errorCode = errorCode;
+    }
+
+    public int getErrorCode()
```
ahelwer, line 28: Similarly add a javadoc link to the error code domain on this method.

ahelwer (approved):
Looks good! Only nits. You'll need to implement DCO signoff on your commits. See the failing GitHub merge validation check for more info, it should give you instructions on how to do that.

tlatools/org.lamport.tlatools/src/tlc2/output/MP.java
lemmy: Could we handle this within ModelConfig by invoking printError and exiting when no configuration file is found? This would allow MP to remain free of CFE.

ahelwer: It's usually nicer to separate error reporting from error presentation (as in it's nicer for library usage reasons to avoid printing things out far down the call stack) but maybe it is still a good idea here.

mvanhorn:
Thanks @ahelwer and @lemmy. Force-pushed 91287c7 on top of the two existing fix commits:

- DCO: rebased with `--signoff` so all three commits now carry the Signed-off-by trailer.
- 91287c7 adds javadoc on both the `errorCode` field and `getErrorCode()` pointing at `tlc2.output.EC`, per @ahelwer's nits.

@ahelwer's second nit: the pattern-match simplification on `MP#1717` depends on #1379. Happy to fold it in as soon as that lands, or in a follow-up PR if you'd rather keep this one focused.

@lemmy's suggestion (handle the missing-config case in `ModelConfig` via `printError` + exit so `MP` stays free of `ConfigFileException`) is the cleaner split. @ahelwer's caveat about separating error reporting from presentation is fair, but `ModelConfig` is already doing I/O and throwing, so having it call `printError` + exit for this one routine user case doesn't blur that line much. Want me to do that refactor in this PR, or land this one narrow (as originally scoped against #1234) and open a follow-up for the `MP`/`ModelConfig` cleanup?

lemmy:
> Want me to do that refactor in this PR, or land this one narrow (as originally scoped against https://github.com/tlaplus/tlaplus/issues/1234)

Yes, please.

mvanhorn:
Thanks @lemmy - pushed 7e9637e with the refactor:

- `ModelConfig#parse`: both CFG_ERROR_READING_FILE throw sites (file-not-found at the top of the try block, and the IOException-wrap in the catch) now call `MP.printError(EC.CFG_ERROR_READING_FILE, ...)` + `System.exit(EC.ExitStatus.errorConstantToExitStatus(EC.CFG_ERROR_READING_FILE))` directly. `ConfigFileException` no longer propagates out of the config-read path.
- `MP.ECGeneralMsg`: the special case that unwrapped `ConfigFileException` when `getErrorCode() == CFG_ERROR_READING_FILE` is reverted. `import tlc2.tool.ConfigFileException` removed from MP.java.
- `ConfigFileException`: the `errorCode` field and `getErrorCode()` method are no longer consulted anywhere (grep clean across `tlatools/`), so they're removed. The class is still thrown for CFG_MISSING_ID / CFG_TWICE_KEYWORD / CFG_EXPECT_ID / CFG_EXPECTED_SYMBOL / CFG_GENERAL, which continue to propagate through the existing 'TLC threw an unexpected exception' wrapper path - same behavior as before the PR.

Compiles clean with `javac 25`. The full ant-based build and test matrix isn't wired up on this machine, so I'm relying on TLA+ CI to exercise the test suite end-to-end.

lemmy (approved):
LGTM, but please revert the formatting changes in `CFE`.

mvanhorn:
Done in 38fbd3d. ConfigFileException.java is restored to exactly match master, so the PR diff is now just the ModelConfig.java change. Sorry for the churn, and thanks for the review.

lemmy:
@mvanhorn Thanks!

mvanhorn:
Thank you for the merge, @lemmy. Config errors reading as config errors instead of 'unexpected exception' should save folks some head-scratching.

Issue #136, opened 2018-01-30
PlusCal does not allow INSTANCE with multiple definitions


hwayne:
Setup:

Create one file `foo`
```
---- module foo ----
CONSTANTS Foo1, Foo2
====
```

Create second file `bar` as root module in the spec.

```
---- module bar ----
(*--algorithm bar
define
INSTANCE foo WITH Foo1 <- {}, Foo2 <- {}
end define;
begin
skip;
end algorithm; *)
```

Attempt the compile the pluscal.

Expected behavior: the pluscal should compile into TLA+.

Actual behavior: 

> Unrecoverable Error
> -- Expected "end" but found ","

Notes:
* Does not occur if you only assign one of the constants in `Foo`. `INSTANCE foo WITH Foo1 <- {}` will successfully compile (although it will fail to parse due to the missing substitution).
* Does not occur if you move the `INSTANCE` out of the pluscal and into the top of the file. The reason I can't do this is because I want to substitute in a variable in my actual spec.

This was done on Toolbox version 1.5.5 and should not be related to the new hotfixes.

bgamari:
Indeed my first attempt to use PlusCal on a real project ran head-long into this issue.

dirtybytes:
I'm also affected by this, is there a workaround?

lemmy:
https://github.com/tlaplus/tlaplus/pull/1388 addresses this limitation. This is a call to action: with sufficient confidence in the PR, for example through additional reviews, this limitation could be resolved soon.

lemmy:
Fixed with https://github.com/tlaplus/tlaplus/commit/5d411bfabdd2d7d4c0dc3b8b34494c5fd55f74c6

Issue #1224, opened 2025-08-05
PlusCal translator incorrectly renames bound variable


muenchnerkindl:
The following PlusCal algorithm illustrates how a bound identifier inside a macro can be incorrectly identified with a parameter of the code at which the macro is used, and then incorrectly replaced by the expression with which the parameter is instantiated. This leads to a semantically incorrect TLA+ specification that may be hard to debug. Renaming the bound identifier ``id`` in macro ``release`` below to a fresh name (say, ``x``) fixes the problem.

The basic problem underlying this issue is similar to https://github.com/tlaplus/tlaplus/issues/505.

```
----------------------------- MODULE PlusCalBug -----------------------------
(***************************************************************************)
(* The following PlusCal code illustrates a bug where a bound variable     *)
(* appearing in a macro is confused with an identifier present at the      *)
(* place where the macro is used.                                          *)
(***************************************************************************)

CONSTANTS Proc 

(*
--algorithm Bug {
    variables waiting = {};

    macro wait(id) {
        waiting := waiting \union {id}
    }

    macro release() {
        if (waiting # {}) {
            with (id \in waiting) { waiting := waiting \ {id} }    
        }
    }

    macro do(id) {
        either { wait(id) }
        or     { release() }
    }

    process (p \in Proc) {
p0:     while (TRUE) {
            do(self);
        }
    }
}
*)
\* BEGIN TRANSLATION (chksum(pcal) = "dc8dea26" /\ chksum(tla) = "34ea0e01")
VARIABLE waiting

vars == << waiting >>

ProcSet == (Proc)

Init == (* Global variables *)
        /\ waiting = {}

p(self) == \/ /\ waiting' = (waiting \union {self})
           \/ /\ IF waiting # {}
                    THEN /\ \E id \in waiting:
                              waiting' = waiting \ {self} \* self --> id !!!
                    ELSE /\ TRUE
                         /\ UNCHANGED waiting

Next == (\E self \in Proc: p(self))

Spec == Init /\ [][Next]_vars

\* END TRANSLATION 
```

Issue #505, opened 2020-09-18
PlusCal translator rewrites "LAMBDA" operator argument name


Calvin-L:
In this specification, `LAMBDA x: 1` is incorrectly rewritten into `LAMBDA x[self]: 1`. The PlusCal translator should either throw an exception or it should report an error due to name shadowing.

PlusCalLambdaRename.tla:
```TLA
---- MODULE PlusCalLambdaRename ----

\* "dummy" operator that might take a LAMBDA argument
Apply(object, P(_)) = P(object)

(*

--algorithm algo {

  process (myself \in {"myself"})
  variable x;
  {
    "start":
    x := Apply(x, LAMBDA x: 1)
  }

}

*)

------------------------------------

\* BEGIN TRANSLATION PCal-9ce38a181cb523f03090ec0d193ed058
CONSTANT defaultInitValue
VARIABLES pc, x

vars == << pc, x >>

ProcSet == ({"myself"})

Init == (* Process myself *)
        /\ x = [self \in {"myself"} |-> defaultInitValue]
        /\ pc = [self \in ProcSet |-> "start"]

start(self) == /\ pc[self] = "start"
               /\ x' = [x EXCEPT ![self] = Apply(x[self], LAMBDA x[self]: 1)]
               /\ pc' = [pc EXCEPT ![self] = "Done"]

myself(self) == start(self)

(* Allow infinite stuttering to prevent deadlock on termination. *)
Terminating == /\ \A self \in ProcSet: pc[self] = "Done"
               /\ UNCHANGED vars

Next == (\E self \in {"myself"}: myself(self))
           \/ Terminating

Spec == Init /\ [][Next]_vars

Termination == <>(\A self \in ProcSet: pc[self] = "Done")

\* END TRANSLATION TLA-3532fee2e0cb8fcf85a1bff6b4d21cc3

====================================
```

lemmy:
@Calvin-L Thanks for reporting. Can you also provide a PR that renames the argument similar to how identical labels or variables in different procedures are renamed?

Calvin-L:
Sure, I'll see if I can put together a PR over the weekend.

Calvin-L:
[...]

PR #1388, opened 2026-05-11
Allow INSTANCE with multiple substitutions in PlusCal `define` blocks.

lemmy:
The PlusCal translator rejected legal TLA+ definitions inside a `define` block when they contained a top-level comma, e.g.

    define
        Inner == INSTANCE M WITH A <- 1, B <- 2
    end define;

failed with "Unrecoverable error: Expected `end' but found `,'". The single-substitution form `... WITH A <- 1` was accepted, and the same INSTANCE statement was accepted outside the `define` block, so the restriction was both inconsistent and undocumented. Section 3.6 of the PlusCal manual describes the body of `define` as "a sequence of TLA+ definitions" copied verbatim into the translation, with no exception for INSTANCE.

The root cause is in `Tokenize.TokenOut`, which terminates an expression on any top-level `,` because the tokenizer was designed for PlusCal operand positions (e.g. `if (...)`, `with (x \in S, ...)`) where a top-level comma legitimately ends the expression. When `define` was retrofitted onto this tokenizer in 2006, that rule was inherited even though the body of `define` is module-unit territory in which top-level commas are part of valid TLA+ constructs.

Add a new entry point `Tokenize.TokenizeDefineBlock` that sets an `inDefineBlock` flag (parallel to the existing `inQuantifier`) which suppresses the comma terminator while a `define` body is being read. All other terminators (`end`, `}`, `;`, `:=`, `||`, ...) are unchanged, so the block still ends at the expected `end define` (p-syntax) or `}` (c-syntax). `ParseAlgorithm.GetDefineExpr` routes the single `define`-body call site through the new entry point; the other 15 `GetExpr` call sites are unchanged. Downstream stages already treat the body as opaque text, so no further changes are required.

Fixes Github issue #136
https://github.com/tlaplus/tlaplus/issues/136

[Bug][PlusCal]


ahelwer:
Could be a good idea to call SANY instead here, and ask it to parse/consume a series of definitions.

lemmy:
As far as I’m aware, the PlusCal translator does not currently depend on SANY (beyond Region and Validator).

It takes a PlusCal algorithm, together with any TLA+ definitions, and performs a syntactic translation into TLA+. If the generated TLA+ is invalid, SANY will catch that in the subsequent parsing/semantic-analysis step.

ahelwer:
It performs a not-very-good parse of TLA+ for things like definitions so it can skip over them and get back to the PlusCal; that is the cause of this bug. If this were replaced with calls to SANY instead (using the relatively new incremental parse approach), it would be much more robust and foreclose further issues of this type.

lemmy:
I don't have the bandwidth to expand the scope of this PR.

Calvin-L (approved):
This change looks fine and fixes the immediate issue.

(I do agree with @ahelwer though: the PlusCal translator would benefit from more robust parsing. See e.g. #505 and #1224, both of which would be easier to fix if PlusCal had a proper syntax tree for TLA+ expressions instead of just a list of tokens.)

Issue #1389, opened 2026-05-15
Recursive operators cannot be used to define a property


craft095:
Recursive operators cannot be used to define properties:

This spec works good:

```tla
---- MODULE M0 ----
VARIABLE x
ap_0(u) == [][TRUE]_x
Spec == x=TRUE /\ [][x'=~x]_x
Prop == ap_0(TRUE)
====
```

This one results in error: "TLC cannot handle the temporal formula"

```tla
---- MODULE M0 ----
VARIABLE x
RECURSIVE ap_0(_)
ap_0(u) == IF u THEN [][TRUE]_x ELSE ap_0(~u)
Spec == x=TRUE /\ [][x'=~x]_x
Prop == ap_0(TRUE)
====
```

Related issue #720

ahelwer:
I don't know whether this should actually be allowed. The level checker is supposed to reject any recursive operators higher than variable level, since it is too easy to write a double-primed expression otherwise. I would have assumed this would be rejected at the SANY level, not TLC.

lemmy:
Changed this to an enhancement request, since this behavior has never been supported.

ahelwer:
The deciding factor here seems to be whether a parameter to the recursive operator is given as an argument to an action-level or higher operator. So this is rejected by SANY with error code `RECURSIVE_OPERATOR_PRIMES_PARAMETER (4290)`:
```tla
---- MODULE Test ----
VARIABLE v
RECURSIVE op(_)
op(x) ≜ □[TRUE]_x
====
```
But this is not:
```tla
---- MODULE Test ----
VARIABLE v
RECURSIVE op(_)
op(x) ≜ □[TRUE]_v
====
```
I suppose this makes sense and cannot think of a way to use these rules to make a double-prime. So this feature seems plausibly useful.

lemmy:
@craft095 In your second example, the guard `u` is constant-level. Do you have use cases where `u` is above constant-level?

lemmy:
Implemented in https://github.com/tlaplus/tlaplus/pull/1390 and awaiting reviews.

craft095:
> [@craft095](https://github.com/craft095) In your second example, the guard `u` is constant-level. Do you have use cases where `u` is above constant-level?

No, I don't think so.

lemmy:
> > [@craft095](https://github.com/craft095) In your second example, the guard `u` is constant-level. Do you have use cases where `u` is above constant-level?
> 
> No, I don't think so.

[...]

PR #1014, opened 2024-09-10
Refinement checking for INSTANCEs with arguments.

lemmy:
Addresses Github issue #687
https://github.com/tlaplus/tlaplus/issues/687

[Feature][TLC][Changelog]

PR #1390, opened 2026-05-18
Allow RECURSIVE operators to define liveness properties.

lemmy:
TLC rejected a RECURSIVE operator used as a temporal property with "TLC cannot handle the temporal formula", even when the recursion reaches a base case at translation time via a constant-evaluable IF guard. The reproducer from the issue:

```tla
    RECURSIVE ap_0(_)
    ap_0(u) == IF u THEN [][TRUE]_x ELSE ap_0(~u)
    Prop == ap_0(TRUE)
```

The non-recursive variant of this pattern was fixed in PR #1014 for parameterized instantiation, but RECURSIVE definitions remained unsupported.

Expand a RECURSIVE operator when its call resolves to a temporal-level body, and short-circuit constant-evaluable IF guards so recursive expansion reaches its base case. If the recursion does not reach a base case at tableau construction time, the JVM raises a StackOverflowError, which TLC reports under the existing EC.SYSTEM_STACK_OVERFLOW error code.

Keep state- and action-level formulas opaque, preserving runtime expansion and evaluation against concrete states.

Addresses Github issue #1389
https://github.com/tlaplus/tlaplus/issues/1389

[Feature][TLC][Changelog]


tlatools/org.lamport.tlatools/src/tlc2/tool/liveness/Liveness.java
```diff
@@ -217,7 +264,25 @@ private static LiveExprNode astToLiveAppl(ITool tool, OpApplNode expr, Context c
...
+						// such fallback: Tool.eval returns an IValue and
+						// rejects temporal operators, so the tableau's
+						// LiveExprNode subtree must come from expansion
+						// alone. Re-throw rather than fall through.
+						// Re-throw also prevents an enclosing OPCODE_ite
+						// / OPCODE_case from catching the failure and
+						// retrying via the disjunctive expansion, which
+						// would re-enter the recursion.
+						if (recursive) {
+							throw e;
```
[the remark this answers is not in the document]
lemmy: Closing because bogus.

tlatools/org.lamport.tlatools/src/tlc2/tool/liveness/Liveness.java
```diff
@@ -352,6 +417,35 @@ private static LiveExprNode astToLiveAppl(ITool tool, OpApplNode expr, Context c
 		}
 		case OPCODE_ite: // IfThenElse
 		{
+			// If the guard evaluates to a constant boolean, translate
```
craft095, line 437: Can the same short-circuit be applied to `=>`, `\/`, `/\` for completeness?

Examples:

```tla
RECURSIVE ap(_)
ap(u) == 
    \/  u /\ [][TRUE]_x
    \/ ~u /\ ap(~u)
Prop  == ap(TRUE)
```

```tla
RECURSIVE ap(_)
ap(u) == 
    /\  u => [][TRUE]_x
    /\ ~u => ap(~u)
Prop  == ap(TRUE)
```

tlatools/org.lamport.tlatools/test-model/Github1389Violated.tla
```diff
@@ -0,0 +1,32 @@
...
+\* the constant-guard handling of OPCODE_ite) to <>(x = 99).  Since x
+\* toggles deterministically over {0, 1}, no reachable behavior ever has
+\* x = 99, so the property fails.  <> (rather than []) is used so
+\* SpecProcessor cannot route the formula to invariant checking; it must
+\* flow through the liveness tableau and surface as
+\* TLC_TEMPORAL_PROPERTY_VIOLATED.  WF_x(Next) forces progress so the
+\* counterexample is the deterministic two-state cycle 0 -> 1 -> 0
+\* rather than a trivial stuttering trace.
+RECURSIVE Reach(_)
+Reach(n) == IF n = 0 THEN <>(x = 99) ELSE Reach(n - 1)
```
craft095, line 25: Will this variant work:

```
RECURSIVE Reach(_)
Reach(n) == IF n = 0 THEN (x = 99) ELSE <>(Reach(n - 1))
```

lemmy: Yes, see newly added tests.

tlatools/org.lamport.tlatools/src/tlc2/tool/liveness/Liveness.java
lemmy: Scoping this PR to `OPCODE_ite` only:

1. IF is the canonical form for recursive operators in TLA+: every recursive example in Specifying Systems uses IF, Apalache normalizes to IF before any other pass, and the issue's reproducer uses IF.
2. Every other branch-selection construct is reducible to IF (a /\ b ≡ IF a THEN b ELSE FALSE, etc.). Users who hit the omitted case have a one-line rewrite.
3. Each additional opcode is ~30 lines of near-duplicated code plus its own tests. Keeping the change to one opcode keeps the review surface and per-opcode duplication minimal.

tlatools/org.lamport.tlatools/src/tlc2/tool/liveness/Liveness.java
```diff
@@ -192,12 +192,62 @@ private static LiveExprNode astToLiveAppl(ITool tool, OpApplNode expr, Context c
...
+				// separate concern from this decision: it depends on the
+				// body's IF guard evaluating to a constant during tableau
+				// construction.  See the OPCODE_ite handling below.  If a
+				// guard depends on a state variable the JVM eventually
+				// raises a StackOverflowError that TLC.process translates
+				// into EC.SYSTEM_STACK_OVERFLOW.  Other branch-selection
+				// constructs (CASE, /\, \/, =>) are not handled here;
+				// users should write recursive operators with IF (the
+				// most natural form for recursion in TLA+, and every
+				// other branch-selection construct is reducible to it).
```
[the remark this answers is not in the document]
lemmy: Update PR description

tlatools/org.lamport.tlatools/src/tlc2/tool/liveness/Liveness.java
lemmy: Marking resolved after offline discussion.

tlatools/org.lamport.tlatools/test-model/Github1389Violated.tla
lemmy: Marking resolved after offline discussion

lemmy:
TODOs:
- [x] Document that support is currently limited to ITE; CASE, /\\, \\/, and => remain unsupported for now
- [x] Add a test case for a "counting safety property", which is the primary use case this PR is intended to support. @craft095 will share the property

tlatools/org.lamport.tlatools/src/tlc2/tool/liveness/Liveness.java
```diff
@@ -192,12 +192,62 @@ private static LiveExprNode astToLiveAppl(ITool tool, OpApplNode expr, Context c
...
+				boolean expand = opcode == 0;
+				boolean recursive = false;
+				if (expand && opDef.getInRecursive()) {
+					recursive = true;
+					// Use ITool#getLevelBound (same helper as line 115);
+					// only expand when the level exceeds ActionLevel.
+					expand = tool.getLevelBound(expr, con, tool.getId()) > LevelConstants.ActionLevel;
+				}
```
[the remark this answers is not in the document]
lemmy: S`ymbolNodeValueLookupProvider#getLevelBoundAppl` does take the max over all argument levels, but the failure mode is unreachable: SANY's `OpApplNode.levelCheck` rejects any RECURSIVE-operator call whose arguments exceed action level. With args <= ActionLevel guaranteed, the two checks agree on every accepted spec.

`tool.getLevelBound` is also used elsewhere in Liveness, so this PR is consistent with the surrounding code.

PR #1398, opened 2026-07-24
Fix POSIX checkpoint atomicity and durability

marete:
This PR comprehensively overhauls the TLC checkpointing architecture to guarantee data durability and structural atomicity on POSIX filesystems, explicitly mitigating split-brain state corruption resulting from hard terminations (e.g., AWS Spot instance reclamation).

Includes:
- POSIX Durability fsync barriers and SyncingFileOutputStream injections.
- Atomic Write-Ahead-Log (WAL) Protocol for ModelChecker.
- Zero-Copy Checkpoint Rollbacks via zero-byte hardlinks.
- Renaming Atomicity natively utilizing StandardCopyOption.ATOMIC_MOVE.
- Refactored DiskStateQueue and DiskByteArrayQueue to eradicate rogue asynchronous StatePoolCleaner threads during crash recovery.


lemmy:
Thanks for the PR. The StateQueue is among the most sensitive parts of TLC, so we will need strong evidence that these changes preserve correctness, especially around concurrency and consistency.

Please take a look at the existing verification work in the repo, in particular tlatools/org.lamport.tlatools/test-verify/tlc2/tool/queue/StateQueueJPFTest.java and the JPF setup under tlatools/org.lamport.tlatools/test-verify/. The repo already contains a Java Pathfinder model for StateQueue, and changes in this area should either be covered by that model or come with a clear explanation of why the model does not apply and what equivalent evidence supports the change.

tlatools/org.lamport.tlatools/src/tlc2/tool/queue/IStateQueue.java
```diff
@@ -113,6 +113,8 @@ public static IStateQueue get(final String metaDir) {
 	public abstract void beginChkpt() throws IOException;
 
 	public abstract void commitChkpt() throws IOException;
+	
+	public abstract void pruneOldPools() throws IOException;
```
Calvin-L: This is one of the most important interfaces in the tools, and I already think it's choked with unnecessary/redundant methods with insufficient documentation.

My top choice would be to make cleanup tasks like this one the responsibility of _each implementation_ rather than exposing this globally to clients.

If that isn't reasonable, my second choice would be to rename this to `cleanup` or `vacuum` and give it a very clear docstring about what it's allowed to do and when it should be called.

tlatools/org.lamport.tlatools/src/tlc2/tool/ModelChecker.java
```diff
@@ -742,6 +742,32 @@ protected void checkpoint() throws IOException {
 		{
 			liveCheck.beginChkpt();
 		}
+
+        // ---- ATOMIC COMMIT MANIFEST WAL ----
```
Calvin-L: This kind of protocol is actually a perfect use case for TLA+. :) It might be worth writing a spec to demonstrate that this protocol really does ensure crash-safety.

tlatools/org.lamport.tlatools/src/tlc2/tool/liveness/AbstractDiskGraph.java
```diff
@@ -1,689 +1,689 @@
...
+		} catch (IOException e) {
+			e.printStackTrace();
+		}
+		this.destroyCache();
+	}
+
+	/* Checkpoint. */
+	public synchronized final void beginChkpt() throws IOException {
+		this.nodeRAF.flush();
+		this.nodeRAF.getFD().sync();
```
Calvin-L: I think there's a good case for making this `sync()` call part of the `flush()` implementation. Obviously that carries a greater risk of performance regressions, but it means we won't have to think quite as hard about whether each `flush()` call site deserves a corresponding `sync()`.

tlatools/org.lamport.tlatools/src/tlc2/tool/queue/DiskByteArrayQueue.java
```diff
@@ -61,13 +61,7 @@ public class DiskByteArrayQueue extends ByteArrayQueue {
...
 	protected ByteArrayPoolWriter writer;
 	
-	/**
-	 * The SPC takes care of deleting swap files on the lower end of the range
-	 * (loPool, hiPool). It terminates, when the first checkpoint is written at
-	 * which point checkpointing itself takes care of removing obsolete swap
-	 * files.
-	 */
-	protected final StatePoolCleaner cleaner;
+	protected boolean isPoolCleaningEnabled = true;
```
Calvin-L: I don't quite follow the purpose of this field. It looks like some methods set it to `false` but there is no way for it to ever become `true` again.

tlatools/org.lamport.tlatools/src/util/FileUtil.java
```diff
@@ -48,6 +48,43 @@ public class FileUtil
...
+        }
+        public SyncingFileOutputStream(String name, boolean append) throws FileNotFoundException {
+            super(name, append);
+        }
+        @Override
+        public void close() throws IOException {
+            try {
+                this.getFD().sync();
+            } catch (java.io.SyncFailedException e) {
+            }
```
Calvin-L: Two points here.

First and most importantly: I'm quite certain this exception should not be silently discarded. The block should instead look like

```java
try {
    this.getFD().sync();
} finally {
    super.close();
}
```

Second: I think it's good practice to call `this.flush()` first, in case `FileOutputStream` has some private in-memory buffer. `getFD().sync()` would bypass any such buffer; the buffered bytes would be written to the kernel in `close()` and would not necessarily have reached durable storage.

tlatools/org.lamport.tlatools/src/util/FileUtil.java
```diff
@@ -48,6 +48,43 @@ public class FileUtil
...
+    public static void syncDirectory(String path) {
+        try {
+            java.nio.channels.FileChannel channel = java.nio.channels.FileChannel.open(
+                java.nio.file.Paths.get(path), 
+                java.nio.file.StandardOpenOption.READ
+            );
+            channel.force(true);
+            channel.close();
+        } catch (IOException e) {
+            // Ignore if we can't fsync the directory (e.g. not supported by OS)
```
Calvin-L: If we're serious about doing this, we should not blindly discard these exceptions.

It's unfortunate that [`FileChannel.force()`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileChannel.html#force-boolean-) does not throw a special kind of exception for "unsupported". :( Even so, I think it's best _not_ to catch anything here, provided we can show this works on Windows/NTFS. (I know from experience that this will work on MacOS/HFS and on Linux/most filesystems).

tlatools/org.lamport.tlatools/src/util/FileUtil.java
```diff
@@ -48,6 +48,43 @@ public class FileUtil
      * Parses the directory path from a filename. If the filename
      * is already a basename, returns the empty string.
      */
+    
+    public static void syncDirectory(String path) {
```
Calvin-L: Ah! ⚠️ 

This is incorrect for a very subtle reason:

 > in 4.13 and later kernels, [a program that calls fsync()] will not see any errors that happened before it opened the file

https://lwn.net/Articles/752063/
https://calvin.loncaric.us/articles/CreateFile.html

So

```
rename(dir/f1, dir/f2);
syncDirectory(dir);
```

will not expose any disk errors that occur _between_ the `rename()` and the `syncDirectory()` call.

This is a nightmare for application developers like us, but it's really the kernel's only reasonable choice: if a disk error happens in that window, there is no open userspace file descriptor to report the error to, so the only options are (1) record the error in memory and keep it forever on the off chance that some application opens the file again or (2) log the error, forget about it, and keep going. Of these, option 2 is the better choice, since there's no guarantee any application will ever open the file again.

Calvin-L (commented):
There's a lot to like about this change. In particular, replacing `if exists { delete }; rename` with a single atomic rename call and proper exception propagation seems like a strict win for performance _and_ reliability _and_ code maintainability. (If those changes were in a standalone PR, I would accept them without question.)

BUT: If we're going to take on crash safety as a requirement, we might have a lot more thinking ahead of us. I've done enough messing around with Java and POSIX APIs to know that this can be a nightmare; guarantees at the POSIX filesystem layer are hard to come by.

To really handle crash safety, I would prefer to take a dependency on some library that already solves this problem instead of rolling our own WAL built on filesystem metadata operations. For example, [SQLite might be a good choice](https://sqlite.org/fasterthanfs.html).

If we're set on doing filesystem operations for now, then I've left a few comments on some specific things I spotted. Most of them are just opinions, but I think the comment on `syncDirectory()` might actually be important.

marete:
I will open another PR with additional fixes and further tests.

marete:
Sorry about the noise!

PR #1400, opened 2026-08-08
Document `_POSSIBLE` and name the unwitnessed predicate

lemmy:
`_POSSIBLE` was undocumented, and an unwitnessed possibility condition failed under the generic `TLC_POSTCONDITION_FALSE` without saying which predicate went unwitnessed.

- Add `EC.TLC_POSSIBLE_UNWITNESSED`, reported with the predicate's name and location. Which predicate is named when several are unwitnessed is unspecified.
- Declare `minLevel = VariableLevel` on the `_Counts` override, so a `POSTCONDITION` reading only `_Counts` is no longer folded to the empty record at startup.
- Document both statements in `current-tools.md` and add `docs/possible-conditions.md`, the page the new error links to.

Part of #860

### DIFF
--- a/docs/possible-conditions.md
+++ b/docs/possible-conditions.md
@@ -1,0 +1,73 @@
+# Possibility Conditions with TLC's `_POSSIBLE`
+
+## Why check that something *can* happen
+
+When TLC reports that all invariants and properties hold, the result may be misleading if the model omits intended behaviors. A specification that permits only stuttering steps, an action whose enabling condition is too strong, or a model whose constant assignments are too small can cause TLC to explore too little while reporting no error. Always be suspicious of success.
+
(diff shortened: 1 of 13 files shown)


docs/possible-conditions.md
```diff
@@ -0,0 +1,73 @@
...
+
+```
+[BufferFull |-> 217, LeaderElected |-> 0]
+Error: The _POSSIBLE predicate LeaderElected at line 42, col 21 to line 42, col 38 of module MySpec was never witnessed.
+Others may be unwitnessed too; only witnessed predicates appear with a
+non-zero count in the record printed above.
+See https://explain.tlapl.us/possible-conditions for additional details.
+```
+
+Here no state satisfies `LeaderElected`, or no step does if it is an action. TLC names one unwitnessed condition and stops there; which one is unspecified. Any number of the others can be unwitnessed too, and only the record identifies them all.
```
craft095, line 32: Non-blocking suggestion: it would be better to report all unwitnessed predicates, not just one. It looks like we have all necessary information to do so (_Counts map).

lemmy: I'm worried that this will become unwieldy.  Let's leave it for another day.

PR #1401, opened 2026-08-11
SANY: reproducers for error formatting, subexpression, and XML export bugs

lemmy:
Reproducers requested in https://github.com/tlaplus/tlapm/pull/275#issuecomment-5241074153 (comment), reduced from the syntax corpus entries that triggered them. One test class per commit; 7 of 8 tests fail, deliberately not `@Ignore`d.
  - **`TestErrorMessageFormatting`** — `Errors::getMessage` runs message text through `String.format`, so `x % y` and `A!B!%(x, y)` crash with `UnknownFormatConversionException`, and `x %% y` is silently reported as an error about `%`.
  - **`TestSubexpressionSelectors`** — `op!<<!>>` reaches an `INTERNAL_ERROR` check in `Generator::selectorToNode` and aborts the parse.
  - **`TestXMLExporterStringEscapes`** — a string containing `\f` cannot be exported: XML 1.0 character data cannot hold U+000C. Ref #1313


lemmy:
@konnov Could you review this PR?

PR #1404, opened 2026-08-14
Prevent disk pool writers from exiting on empty wakes

### LINKED ISSUE #1403: Disk state queue writer can exit on an empty wake and deadlock TLC
### Description

`StatePoolWriter.run()` exits whenever `wait()` returns while `poolFile` is still null:

```java
while (this.poolFile == null) {
    this.wait();
    if (this.poolFile == null) {
        return;
    }
}
```

Because `Object.wait()` permits spurious wakeups, the writer can terminate even though the queue has not been shut down. A later disk spill then leaves a worker blocked forever in `StatePoolWriter.ensureWritten()` while holding the `DiskStateQueue` monitor, with the remaining workers blocked behind it.

The same pattern exists in `DiskByteArrayQueue.ByteArrayPoolWriter`.

### Observed failure

This occurred during a real exhaustive TLC 1.8.0 run after approximately 5 days and 14 hours:

- 402,069,694,759 states generated
- 29,916,698,581 distinct states
- depth 77
- 19,423 states left on the queue

CPU and disk activity had gone idle. A thread dump showed one worker waiting in `StatePoolWriter.ensureWritten()` while holding the `DiskStateQueue` monitor, all other workers blocked on that monitor, and no live `TLCStatePoolWriter` thread.

After changing the writer to continue waiting after an empty wake, checkpoint recovery completed the remaining state space successfully with zero states queued and no model error.

### Reproducer

This reproduces on current `master` at `0894c34`:

```java
StatePoolWriter writer = new StatePoolWriter(1);
writer.setDaemon(true);
writer.start();

while (writer.getState() != Thread.State.WAITING) {
    Thread.sleep(10);
}

synchronized (writer) {
    writer.notifyAll();
}

Thread.sleep(100);
assertTrue(writer.isAlive()); // fails: writer has exited
```

### Shutdown caveat

Simply removing the return is not a complete fix. `DiskStateQueue.finishAll()` currently relies on an empty `notifyAll()` to terminate the writer, so the writer needs an explicit shutdown state that is distinguishable from a spurious or unrelated notification.

A proposed fix is to:

1. add an explicit finished flag/method;
2. wait while there is no file and the writer is not finished;
3. exit only when explicitly finished;
4. use that method from `DiskStateQueue.finishAll()`;
5. apply the equivalent change to `ByteArrayPoolWriter`; and
6. add regression coverage for both an empty wake and explicit shutdown.

pdeva:
Fixes #1403.

## Root cause

`StatePoolWriter` treated every wake with no pending pool file as a shutdown request. Java permits `Object.wait()` to wake spuriously, so the writer could exit during an active model-checking run. A later disk spill would then block forever in `ensureWritten()`. The raw-byte disk queue contained the same pattern.

The previous empty notification also doubled as the intentional shutdown mechanism in `finishAll()`, so merely removing the premature return would leave the daemon writers running after queue shutdown.

## Changes

- give both disk pool writers an explicit, persistent finished state
- make empty or spurious wakes return to waiting unless shutdown was requested
- have both disk queues explicitly finish their writer instead of sending an ambiguous empty notification
- add regression tests covering empty wakes and orderly shutdown for both writer implementations

## Impact

This prevents rare, permanent TLC hangs in large disk-backed state-space explorations while preserving writer shutdown behavior. The failure was observed in a real TLC 1.8.0 run after more than five days and 402 billion generated states. A locally patched writer allowed checkpoint recovery to drain the queue and complete without a model error.

## Validation

- reproduced the failure on `master` at `0894c34` (`writer_alive_after_empty_wake=false`)
- `ant -f customBuild.xml info compile compile-test test-set -Dtest.testcases="tlc2/tool/queue/DiskPoolWriterTest.java"`
- `ant -f customBuild.xml info compile compile-test test` with Temurin Java 11 (full suite, 3m26s)
- `ant -f customBuild.xml dist` with Temurin Java 11

### DIFF
--- a/tlatools/org.lamport.tlatools/test/tlc2/tool/queue/DiskPoolWriterTest.java
+++ b/tlatools/org.lamport.tlatools/test/tlc2/tool/queue/DiskPoolWriterTest.java
@@ -11,6 +11,8 @@ import java.nio.file.Files;
 import org.junit.After;
 import org.junit.Test;
 
+import util.FileUtil;
+
 public class DiskPoolWriterTest {
 
 	private static final long TIMEOUT_MILLIS = 5000L;
@@ -24,13 +26,7 @@ public class DiskPoolWriterTest {
 			this.queue.finishAll();
 		}
 		if (this.diskDirectory != null) {
-			final File[] files = this.diskDirectory.listFiles();
-			if (files != null) {
-				for (File file : files) {
-					file.delete();
-				}
-			}
-			this.diskDirectory.delete();
+			FileUtil.deleteDir(this.diskDirectory, true);
 		}
 	}
 
--- a/tlatools/org.lamport.tlatools/test/tlc2/tool/queue/DiskPoolWriterTest.java
+++ b/tlatools/org.lamport.tlatools/test/tlc2/tool/queue/DiskPoolWriterTest.java
@@ -1,0 +1,89 @@
+package tlc2.tool.queue;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.io.File;
+import java.lang.reflect.Field;
+import java.nio.file.Files;
+
+import org.junit.After;
+import org.junit.Test;
+
+public class DiskPoolWriterTest {
+
+	private static final long TIMEOUT_MILLIS = 5000L;
+
+	private File diskDirectory;
+	private IStateQueue queue;
+
+	@After
+	public void tearDown() {
+		if (this.queue != null) {
+			this.queue.finishAll();
+		}
+		if (this.diskDirectory != null) {
+			final File[] files = this.diskDirectory.listFiles();
+			if (files != null) {
+				for (File file : files) {
+					file.delete();
+				}
+			}
+			this.diskDirectory.delete();
+		}
+	}
+
+	@Test
+	public void testStatePoolWriterIgnoresEmptyWakeAndStopsOnFinish() throws Exception {
+		this.diskDirectory = Files.createTempDirectory("DiskPoolWriterTest").toFile();
+		final DiskStateQueue stateQueue = new DiskStateQueue(this.diskDirectory.getAbsolutePath());
+		this.queue = stateQueue;
+
+		assertIgnoresEmptyWake(stateQueue.writer);
+		assertStopsOnFinish(stateQueue.writer);
+	}
+
+	@Test
+	public void testByteArrayPoolWriterIgnoresEmptyWakeAndStopsOnFinish() throws Exception {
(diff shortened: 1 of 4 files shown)


tlatools/org.lamport.tlatools/test/tlc2/tool/queue/DiskPoolWriterTest.java
```diff
@@ -0,0 +1,89 @@
...
+			this.queue.finishAll();
+		}
+		if (this.diskDirectory != null) {
+			final File[] files = this.diskDirectory.listFiles();
+			if (files != null) {
+				for (File file : files) {
+					file.delete();
+				}
+			}
+			this.diskDirectory.delete();
```
Calvin-L: I believe you can make this simpler and more robust by using `FileUtil.deleteDir(diskDirectory, true)` instead.

Calvin-L (approved):
Thanks for finding and fixing this! 👍

tlatools/org.lamport.tlatools/test/tlc2/tool/queue/DiskPoolWriterTest.java
pdeva: Addressed in 73c5be4: test teardown now uses FileUtil.deleteDir(diskDirectory, true).

lemmy:
I don’t have time to investigate this issue further, but I remembered that we have a Java Pathfinder model for StateQueue. So I asked Opus 5 to create a model for #1403. It appears to reproduce a variant of the deadlock and also shows that the proposed fix eliminates it. The synthesized Java Pathfinder tests should still be reviewed carefully. https://github.com/tlaplus/tlaplus/pull/1405

Calvin-L:
@lemmy CI seems to be broken with unrelated errors like

```
ERROR:root:examples/deps/apalache/bin/apalache-mc check --length=5 --config=examples/specifications/SpecifyingSystems/FIFO/APInnerFIFO.cfg examples/specifications/SpecifyingSystems/FIFO/APInnerFIFO.tla
Error: LinkageError occurred while loading main class at.forsyte.apalache.tla.Tool
	java.lang.UnsupportedClassVersionError: at/forsyte/apalache/tla/Tool has been compiled by a more recent version of the Java Runtime (class file version 65.0), this version of the Java Runtime only recognizes class file versions up to 61.0
```

Did Apalache upgrade to a newer version of Java? I think those errors are unrelated to this PR.

lemmy:
> @lemmy CI seems to be broken with unrelated errors like
> 
> [...]
>
> Did Apalache upgrade to a newer version of Java? I think those errors are unrelated to this PR.

Yes, Apalache bumped its Java dependency.

https://github.com/apalache-mc/apalache/pull/3432

lemmy:
Thanks, this PR has been absorbed into https://github.com/tlaplus/tlaplus/pull/1405.

Issue #1407, opened 2026-08-29
Equality of function sets is unsound when a domain or co-domain is empty


lemmy:
## Description

`SetOfFcnsValue.equals` decides equality of two sets of functions by comparing domains and co-domains pairwise. This is unsound whenever an empty set is involved:

- If `S1 = S2 = {}`, then `[S1 -> T1]` and `[S2 -> T2]` are both `{<<>>}`, regardless of `T1` and `T2`.
- If `T1 = T2 = {}` and `S1, S2 # {}`, then both sets are empty, regardless of `S1` and `S2`.

TLC is internally inconsistent: fingerprints and `compareTo` report the two values as equal, and each is `\subseteq` the other, but `=` reports them as different. In other words, the fingerprint used for state-space dedup considers the two states identical while the next-state relation considers them distinct.

The sibling classes already get this right: `SetOfRcdsValue.equals` and `SetOfTuplesValue.equals` short-circuit on `isEmpty()` before comparing their components. `SetOfFcnsValue.equals` has no such guard.

## Steps to Reproduce

Constants:

```tla
---- MODULE EmptyFcnSet ----
EXTENDS Integers, TLC
ASSUME PrintT(<<"expect TRUE", [{} -> {1}] = [{} -> {2}]>>)
ASSUME PrintT(<<"expect TRUE", [{1} -> {}] = [{2} -> {}]>>)
ASSUME PrintT(<<"expect TRUE", [{} -> {1}] \subseteq [{} -> {2}]>>)
====
```

```
<<"expect TRUE", FALSE>>
<<"expect TRUE", FALSE>>
<<"expect TRUE", TRUE>>
```

The next-state relation is affected too, because the primed-variable case of `OPCODE_eq` in `Tool` compares an already bound `x'` with `Value#equals`:

```tla
---- MODULE EmptyFcnSetEq ----
VARIABLE x
Init == x = 0
DomainEmpty == /\ x' = [{} -> {1}]
               /\ x' = [{} -> {2}]
RangeEmpty == /\ x' = [{1} -> {}]
              /\ x' = [{2} -> {}]
Next == DomainEmpty \/ RangeEmpty
Spec == Init /\ [][Next]_<<x>>
====
```

Both actions are enabled, so the initial state has two successors. TLC instead reports `Deadlock reached` after one state.

## Expected Behavior

`TRUE` for all three assumptions, and two successor states for `EmptyFcnSetEq`.

## Actual Behavior

[...]

PR #1408, opened 2026-08-30
Equality of function sets is unsound when a domain or co-domain is empty

lemmy:
### Test suite

- **TLA+ level, TLC as a black box, TLAPS as the oracle.** `EmptySetEqCases.tla` names the equalities, finiteness questions, and cardinalities of sets of functions `[S -> T]`, sets of records `[h : S]`, and Cartesian products `S \X T` built from empty, non-empty, and infinite sets (`Nat`, `Int`, `STRING`, `Seq(S)`). `EmptySetEqAssume.tla` assumes them, `EmptySetEqCases_proofs.tla` proves each of them from the rules of `EmptySetEqTheorems.tla`. Assumption and theorem share one definition, so the two cannot drift apart.
- **Answer correctly or refuse.** Most assumptions state a case, the rest pin an error with `TLCExt!AssertError`. Some of those errors are facts TLAPS proves, so they record a limitation instead of an acceptable refusal.
- **State and action level.** `EmptySetEqStates.tla` offers every enumerable set as an initial state, so TLC reaches `Value#fingerPrint`, and a wrong answer becomes an extra state, a deadlock, or a violated `Inv`.
- **JUnit where no expression reaches the code.** `ValueTest` builds such values in Java, e.g. for the `\cap` arm of `Value#isEmpty`; `ValueTest.tla` proves the formulas behind its assertions.

### Fix

Every set the suite names is `{<<>>}` or `{}`, so one empty argument decides the answer whatever the others contain. TLC read the others anyway, wherever it decided equality, finiteness, cardinality, or printing.

- **Equality.** `Value#isEmpty` decides `[S -> T]`, `[h : S]`, and `S \X T` structurally instead of by enumerating, and a standard module answers for itself through the new `UserObj#isEmpty`. `SetOfFcnsValue#equals` settles an empty domain or an empty range first, as `SetOfRcdsValue#equals` and `SetOfTuplesValue#equals` have since 075246eea of 2014-06-09.
- **Finiteness.** `IsFiniteSet([{} -> Nat])` answered `FALSE` although `[{} -> Nat] = {<<>>}`, and TLC denied `[Nat -> {}]`, `[n1 : {}, n2 : Nat]`, and `{} \X Nat` the same way. An empty argument decides, in either order. So does a singleton range, because `[S -> {w}]` is the one function `[x \in S |-> w]` for an infinite `S`, which settles `IsFiniteSet([Nat -> {"d1"}])`; and `Seq({})` is `{<<>>}` without a bound on the length, which settles `IsFiniteSet(Seq({}))`.
- **Cardinality.** `Cardinality([n1 : 1..50000, n2 : 1..50000, n3 : {}])` reported an overflow and `Cardinality([{} -> Nat])` refused, though the empty argument decides both. An empty field set, component, domain, or range decides the count, as does a singleton range, and `needBigInteger` short-circuits on the same ones, because the cardinality of `{}` needs no `BigInteger`.
- **Printing.** `ToString([n1 : 1..50000, n2 : 1..50000, n3 : {}])` printed the expression that denotes the set rather than the set, and `ToString([{} -> SUBSET (1..40)])` overflowed. TLC prints `{}` and `{<<>>}` whether or not the remaining sets can be counted.

### Accepted limitations

Each is pinned with `AssertError`, and proved where TLA+ decides the fact, so it is recorded rather than hidden.

- **Printing a set whose other argument TLC cannot handle.** Printing sizes the set and then enumerates it, so TLC gives up on `ToString([n1 : Nat, n2 : {}])`, `ToString([{} -> Nat])`, and `ToString([(1..50000) \X (1..50000) -> {}])`. `Cardinality([n1 : (Nat \ {0}), n2 : {}])` gives up in any position, because `Nat \ {0}` cannot be enumerated at all.
- **A set a standard module cannot count.** `IsFiniteSet([Nat -> Seq({})])` refuses, although `Seq({})` is `{<<>>}` and the set is the singleton `{[x \in Nat |-> <<>>]}`. Deciding it asks `Sequences` for `Cardinality(Seq({}))`, which the module does not answer.
- **A comparison of a set with itself.** TLC refuses `[1 -> 2] = [1 -> 2]` and the like, which TLA+ decides by congruence, because the guards `SetOfFcnsValue#equals` gains ask whether `1` is empty; `SetOfRcdsValue` and `SetOfTuplesValue` have refused theirs since 075246eea. Deciding them needs an emptiness test that may answer neither.
- **An argument whose emptiness TLA+ leaves open.** `IsFiniteSet([1 -> 2])`, `IsFiniteSet(1 \X {"d1"})`, `IsFiniteSet(Seq(1))`, and the like are open in TLA+, so refusing is the answer, and TLC refuses instead of returning `FALSE`.

Every commit is green: what TLC gets wrong is commented out where the suite lands, and uncommented by the commit that fixes it.

Addresses #1407

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
@@ -151,10 +151,39 @@ public class SetOfFcnsValue extends SetOfFcnsOrRcdsValue implements Enumerable {
 
   // THEOREM ASSUME NEW S, NEW T, IsFiniteSet(S), IsFiniteSet(T)
   //         PROVE  IsFiniteSet([S -> T])
+  //
+  // THEOREM ESE_UnitFinite ==
+  //   ASSUME NEW S, NEW T, S = {}
+  //   PROVE  IsFiniteSet([S -> T])
+  //
+  // THEOREM ESE_EmptyRangeFinite ==
+  //   ASSUME NEW S, NEW T, T = {}
+  //   PROVE  IsFiniteSet([S -> T])
+  //
+  // THEOREM ESE_SingletonRangeFinite ==
+  //   ASSUME NEW S, NEW T, NEW w, T = {w}
+  //   PROVE  IsFiniteSet([S -> T])
+  //
+  // The first theorem misses { <<>> } and {}: an empty domain or co-domain
+  // makes [S -> T] finite whatever the other argument is, so neither argument
+  // has to be finite. isEmpty is only asked of an argument that isFinite has
+  // answered for, because an empty set is finite, and asking one that TLC
+  // cannot enumerate, such as Nat \ {0}, would fail instead of answering.
   @Override
   public final boolean isFinite() {
     try {
-      return this.domain.isFinite() && this.range.isFinite();
+      final boolean finiteDomain = this.domain.isFinite();
+      if (finiteDomain && this.domain.isEmpty()) {
+        return true;
+      }
+      final boolean finiteRange = this.range.isFinite();
+      if (finiteRange && this.range.isEmpty()) {
+        return true;
+      }
+      if (finiteDomain && finiteRange) {
+        return true;
+      }
+      return finiteRange && this.range.size() == 1;
     }
     catch (RuntimeException | OutOfMemoryError e) {
       if (hasSource()) { throw FingerprintException.getNewHead(this, e); }
--- a/tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
@@ -414,26 +414,29 @@ public class SetOfFcnsValue extends SetOfFcnsOrRcdsValue implements Enumerable {
   @Override
   public final StringBuffer toString(StringBuffer sb, int offset, boolean swallow) {
     try {
-      boolean unlazy = TLCGlobals.expand;
+      Value val = null;
       try {
-        if (unlazy) {
-          int dsz = this.domain.size();
-          int rsz = this.range.size();
-          long sz = 1;
-          for (int i = 0; i < dsz; i++) {
-            sz *= rsz;
-            if (sz < -2147483648 || sz > 2147483647) {
-              unlazy = false;
-              break;
+        if (TLCGlobals.expand) {
+          long sz;
+          if (this.domain.isEmpty()) {
+            sz = 1;
+          }
+          else {
+            final int dsz = this.domain.size();
+            final int rsz = this.range.size();
+            sz = 1;
+            for (int i = 0; i < dsz && sz <= 2147483647; i++) {
+              sz *= rsz;
             }
           }
-          unlazy = sz < TLCGlobals.enumBound;
+          if (sz < TLCGlobals.enumBound) {
+            val = this.toSetEnum();
+          }
         }
       }
-      catch (Throwable e) { if (swallow) unlazy = false; else throw e; }
+      catch (Throwable e) { if (!swallow) throw e; }
 
-      if (unlazy) {
-        Value val = this.toSetEnum();
+      if (val != null) {
         return val.toString(sb, offset, swallow);
       }
       else {
--- a/tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
+++ b/tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
@@ -56,10 +56,41 @@ public class SetOfFcnsValue extends SetOfFcnsOrRcdsValue implements Enumerable {
     }
   }
 
+  // THEOREM ESE_UnitDomain ==
+  //   ASSUME NEW S, NEW T, S = {}
+  //   PROVE  [S -> T] = { <<>> }
+  //
+  // THEOREM ESE_EmptyRange ==
+  //   ASSUME NEW S, NEW T, NEW w, w \in S, T = {}
+  //   PROVE  [S -> T] = {}
+  //
+  // THEOREM ESE_DomainDecides ==
+  //   ASSUME NEW S, NEW T, NEW R, NEW w, w \in R
+  //   PROVE  [S -> R] = [T -> R] <=> S = T
+  //
+  // THEOREM ESE_RangeDecides ==
+  //   ASSUME NEW S, NEW R, NEW Q, NEW w, w \in S
+  //   PROVE  [S -> R] = [S -> Q] <=> R = Q
+  //
+  // The witness w that the last two theorems need is what the two guards below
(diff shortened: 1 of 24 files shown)


tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
```diff
@@ -56,10 +56,41 @@ public final int compareTo(Object obj) {
...
+  // is. Either set being one of the two therefore settles the equality on its
+  // own, and only the sets that are neither are left for the domains and the
+  // co-domains to decide.
   public final boolean equals(Object obj) {
     try {
       if (obj instanceof SetOfFcnsValue) {
         SetOfFcnsValue fcns = (SetOfFcnsValue)obj;
+
+        boolean isUnit1 = this.domain.isEmpty();
+        boolean isUnit2 = fcns.domain.isEmpty();
```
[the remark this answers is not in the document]
lemmy: The emptiness test is not new. Commit `075246eea` from 2014 prefixed `SetOfRcdsValue#equals` and `SetOfTuplesValue#equals` with `isEmpty()`, before the pairwise comparison from which congruence follows.

On the parent of this branch, `[n1 : 1, n2 : {"d1"}] = [n1 : 1, n2 : {"d1"}]` and `(1 \X {"d1"}) = (1 \X {"d1"})` already fail, only with `"can't enumerate the value of the n1 field: 1"` rather than `"Shouldn't call isEmpty() on value 1"`. Thus, this PR changes the error *message*, but not the fact that TLC errors on these inputs.

What is new in this PR are the three `Fcn*` cases. `SetOfFcnsValue#equals` previously lacked the emptiness test, which is the root cause of #1407. In other words, `SetOfFcnsValue#equals` now behaves like `SetOfRcdsValue#equals` and `SetOfTuplesValue#equals` have behaved since 2014.

tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
```diff
@@ -177,9 +237,21 @@ public final Value takeExcept(ValueExcept[] exs) {
...
   @Override
   public final int size() {
     try {
+      if (this.domain.isEmpty()) {
+        return 1;
+      }
+      if (this.range.isEmpty()) {
+        return 0;
+      }
+      if (!this.domain.isFinite() && this.range.size() == 1) {
```
craft095: The first conjunct is not needed:
```
if (this.range.size() == 1) {
    return 1;
}
```

tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
```diff
@@ -200,6 +272,12 @@ public final int size() {
 
 	@Override
 	protected boolean needBigInteger() {
+		if (this.domain.isEmpty() || this.range.isEmpty()) {
+			return false;
+		}
+		if (!this.domain.isFinite() && this.range.size() == 1) {
```
craft095: The same: range check is enough.

tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
```diff
@@ -336,26 +414,29 @@ public final void write(final IValueOutputStream vos) throws IOException {
...
+        if (TLCGlobals.expand) {
+          long sz;
+          if (this.domain.isEmpty()) {
+            sz = 1;
+          }
+          else {
+            final int dsz = this.domain.size();
+            final int rsz = this.range.size();
+            sz = 1;
+            for (int i = 0; i < dsz && sz <= 2147483647; i++) {
```
craft095, line 428: `TLCGlobals.enumBound` could be used instead of `2147483647` (or min of these two numbers) to skip unnecessary iterations.

tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfRcdsValue.java
```diff
@@ -381,24 +399,28 @@ public final void write(final IValueOutputStream vos) throws IOException {
...
           for (int i = 0; i < this.values.length; i++) {
-            sz *= this.values[i].size();
-            if (sz < -2147483648 || sz > 2147483647) {
-              unlazy = false;
+            final int fsz = this.values[i].size();
+            if (fsz == 0) {
+              sz = 0;
               break;
             }
+            if (sz <= 2147483647) {
```
craft095: As in SetOFcnsValue, we don't really need to know true size:
```
for (int i = 0; i < this.values.length; i++) {
    final int fsz = this.values[i].size();
    if (fsz == 0) {
          sz = 0;
          break;
    }
    if (sz >=  TLCGlobals.enumBound) {
        break;
    }
    sz *= fsz;
}
```

tlatools/org.lamport.tlatools/src/tlc2/value/impl/Value.java
```diff
@@ -176,24 +176,46 @@ public final boolean isEmpty() {
...
+            return false;
           }
+        // THEOREM ESE_TupleSetEmpty ==
+        //   ASSUME NEW S, NEW T
+        //   PROVE  S \X T = {} <=> (S = {} \/ T = {})
         case SETOFTUPLESVALUE:
           {
             SetOfTuplesValue stv = (SetOfTuplesValue)this;
-            return stv.elements().nextElement() == null;
+            for (int i = 0; i < stv.sets.length; i++) {
```
craft095, line 209: Just to be sure: it is impossible for `stv.sets` to be empty, right?

tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
lemmy: Yes, the method comments even carry:

```tla
THEOREM ASSUME NEW S, NEW T, NEW w, T = {w}
        PROVE  Cardinality([S -> T]) = 1
```

tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
lemmy: Yes, the method comments even carry:

```tla
THEOREM ASSUME NEW S, NEW T, NEW w, T = {w}
        PROVE  Cardinality([S -> T]) = 1
```

tlatools/org.lamport.tlatools/src/tlc2/value/impl/SetOfFcnsValue.java
lemmy: Let's keep this for another PR.

tlatools/org.lamport.tlatools/src/tlc2/value/impl/Value.java
lemmy: Added clarifying comments and assertions.

PR #1420, opened 2026-09-09
Fix crash parsing long radix-prefixed numerals (\b, \o, \h)

### LINKED ISSUE #1419: Radix-prefixed numeral literals (\b, \o, \h) crash the parser once long enough
## Description

A radix-prefixed numeral literal (`\b` binary, `\o` octal, `\h` hex) causes SANY to crash with an uncaught `java.lang.NumberFormatException` once it has enough digits, well within limits TLA+ otherwise allows (arbitrary-precision integers). The exception message shows a corrupted fragment of the literal that still has the radix-prefix marker (`\b`/`\o`/`\h`) attached to it — e.g. `"\h012"` — suggesting the lexer/parser splits long numeral tokens into chunks internally and mishandles reattaching (or fails to strip) the prefix on a boundary chunk before handing the string to `Integer.parseInt`/ `BigInteger`.

Found while diffing TLAPM's own parser against `tla2sany.xml.XMLExporter` output in [tlaplus/tlapm#275](https://github.com/tlaplus/tlapm/pull/275) (the `number.txt` "Bitfield Number Formats" case in TLAPM's standardized syntax test corpus, `op5 == \h0123456789abcdef`). Confirmed independent of TLAPM: reproduces with a bare `java -cp tla2tools.jar tla2sany.xml.XMLExporter` invocation.

## Reproduction

`Test.tla`:
```tla
---- MODULE Test ----
op == \h0123456789abcdef
====
```

```sh
java -cp tla2tools.jar tla2sany.xml.XMLExporter -t -I . Test.tla
```
(reproduces the same way through `tla2sany.SANY` directly)

## Observed output

```
java.lang.NumberFormatException: For input string: "\h01" under radix 16
ERROR: Failed to parse module.
tla2sany.drivers.FrontEndException: java.lang.NumberFormatException: For input string: "\h01" under radix 16
	at tla2sany.drivers.SANY.parse(SANY.java:220)
	at tla2sany.xml.XMLExporter.parseSpec(XMLExporter.java:250)
	at tla2sany.xml.XMLExporter.moduleToXML(XMLExporter.java:219)
	at tla2sany.xml.XMLExporter.run(XMLExporter.java:130)
	at tla2sany.xml.XMLExporter.main(XMLExporter.java:116)
Caused by: java.lang.NumberFormatException: For input string: "\h01" under radix 16
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Integer.parseInt(Integer.java:565)
	at java.base/java.math.BigInteger.<init>(BigInteger.java:583)
	at tla2sany.semantic.NumeralNode.<init>(NumeralNode.java:69)
	at tla2sany.semantic.Generator.generateExpressionOrLAP(Generator.java:3073)
	at tla2sany.semantic.Generator.generateExpression(Generator.java:3051)
	at tla2sany.semantic.Generator.processOperator(Generator.java:2660)
	at tla2sany.semantic.Generator.generateModule(Generator.java:2225)
	at tla2sany.semantic.Generator.generate(Generator.java:2166)
	at tla2sany.drivers.SANY.frontEndSemanticAnalysis(SANY.java:452)
	at tla2sany.drivers.SANY.parse(SANY.java:196)
```

## Minimal per-radix thresholds

Bisected with a fragment of repeated `1` digits after the radix prefix (chosen so bit-width scales predictably with digit count):

| Radix | Prefix | Longest that parses OK | Shortest that crashes |
|---|---|---|---|
| Binary | `\b` | 31 digits (31 bits) | **32 digits** (32 bits) |
| Octal | `\o` | 11 digits (33 bits) | **12 digits** (36 bits) |
| Hex | `\h`/`\H` | 8 digits (32 bits) | **9 digits** (36 bits) |

(`\B`, `\O`, `\H` — the uppercase format variants — were not separately re-tested but are expected to share the same underlying numeral-parsing code path.)

Binary's boundary lines up exactly with 32 bits; octal and hex don't line up quite as cleanly, so this may not be a single unified bit-width limit — but in all three cases there is a small, fixed digit-count past which every literal of that radix crashes, regardless of what the actual represented value is (TLA+ integers are otherwise unbounded).

## Expected

All three parse to their numeric value with no length limit, the same way plain decimal numerals of arbitrary length already do.

glondu:
NumeralNode fell back to BigInteger when a numeral overflowed a Java int, but passed the original string (still carrying the \b/\o/\h prefix) to the BigInteger constructor instead of the prefix-stripped string used for the initial Integer.parseInt attempt. BigInteger does not understand TLA+'s radix-prefix syntax, so this threw a NumberFormatException for any sufficiently long binary (>31 digits), octal (>11 digits), or hex (>8 digits) numeral, e.g. \h0123456789abcdef.

Fixes #1419.

### DIFF
--- a/tlatools/org.lamport.tlatools/src/tla2sany/semantic/NumeralNode.java
+++ b/tlatools/org.lamport.tlatools/src/tla2sany/semantic/NumeralNode.java
@@ -66,7 +66,7 @@ public class NumeralNode extends ExprNode {
     try {
       this.value = Integer.parseInt( num, radix );
     } catch ( NumberFormatException e ) {
-      this.bigValue = new BigInteger( s, radix );
+      this.bigValue = new BigInteger( num, radix );
     }
   }
 
--- a/tlatools/org.lamport.tlatools/test/tla2sany/semantic/IncrementalSemanticParseTests.java
+++ b/tlatools/org.lamport.tlatools/test/tla2sany/semantic/IncrementalSemanticParseTests.java
@@ -25,6 +25,7 @@ package tla2sany.semantic;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.math.BigInteger;
 
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


lemmy (approved):
LGTM

Issue #596, opened 2021-03-21
SANY differs from TLA+ language spec for parsing number values with decimals


ahelwer:
According to the formal TLA+ grammar found in Specifying Systems, this statement should be syntactically valid:
```tla
x == .3
```
See [this line](https://github.com/tlaplus/Examples/blob/a9c08fd68e4e646855ca9fb5e49db4804ffc07d4/specifications/SpecifyingSystems/Syntax/TLAPlusGrammar.tla#L32) of the formal grammar; it is `Numeral^*` instead of `Numeral^+` as SANY seems to treat it. No number is required prior to the decimal. SANY instead gives a parse error:

```
***Parse Error***
Was expecting "Expression or Instance"
Encountered "." at line X, column Y in module ZZZZZZ and token "=="
```

lemmy:
Given that there exists an easy workaround and the fact that changing SANY is a major undertaking (manually modified javacc source files), this has very low priority.

lemmy:
Closing as per https://github.com/tlaplus/rfcs/issues/19

PR #1428, opened 2026-09-11
Accept real number literals without a leading zero

glondu:
Allow real number literals written as a dot followed by one or more digits, e.g. `x == .5`, as sanctioned by the formal grammar (Numeral^* & {"."} & Numeral^+). SANY previously required a NUMBER_LITERAL before the DOT, so `.5` was rejected with a parse error even though TLAPM and other tools accept it.

The Number() production now also handles a leading DOT by fabricating a NUMBER_LITERAL token with image "0" for the empty integer part, so the rest of the pipeline (N_Real nodes, AST translators, XML export) sees the usual NUMBER_LITERAL DOT NUMBER_LITERAL shape. "0" was chosen over "" to preserve round-trip formatting; consumers such as TLCs DecimalNode and the TLAPM SANY reader expect a non-empty integer-part image.

The semantic lookahead in PrimitiveExp is extended accordingly to also enter Number() when the lookahead token is a DOT.

`5.` and `.` alone are still rejected, matching the grammar. The corpus test for this case (GH tlaplus/tlaplus #596) was removed from the expected-failures list and now parses to (real_number).

[Feature][SANY][Changelog]


glondu:
The CI check failure seems due to a flakiness and not related to the changes introduced in this PR.

tlatools/org.lamport.tlatools/javacc/tla+.jj
```diff
@@ -3033,15 +3033,31 @@ Number() : {
...
+       sn = new SyntaxTreeNode[3];
+       sn[0] = new SyntaxTreeNode(mn, t1);
+       sn[1] = new SyntaxTreeNode(mn, t2); }
+      t1 = <NUMBER_LITERAL> {
+       sn[2] = new SyntaxTreeNode(mn,t1); } ]
+  | t2 = <DOT> {
+       kind = N_Real;
+       decimalFlag = true;
+       sn = new SyntaxTreeNode[3];
+       Token zero = new Token();
```
lemmy, line 3049: How does the introduction of an artificial token affect the locations of the following expressions? In particular, will all reported locations be shifted by one?

glondu: Only the real-literal node's own location is affected, not the locations of following expressions. The artificial token is a plain Java object referenced solely by the N_Real node's children; it is never inserted into the JavaCC token stream, so the DOT, the fractional digits, and every subsequent token keep their tokenizer-derived begin/end positions.

However, you're right that there was a one-column shift in the literal's own reported start: the fabricated "0" token was placed at `dotColumn - 1`, and since a node's location is the min over its heirs (`SyntaxTreeNode.updateLocation`), `.5` at columns 8-9 was reported as beginning at column 7 (the space before the dot) — and `.5` at line-initial column 1 would even yield `beginColumn = 0`. Fixed in b42d3902b by giving the fabricated token the DOT's own location, so `N_Real` now spans exactly `[dotColumn, fractionalPartEnd]` (verified: `.5` parses to `2:8-2:9` with the dot at column 8).

lemmy: I don't know if the corpus tests have infrastructure for locations, but could you please add a (regression) test for this?

glondu: Done.

ahelwer:
https://github.com/tlaplus/rfcs/issues/19

lemmy:
IIRC, there is a TLAPS extension for proving theorems about real numbers. It would be useful to know whether that extension relies on the TLA+ grammar accepting literals without a leading zero. If it does not, I would also support banning such literals at the language level.

/cc @muenchnerkindl

muenchnerkindl:
After discussion in the SLC, it was decided to sanction SANY's current behavior and change the grammar to require a leading zero.

lemmy:
Closing as per https://github.com/tlaplus/rfcs/issues/19

Issue #487, opened 2020-07-15
Combination of conjunction and nested case-expressions raises parsing error


anton-trunov:
This snippet
```tla
Foo ==
  CASE TRUE ->
        /\ CASE TRUE -> TRUE
             [] TRUE -> TRUE
    [] TRUE -> TRUE
(*     ^  error here *)
```
 raises the following parsing error:
```
Encountered "TRUE" at line 5, column 8, in module Bar and token "[]"
```

I'm actually not 100% sure this is a genuine issue in the parser. This is a minimized and heavily simplified example extracted from auto-generated TLA+ code produced by a transpiler I'm working on. I can fix it in the transpiler itself, but I thought you might be interested.

Note that the following snippets do not raise any parse errors:
```tla
Foo ==
  CASE TRUE ->
        /\ CASE TRUE -> TRUE
             [] OTHER -> TRUE
    [] TRUE -> TRUE
```

```tla
Foo ==
  CASE TRUE ->
        /\ CASE TRUE -> TRUE
             [] TRUE -> TRUE
        /\ TRUE
    [] TRUE -> TRUE
```

ahelwer:
Possibly related: https://github.com/tlaplus-community/tree-sitter-tlaplus/issues/9

The interaction between `CASE` and conjunction/disjunction lists (jlists) is very difficult to parse, because `CASE` constructs are not right-delimited and suffer from [dangling else](https://en.wikipedia.org/wiki/Dangling_else) ambiguity; for example:
```tla
CASE a -> CASE b -> c [] d -> e
```
can be parsed as left-associative:
```tla
CASE a -> (CASE b -> c) [] d -> e
```
or right-associative:
```tla
CASE a -> (CASE b -> c [] d -> e)
```
TLA+ chooses the latter. Further difficulty arises as these are all by themselves valid `CASE` statements:

1. `CASE a -> b`
2. `CASE a -> b [] c -> d`
3. `CASE a -> b [] OTHER c -> d`
4. `CASE a -> b [] c -> d [] OTHER e -> f`

[...]

PR #1429, opened 2026-09-11
Fix parsing of CASE inside conjunction/disjunction lists

glondu:
caseSep() previously treated any `[]` or `□` token as a CASE arm separator, regardless of indentation. In an expression of the form

```tla
    Foo ==
      CASE A ->
        /\ CASE B -> C
           [] D -> E
      [] F -> G
```

the inner CASE therefore swallowed the outer CASE's `[]` arm (`[] F -> G`), because its arm-continuation lookahead ignored the indentation of the `[]`. SANY then rejected the whole expression in checkIndentation() with a spurious "not properly indented" error
(https://github.com/tlaplus/tlaplus/issues/487).

caseSep() now also checks junctionListCtx.isAboveCurrent(t.beginColumn): a `[]` only continues a CASE when it starts strictly to the right of the current junction list's alignment column. At or below that column the CASE terminates and the `[]` is left for the enclosing construct, which matches TLAPM's parsing of these expressions.

TLAplusParser.java was regenerated, and the corresponding ASCII and Unicode corpus tests were removed from the expected-failures list.

[Bug][SANY][Changelog]


lemmy:
Nit: Terminating the `CASE` earlier makes genuinely misindented input lose the specific indentation diagnostic. For example:

```tla
---- MODULE AlignedSep ----
CONSTANT A, B, C, D
op ==
  /\ CASE A -> B
  [] C -> D
====
```

Previously SANY reported:

> Item at line 5, col 3 to line 5, col 4 of module AlignedSep is not properly
> indented inside conjunction or disjunction list item at line 4, col 3 to
> line 5, col 11 of module AlignedSep

It now reports:

> Encountered "[]" at line 5, column 3 and token "B"

Both correctly reject the module, but the previous message was more actionable.

glondu:
> Nit: Terminating the `CASE` earlier makes genuinely misindented input lose the specific indentation diagnostic. For example:
> [...]
> Both correctly reject the module, but the previous message was more actionable.

The old wording depends on `checkIndentation()` running, which only happens after a parse success that the fix is specifically eliminating — so the nicer message is not recoverable without reintroducing the ambiguity (#487). I don't think the caseSep-localized error-message special case is worth the parser complexity here.

lemmy (approved):
LGTM

lemmy:
Merged in https://github.com/tlaplus/tlaplus/pull/1442

Issue #430, opened 2020-02-12
naming a theorem after a submodule raises a parsing error


johnyf:
```tla
---------------------------- MODULE SubmodParsingError -------------------------
(* It appears that naming a theorem after a submodule raises a parsing error. *)


--------------- MODULE Inner -------------------

================================================


THEOREM T == TRUE
(* SANY raises ***Parse Error***
Encountered "THEOREM Beginning of definition"
*)

THEOREM TRUE  (* no parsing error *)

================================================================================
```

quaeler:
This issue would be best remediated at the grammar file level approximately [here](https://github.com/tlaplus/tlaplus/blob/master/tlatools/javacc/tla%2B.jj#L1761). @lemmy is this something we want to pursue? (I ask because the grammar files are rarely touched.)

PR #1430, opened 2026-09-11
Fix parsing of a named theorem following a submodule

glondu:
belchDEF() is invoked in Body() after a Module() definition to insert a DEFBREAK before the next "==", which OperatorOrFunctionDefinition() needs to recognize the start of a definition. Its scanning loop stopped only on nextT.kind, never on currentT.kind, so when the token immediately after a submodule was a THEOREM (or PROPOSITION/LEMMA/COROLLARY), the loop still stepped onto the theorem's name and "==" and inserted a DEFBREAK between the THEOREM keyword and its name. That corrupts the token stream and makes the Theorem() lookahead fail, so the theorem was never consumed as a body element and EndModule() failed with "Encountered THEOREM ... expecting ====" (https://github.com/tlaplus/tlaplus/issues/430).

The while condition now also stops when currentT itself is a THEOREM or PROPOSITION token, so a theorem following a submodule is left untouched for Theorem() to parse. Named theorems/propositions/lemmas/corollaries in plain modules and definitions following a submodule are unaffected.

TLAplusParser.java was regenerated, and the corresponding corpus test was removed from the expected-failures list.

[Bug][SANY][Changelog]

### DIFF
--- a/tlatools/org.lamport.tlatools/javacc/tla+.jj
+++ b/tlatools/org.lamport.tlatools/javacc/tla+.jj
@@ -266,7 +266,9 @@ public class TLAplusParser implements tla2sany.st.SyntaxTreeConstants, ParseTree
            * date described below.                                         *
            ****************************************************************/
            && nextT.kind != ASSUMPTION 
-           && nextT.kind != END_MODULE) {
+           && nextT.kind != END_MODULE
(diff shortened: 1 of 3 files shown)


lemmy (approved):
LGTM

PR #1405, opened 2026-08-20
Add JPF verification for the DiskStateQueue pool-writer protocol.

lemmy:
Related to Github issue #1403
https://github.com/tlaplus/tlaplus/issues/1403

Issue #1113, opened 2025-01-09
-recover doesn't work: the checkpoint file is probably corrupted.   


arssher:
## Description

I'm running TLC on a big model. If model checking is interrupted, `-recover` often fails. It seems to reproduce well if model checking is left for several days running, so `states/` grows to dozens of gigabytes. If interrupted quickly after start it works. Recover errors are different, the last one I saw is
```
Starting... (2025-01-09 11:52:32)                                                                                                                                           
Starting recovery from checkpoint states/24-12-27-13-51-35.505/                                                                                                             
Error: TLC encountered the following error while restarting from a checkpoint;                                                                                              
 the checkpoint file is probably corrupted.                                                                                                                                 
Finished in 25s at (2025-01-09 11:52:58)
```

One time I saw
```
Starting recovery from checkpoint states/24-11-02-08-58-10.765/
The exception was a util.WrongInvocationException
: ValueInputStream: Can not unpickle a value of kind -1
```
but haven't preserved directory for that one.

It is interesting that once recovery starts I see that states directory quickly gets significantly smaller, e.g. from 142GB becomes 106G in the first run reported here. Probably it is expected, i.e. uncheckpointed tail is removed, but looks like a bit too much, not sure.

## Steps to Reproduce

- clone https://github.com/neondatabase/neon
- cd into `safekeeper/spec/`
- run `./modelcheck.sh models/MCProposerAcceptorReconfig_p2_a4_t2_l2_c3.cfg MCProposerAcceptorReconfig.tla`. I adjust `modelcheck.sh` to run on 40 cores and with 30GB mem, but not sure it matters.
- Wait a couple days
- Kill the process or reboot the machine
- Run script again, adding `-recovery <state-directory>` e.g. `-recovery states/24-12-27-13-51-35.505` to TLC invocation.

This takes a lot of time and cpu though. Alternatively I can send these 106GB from the last run somewhere if that helps.  I'm also attaching full TLC output:
```
Running breadth-first search Model-Checking with fp 11 and seed 4383372531238629468 with 40 workers on 80 cores with 27307MB heap and 30720MB offheap memory [pid: 3219231] 
(Linux 6.8.0-49-generic aarch64, Ubuntu 21.0.5 x86_64, OffHeapDiskFPSet, DiskStateQueue).                                                                                   
Parsing file /home/arseny/neon2/safekeeper/spec/MCProposerAcceptorReconfig.tla                                                                                              
[...]

PR #1431, opened 2026-09-13
TLC: Fix race condition when checkpointing liveness graph

ahelwer:
Wait to restart workers until liveness graph has been snapshot

When TLC takes a snapshot while running multiple workers and when liveness checking is enabled, it can create an invalid checkpoint file which will trigger this assertion after restoring from it:
https://github.com/tlaplus/tlaplus/blob/867aefb69ffc2452031292587b389d1fc3eb43ff/tlatools/org.lamport.tlatools/src/tlc2/tool/liveness/LiveCheck.java#L770-L778

`theStateQueue.beginChkpt()` snapshots the pending-state queue while workers are suspended. Previously, `resumeAll()` restarted those workers before `liveCheck.beginChkpt()` snapshotted the liveness graph. During that interval, a worker could remove a state from the live queue and add its outgoing transitions to the liveness graph. The checkpoint would then contain both:
1. a queue snapshot in which the state is still pending, and
2. a liveness-graph snapshot in which that state has already been expanded.

On recovery, TLC processes the queued state again. `LiveCheck.addNextState` finds that its graph node already contains all of the same successors, so the successor count does not increase. That case is expected only during simulation, and model checking consequently triggers the assertion linked above.

Moving `resumeAll()` after `liveCheck.beginChkpt()` ensures that the queue and liveness graph describe the same suspended point in the exploration.

An invalid checkpoint file triggering this assertion can be downloaded here: [trigger.zip](https://github.com/user-attachments/files/32165913/trigger.zip)

However, this checkpoint file is not actually valid against TLC built on the master branch and requires running a build of TLC against the branch of #1359, which changes state variable order. For basic security hygiene, consider doing this inside a sandbox as you are building unknown code and executing it on an unknown checkpoint file. After building TLC on that branch, to reproduce the bug:
1. Start shell in `tlatools/org.lamport.tlatools` dir
4. `unzip trigger.zip -d trigger` (this must be done before every run as subsequent execution modifies the unzipped files)
5. `cd test-model/CodePlexBug08` (oddly this fails with a different error when running from outside this directory)
6. `java -ea -cp ../../dist/tla2tools.jar tlc2.TLC EWD840MC2 -config EWD840MC2.cfg -gzip -recover ../../trigger/checkpoint`

This will produce the following error trace:
>TLC2 Version 2026.09.13.182427 (rev: 3c7cdf3)
[...]
Starting... (2026-09-13 11:31:32)
Implied-temporal checking--satisfiability problem has 1 branches.
Starting recovery from checkpoint ../../trigger/checkpoint/
AAAAAA
Recovery completed. 1025 states examined. 1013 states on queue.
Checking temporal properties for the current state space with 1055 total distinct states at (2026-09-13 11:31:36)
Finished checking temporal properties in 00s at 2026-09-13 11:31:36
Progress(2) at 2026-09-13 11:31:36: 1,395 states generated (1,395 s/min), 1,236 distinct states found (1,236 ds/min), 1,136 states left on queue.
The exception was a java.lang.RuntimeException
:
11122 states generated, 2180 distinct states found, 1151 states left on queue.
The depth of the complete state graph search is 3.
The average outdegree of the complete state graph is 1 (minimum is 0, the maximum 5 and the 95th percentile is 4).
Finished in 08s at (2026-09-13 11:31:40)


lemmy:
The PR LGTM, but ideally, this data race should be confirmed with JPF, similar to how JPF was used to confirm the concurrency issue in #1405.

ahelwer:
I would like to do that but unfortunately I pretty consistently trip the cybersecurity filter when asking for tests that reproduce bugs (same reason the checkpoint file is for a non-master branch).

lemmy:
This apparently also fixes a race in `UniqueString.internTbl`: workers previously resumed before its unsynchronized checkpoint serialization completed, allowing concurrent interning and a potentially inconsistent `vars.chkpt`.

lemmy:
https://github.com/tlaplus/tlaplus/issues/1113
https://github.com/tlaplus/tlaplus/pull/1398

lemmy:
Thanks, (logically) merged in 078405c22df8037571860457b2d078e8281bd851.

Issue #1433, opened 2026-09-14
IntervalValue enumeration wraps after MaxInt


lemmy:
`IntervalValue#elements` and the `i <= high` loops in `#diff`, `#cap`, `#cup`, and `#fingerPrint` increment past `Integer.MAX_VALUE` and overflow. `MaxInt..MaxInt` therefore does not terminate as the singleton `{MaxInt}`, so subset, set operations, fingerprints, and conversion of `[x \in MaxInt..MaxInt |-> 0]` are unsound.

```tla
---- MODULE MaxIntInterval ----
EXTENDS Integers, FiniteSets, TLC
MaxInt == 2147483647

ASSUME MaxInt..MaxInt \subseteq {MaxInt}              \* FALSE
ASSUME MaxInt..MaxInt \ {} = {MaxInt}                \* runaway
ASSUME [x \in MaxInt..MaxInt |-> 0] = [MaxInt |-> 0] \* ArrayIndexOutOfBoundsException
ASSUME Cardinality(MaxInt..MaxInt) = 1               \* TRUE
ASSUME MaxInt \in MaxInt..MaxInt                     \* TRUE
====
```

Issue #1432, opened 2026-09-14
Empty interval-domain functions disagree on compareTo and equals


lemmy:
Equality treats every empty interval as the empty set, so `[x \in 2..1 |-> x] = [x \in 3..2 |-> x]` is `TRUE`. Ordering still uses the unused bounds (`2` vs `3`), so a set containing both keeps two elements. Nested cardinality, `SUBSET`, union, fingerprints, and `[S -> T]` over mixed empty functions are therefore unsound even when direct `=` holds.

```tla
---- MODULE EmptyIntervalFcn ----
EXTENDS Integers, FiniteSets, TLC

ASSUME [x \in 2..1 |-> x] = [x \in 3..2 |-> x]                    \* TRUE

ASSUME Cardinality({[x \in 2..1 |-> x], [x \in 3..2 |-> x]}) = 1  \* FALSE (2)
ASSUME {[x \in 2..1 |-> x], <<>>} = {<<>>}                        \* FALSE
ASSUME Cardinality(SUBSET {[x \in 2..1 |-> x], <<>>}) = 2         \* FALSE (4)
ASSUME <<>> = [x \in 1..(-1) |-> x]                               \* NegativeArraySizeException thrown when TLC converts  to a tuple, which allocates an array of length `-1`.
====
```

PR #1434, opened 2026-09-14
Soundness issues TLC Value classes

lemmy:
https://github.com/tlaplus/tlaplus/issues/1432
https://github.com/tlaplus/tlaplus/issues/1433


tlatools/org.lamport.tlatools/src/tlc2/value/impl/FcnRcdValue.java
```diff
@@ -216,6 +216,9 @@ private final int compareOtherInterval(final FcnRcdValue fcn) {
 
   private final int compareToInterval(final FcnRcdValue fcn) {
   	int result;
+  	if (this.values.length == 0) {
+  		return 0;
+  	}
   	if (fcn.intv != null) {
   		result = this.intv.low - fcn.intv.low;
```
craft095: `result = this.intv.low - fcn.intv.low;`

It can also overflow and e.g., return less than 0 result when ` `this.intv.low > fcn.intv.low`

tlatools/org.lamport.tlatools/src/tlc2/value/impl/FcnRcdValue.java
craft095: This also looks suspicious: it's difficult to argue that there is no overflow possible.

craft095: Can this calculation overflow?

lemmy:
More arithmetic overflows fixed here were found by applying OpenJML verification to FcnRcdValue.java. The resulting assertions are not committed for now, but may be added in the future.

PR #1436, opened 2026-09-15
Fix SANY parsing of `-` in nonfix operator application

glondu:
Expression() and ExtendableExpr() use a lookahead that commits to the prefix interpretation of "-" (token op_77) whenever an expression starts with it, so `op == - (1, 2)` is rejected with an unexpected-comma parse error. Guard the lookahead with a check that the parenthesized argument list following the operator contains a top-level comma, in which case the operator is applied nonfixly to at least two arguments and must not be taken as a prefix.

The Nonfix Minus corpus case now parses to the expected bound_nonfix_op tree.  This allows progress on fixing the tlapm parser test suite when it is run with the SANY backend, in which the remaining #884 cases and a handful of other corpus cases still fail.

[Bug][SANY]

Ref: https://github.com/tlaplus/tlapm/pull/275
Ref: https://github.com/tlaplus/tlaplus/issues/884


glondu:
I am investigating the tlaplus/examples CI failures.

lemmy:
@ahelwer, would you have a chance to review this PR? Given your experience with the TLA+ grammar and SANY—especially your work on the SANY corpus tests and the TLA+ tree-sitter grammar—your feedback would be particularly valuable here. Thanks!

lemmy:
@glondu I'm inclined to leave the PR open until a second reviewer approves it, unless this issue is blocking the TLAPS SANY migration.

PR #1439, opened 2026-09-16
Fix SANY parsing of `!!` in nonfix operator application

glondu:
JavaCC's maximal-munch lexer cannot split a run of "!" characters into the alternating subexpression separator "!" and nonfix infix operator "!!" tokens that the parser needs, so `A!!!(1,2)` was rejected with an unexpected-token parse error.  Lex a run of three or more bangs as a single BANGRUN token and split it in BangExt into a "!" separator followed by a "!!" operator (registering the "!!" via the same path InfixOp() uses, including lastOp).  Longer runs have no unambiguously intended parse and are rejected with a message recommending inserting a space to break them up, matching how the tree-sitter corpus punts on such runs.

The "Nonfix Submodule Excl" and "Nonfix Double Exclamation Operator" corpus cases now parse to the expected prefixed_op/subexpr_prefix trees; the corpus output translator additionally learns to translate an operator-symbol node (prefix, infix, or postfix) used as a subexpression component, which these trees introduce.  This allows progress on fixing the tlapm parser test suite when it is run with the SANY backend.

[Bug][SANY]

Ref: https://github.com/tlaplus/tlapm/pull/275
Ref: https://github.com/tlaplus/tlaplus/issues/884


lemmy:
@ahelwer, would you have a chance to review this PR? Given your experience with the TLA+ grammar and SANY—especially your work on the SANY corpus tests and the TLA+ tree-sitter grammar—your feedback would be particularly valuable here. Thanks!

lemmy:
@glondu I'm inclined to leave the PR open until a second reviewer approves it, unless this issue is blocking the TLAPS SANY migration.

PR #1441, opened 2026-09-18
Check TLC value arithmetic with JBMC

lemmy:
Add JBMC to CI and use solver-generated inputs to check arithmetic and value semantics. The harnesses encode expected behavior without known boundary values. JBMC independently discovers inputs such as Integer.MAX_VALUE that expose overflow defects. Narrow overlays isolate unsupported JDK exact-arithmetic methods and irrelevant error formatting.


lemmy:
Will be part of https://github.com/tlaplus/tlaplus/pull/1434

